PR #62, opened 2015-01-24
Convenientce: Enable opening an ELF file by pathname rather than file()

zachriggle:


eliben:
I don't think the small convenience this buys is worth it. Esp. with Python 2 & 3 compatibility (w.r.t. file names being unicode or bytesrings)

PR #64, opened 2015-01-24
Using the MutableMapping type is unnecessary.  Subclassing dict and addi...

zachriggle:
...ng a few functions is all that's required.  Additionally, this makes the output more readable when printing a Container type.


eliben:
Please note that `construct` is a 3rd party library I'm using almost "as is" here; I don't want to create needless deviations. They did have more modern releases though - now, a pull request merging that would be very welcome ;-)

PR #68, opened 2015-02-22
rename readelf.py to readelf

thinred:
This PR renames readelf.py to pyreadelf.


eliben:
Why would you install it into /usr/bin anyway? It's just an example script and useful for testing - I don't intend it to replace readelf. Renaming it in a way that obscures its purpose just for the sake of an unusual requirement of one distribution doesn't make much sense to me.

thinred:
Well, "setup.py install" installs readelf.py, so I thought it is part of the distribution. If it is not, then why is it in "scripts" in setup.py? Also, I don't think it is unusual requirement to separate the implementation language from the name of the program.
Ok, so I assume that readelf.py is not a part of pyelftools "build". It actually makes the situation simpler. Thanks Eli!

PR #69, opened 2015-04-07
ELF: Find all symbols of a given name in `get_symbol_by_name`.

Smattr:
It is possible for an ELF file's symbol table to contain multiple entries under
the same symbol name. Prior to this commit, it was only possible to retrieve
the first of these via `get_symbol_by_name`. This commit alters this function
to return a list of all symbols under the given name, rather than just the
first entry. Functionality when a symbol name does not exist remains
unaffected.

----

This PR relates to comments on 46ae4bd3a5f7b0e355913b3497f0e96ac2778ef5.

By the way, is there a way to test changes on Travis before submitting a pull request?


Smattr:
Done both corrections.

PR #82, opened 2015-08-17
Added support for DWARF v4 line program headers.

### LINKED ISSUE #81: DWARF V4 Line Programs not supported
The current code doesn't support DWARF v4 Line program headers, specifically it doesn't understand the maximum_operations_per_instruction field resulting in everything getting out of alignment and the program crashing out.

I have a fix for this on my work laptop which I could push if desired?

Tustvold:
I had to use a MetaField as unfortunately a non DWARF v4 compile unit can reference a DWARF 4 line program and therefore an if statement in structs.py is insufficient.

This fixes #81


elftools/dwarf/structs.py
```diff
@@ -11,7 +11,7 @@
     UBInt8, UBInt16, UBInt32, UBInt64, ULInt8, ULInt16, ULInt32, ULInt64,
     SBInt8, SBInt16, SBInt32, SBInt64, SLInt8, SLInt16, SLInt32, SLInt64,
     Adapter, Struct, ConstructError, If, RepeatUntil, Field, Rename, Enum,
-    Array, PrefixedArray, CString, Embed, StaticField
+    Array, PrefixedArray, CString, Embed, StaticField, IfThenElse, Value
```
eliben: No need to import these if unused ?

elftools/dwarf/structs.py
```diff
@@ -226,6 +226,11 @@ def _create_lineprog_header(self):
             self.Dwarf_uint16('version'),
             self.Dwarf_offset('header_length'),
             self.Dwarf_uint8('minimum_instruction_length'),
+			
+			If(lambda ctx: ctx['version'] >= 4, 
+				self.Dwarf_uint8("maximum_operations_per_instruction"),
+				1),
```
eliben: I'm actually surprised just a plain `1` works here.

eliben:
Looks much better, thanks! Can you squash into a single commit with as descriptive a commit message as you can (explain the new field added and what it means)? Thanks

elftools/dwarf/structs.py
Tustvold: The If macro encapsulates the 1 in a Value object and so this behaves correctly.

From macros.py:
```python
def If(predicate, subcon, elsevalue = None):
    return IfThenElse(subcon.name,
        predicate,
        subcon,
        Value("elsevalue", lambda ctx: elsevalue)
    )
```

PR #84, opened 2015-09-05
Fixed dynamic symbols iteration in the dynamic segment

ramti:
The previous dynamic symbol lookup worked by heuristically determine the end of the dynamic symbol table, because the dynamic segment does not provide the symbol table length, but only a pointer to the first entry. This heuristic attempts to iterate the dynamic segment tags in order to find the nearest higher pointer to the symbol table entry (the end of the table), which supposes to work in most cases. The problem is that not all the tag values in the dynamic segment represent pointers (d_ptr), some of them represent other values (d_val, for example in DT_NEEDED), and the code assumes that all the tag values represent a pointer. That leads to a wrong result and to an iteration on a partial symbol table.

Instead of the above logic, I decided to implement another method to determine the length of the dynamic symbol table: the hash table. The hash table contains an array called "chains", which contains (according to the specification) the same number of entries as the dynamic symbol table, and therefore this number (found in the hash table header) can be used to determine the number of the dynamic symbols in the dynamic symbol table.

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -199,35 +199,16 @@ class DynamicSegment(Segment, Dynamic):
         if tab_ptr is None or tab_offset is None:
             raise ELFError('Segment does not contain DT_SYMTAB.')
 
+        # Get number of symbols from the hash table header. The size of the array "chain"
+        # equals to the number of the dynamic symbols.
+        hash_ptr, hash_offset = self.get_table_offset('DT_HASH')
+        hash_tbl = struct_parse(self._elfstructs.Elf_SymHashHdr, self._stream, hash_offset)
+        number_of_symbols = hash_tbl.nchain
+        
+        # Iterate the symbol table
         symbol_size = self._elfstructs.Elf_Sym.sizeof()
-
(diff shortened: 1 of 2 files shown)


vadmium:
Although the hash table is meant to be mandatory, on Arch Linux I found most of the ELFs (elves?) do not have a DT_HASH entry, although they have a DT_GNU_HASH instead. I have code based on _pyelftools_ that can iterate over the symbols in both kinds of hash tables (rather than trying to iterate the symbol table directly). Standard hash table: <https://github.com/vadmium/pacman-tools/blob/e462670/elf.py#L284>; Gnu hash table: <https://github.com/vadmium/pacman-tools/blob/e462670/elf.py#L333>. I could port into _pyelftools_ itself if people were interested.

ramti:
Thank you, I checked again and I did found some binaries that didn't have DT_HASH values. Anyway, the current logic needs to be changed, because it doesn't work well, as I've described above. I will add an implementation to the pool request (in the next 24-hours) that can iterate the GNU hash table.

eliben:
@ramti  have you updated the patch to address @vadmium's comments?

eliben:
Closing due to inactivity. If anyone is willing to reopen and rebase this, please open a new PR

PR #88, opened 2015-09-27
Fix Note Type Id decoding.

dethrophes:
The ELF Note TypeId is actually used as a sub type to the Note Name.
That is to say that the enum values currently being used to detect the Note Type are only valid if the Note Name is 'GNU\0'.

This change rectifies that. 
It also handles the case that a segment may be slightly larger than the contained notes.

### DIFF
--- a/elftools/elf/segments.py
+++ b/elftools/elf/segments.py
@@ -11,6 +11,7 @@ from ..common.utils import roundup, struct_parse
 from ..common.py3compat import bytes2str
 from .constants import SH_FLAGS
 
+from .enums import ENUM_GNU_NOTE_N_TYPE
 
 class Segment(object):
     def __init__(self, header, stream):
@@ -109,7 +110,7 @@ class NoteSegment(Segment):
         """
         offset = self['p_offset']
         end = self['p_offset'] + self['p_filesz']
-        while offset < end:
+        while offset+self._elfstructs.Elf_Nhdr.sizeof() < end:
             note = struct_parse(
                 self._elfstructs.Elf_Nhdr,
                 self.stream,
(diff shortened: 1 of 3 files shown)


elftools/elf/segments.py
```diff
@@ -11,6 +11,7 @@
 from ..common.py3compat import bytes2str
 from .constants import SH_FLAGS
 
+from .enums import ENUM_GNU_NOTE_N_TYPE
```
eliben: Move this closer to the `from` above it

elftools/elf/segments.py
```diff
@@ -109,7 +110,7 @@ def iter_notes(self):
         """
         offset = self['p_offset']
         end = self['p_offset'] + self['p_filesz']
-        while offset < end:
+        while offset+self._elfstructs.Elf_Nhdr.sizeof() < end:
```
eliben: style - spaces between binary operators

elftools/elf/segments.py
```diff
@@ -119,11 +120,20 @@ def iter_notes(self):
             self.stream.seek(offset)
             # n_namesz is 4-byte aligned.
             disk_namesz = roundup(note['n_namesz'], 2)
+            # skip parsing truncated notes.
+            if offset+disk_namesz+roundup(note['n_descsz'], 2) > end:
```
eliben: style - spaces between binary operators

elftools/elf/segments.py
```diff
@@ -119,11 +120,20 @@ def iter_notes(self):
...
+                break;
             note['n_name'] = bytes2str(
                 CString('').parse(self.stream.read(disk_namesz)))
             offset += disk_namesz
 
             desc_data = bytes2str(self.stream.read(note['n_descsz']))
+            if note['n_name'] == 'GNU':
```
eliben: It's not clear why this is needed, whereas it wasn't needed before? We don't usually linearly walk all enum values in elftools to find a match

dethrophes: Well I'm not tied to the linear walk, if there is a better was of doing it in python.
I guess if the enum(dict) was inverted i.e. key, values then it wouldn't be necessary.


But the real problem is that I didn't see anyway way to have multiple enum lists for this value depending on the note name.

dethrophes: I'm not actually sure what you are asking here.
IF you are asking why we have to do a linear walk of a dick to do a lookup by value.. then AFAIK that jsut how dicts and hash tables are implemented. Somebody is having to do this anyway, in elftools you seem to be letting the helper lib do it.

if you are asking why we need to only do this when the name is GNU. well because its the correct way. This Ids are not unique they are only sub types of the note name. i.e. They can only be used if the Note name is GNU, there are other tools, vendors, oems that add notes that aren't GNU and they also start at id 0 for the type. As such the current code incorrectly names these notes as for example note-id.  Which means that with some elf files, elftools returns the wrong content, e.g. a build-id that is not actually a build-id but rather some random note content.

eliben: Thanks for the detailed explanation!

So when enums are embedded in construct, as was the case before your change - they *are* properly "inverted".

Given that you no longer use ENUM_NOTE_N_TYPE in structs.py, how about moving it out of enums (these are specifically for enums embedded in structs) and just put it somewhere in the note segment where you can then invert the map properly (it's not used elsewhere, is it?)

And please put all of this in a comment somewhere explaining this special treatment of notes - I wasn't aware of this and it will be good to keep the knowledge with the code.

eliben:
Ping - did you have a chance to address the latest round of comments?

eliben:
Closing due to inactivity. If anyone is willing to reopen and rebase this, please open a new PR

PR #89, opened 2015-12-03
Improved DWARFv4 support.

sethml:
Handle DW_AT_data_member_location attributes with type DW_FORM_data*, which
store a simple integer offset rather than dwarf expression. This seems to come
up with struct members, at least on ARM.

Handle DW_AT_location attributes with type DW_FORM_sec_offset as not having
dwarf expressions, as allowed by the DWARFv4 standard. This seems to come up
with some local variables, at least on ARM.


eliben:
Thanks for the PRs. It will take me a bit to go over these, though, since I'm backlogged.

eliben:
Looks good, but can you add some tests?

sethml:
Added a test, and removed overlap with #91.

eliben:
Seems like Travis shows testing broke - PTAL

sethml:
Looks like there's a dependency on #89. I thought I'd fixed that, but it's hard to compile a non-trivial C program with gcc 4.8.4 without tripping #89. Let's get that merged first, and then this change should work.

eliben:
Did you mean some other #PR number? This is 89...

eliben:
[also please squash PRs into a single git commit]

sethml:
Squashed. Oops, I meant #91.

Can I rebase on top of #91 to make the test pass? Tried it. You'll probably need to pull #91 first.

eliben:
Both merged and tests seem to pass! Thanks for the contribution

PR #91, opened 2015-12-03
Add DW_TAG_GNU_call_site* for better dumping.

sethml:
This comes up for functions which have tail-calls optimized.


eliben:
Is there any chance you have a test file that can tickle this, and can be placed in the readelf-comparison test suite?

sethml:
Added a test file, and improved formatting to match readelf a bit better.

eliben:
Thanks! Just one more thing - please squash PR into a single commit and then it's ready to go

sethml:
Squashed.

PR #94, opened 2015-12-31
Fix relative import statements for Python 3.x compatibility.

thinred:
When executing: "python3 -m unittest discover" the tests fail.
Python 3 requires a dot in front of the import to make a relative import.

Cheers!


thinred:
It seems that the correct way to run the tests is tu run test/all_tests.py. My bad!

PR #101, opened 2016-05-24
made elf/segments.py not choke on unterminated note names

inglorion:
elf/segments.py tries to parse ELF note names as CStrings. This will fail if the note name does not end in a NUL character. It turns out some binaries have note names that are not NUL-terminated. To prevent elftools from choking on such files, we detect when the name is not NUL-terminated and return all its bytes instead of trying to parse it as a CString.

### DIFF
--- a/elftools/elf/segments.py
+++ b/elftools/elf/segments.py
@@ -8,7 +8,7 @@
 #-------------------------------------------------------------------------------
 from ..construct import CString
 from ..common.utils import roundup, struct_parse
-from ..common.py3compat import bytes2str
+from ..common.py3compat import byte2int, bytes2str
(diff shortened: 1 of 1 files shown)


eliben:
Any chance you have a minimal test binary that demonstrates this?

Also, prior to merging I'll need this squashed into a single commit

Thanks!

eliben:
Any updates here? I'll need a test to merge.

eliben:
Note parsing has been moved to `elf/notes.py`

Please rebase this patch on current master

eliben:
Closing due to inactivity. Feel free to reopen or create a new PR if needed

PR #102, opened 2016-05-24
Support ZLIB compressed debug sections

ttsugriy:
objcopy --compress-debug-sections uses ZLIB compression to reduce debug
sections, which can sometimes be larger than the size of the binary
itself. This change makes pyelftools consider compressed debug sections
when checking for DWARF data.

readelf tool supports other types of compressed sections
(https://github.com/facebook/binutils/blob/master/binutils/readelf.c#L12038)
but their support is outside of scope of this change.

Test plan:
  $ ./test/run_all_unittests.py


eliben:
Thanks for the contribution. I'm currently traveling, so I'll only be able to review this in a few weeks.

ttsugriy:
Sure, no rush. Enjoy your travel!

elftools/elf/elffile.py
```diff
@@ -22,6 +26,8 @@
 from .segments import Segment, InterpSegment, NoteSegment
 from ..dwarf.dwarfinfo import DWARFInfo, DebugSectionDescriptor, DwarfConfig
 
+CHUNK_SIZE = resource.getpagesize()
```
eliben: Since this is used in a single place, I'd prefer it declared next to the usage inside the method

elftools/elf/elffile.py
```diff
@@ -138,32 +145,47 @@ def get_dwarf_info(self, relocate_dwarf_sections=True):
         # present.
         # Sections that aren't found will be passed as None to DWARFInfo.
         #
+
+        compressed = bool(self.get_section_by_name('.zdebug_info'))
```
eliben: why not put compressed next to its usage rather than above section_names?

elftools/elf/elffile.py
```diff
@@ -376,3 +398,40 @@ def _read_dwarf_section(self, section, relocate_dwarf_sections):
...
+        section.stream.seek(0)
+        # According to readelf.c the content should contain "ZLIB"
+        # followed by the uncompressed section size - 8 bytes in
+        # big-endian order
+        compression_type = bytes2str(section.stream.read(4))
+        assert compression_type == 'ZLIB', \
+            'Invalid compression type: %r' % (compression_type)
+
+        uncompressed_size = 0
+        for _ in range(8):
```
eliben: using the struct module here would be cleaner, I think. You probably need `struct.unpack` with the Q format

eliben:
Looks pretty good thanks! I made some minor comments. Please also squash into a single commit so I can merge it

elftools/elf/elffile.py
ttsugriy: good point! It used to be, but after shuffling a code a bit I forgot to relocate this line :)

elftools/elf/elffile.py
ttsugriy: Love it! Thanks for an awesome suggestion!

ttsugriy:
Thanks for the awesome suggestions Eli! I've addressed all of your comments and squashed all commits into a single one.

PR #104, opened 2016-06-19
Fix case where struct is None

P1kachu:
Signed-off-by: Stanislas P1kachu Lejay <p1kachu@lse.epita.fr>


P1kachu:
When there are no CU, it raises an exception which might break a very long program for nothing (which happened in my case). It's not mandatory to handle it though, so if you prefer it might be left to the user to handle it :)

eliben:
OK, it seems like a reasonable "fail-safe" option to return None here. Thanks

PR #105, opened 2016-06-20
Remove construct copy from source tree and change code to use package.

manueljacob:
This worked very seamlessly - other than changing imports no changes to the source code were needed. Tests pass on Python 2.7 and 3.5.

Changes in more detail:
- delete directory elftools/construct
- add construct as a dependency in setup.py
- change imports to use package
- add install command to .travis.yml


eliben:
I really prefer keeping construct bundled in. This package goes in and out of maintenance, and I had problems in the past where I had to fix bugs in it but it  wasn't worked on. I prefer not to tie into it as a dependency, though I'm fine with someone just updating the bundled version (I think it was done in the past).

PR #107, opened 2016-07-01
Fix Windows import error by getting page size using 'mmap' module if 'resource' is not available.

DavidSpickett:
The resource module is Unix only, changed to use mmap for Windows.

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -7,9 +7,17 @@
 # This code is in the public domain
 #-------------------------------------------------------------------------------
 import io
-import resource
 import struct
 import zlib
+
+try:
+    import resource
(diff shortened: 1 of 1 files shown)


elftools/elf/elffile.py
```diff
@@ -7,9 +7,17 @@
...
 import io
-import resource
 import struct
 import zlib
+
+try:
+    import resource
+    PAGESIZE = resource.getpagesize()
+except ImportError:
+    #Windows system
```
eliben: nit: space after # in comment

eliben:
Thanks. Please squash to single commit and I'll merge

DavidSpickett:
Squashed and corrected comment.

PR #108, opened 2016-07-08
parse .debug_aranges section

dorothychen:
parse aranges table in dwarf and provide simple interface to get a CU offset from an address

### DIFF
--- a/elftools/dwarf/aranges.py
+++ b/elftools/dwarf/aranges.py
@@ -1,0 +1,110 @@
+#-------------------------------------------------------------------------------
+# elftools: dwarf/aranges.py
+#
+# DWARF aranges section decoding (.debug_aranges)
+#
+# Dorothy Chen (dorothchen@gmail.com)
+# This code is in the public domain
+#-------------------------------------------------------------------------------
+import os
+from collections import namedtuple
+from ..common.utils import struct_parse
+from bisect import bisect_right
+import math
+
+# An entry in the aranges table; 
+# begin_addr: The beginning address in the CU
+# length: The length of the address range in this entry
+# info_offset: The CU's offset into .debug_info
+# see 6.1.2 in DWARF4 docs for explanation of the remaining fields
+ARangeEntry = namedtuple('ARangeEntry', 
+    'begin_addr length info_offset unit_length version address_size segment_size')
+
+class ARanges(object):
+    """ ARanges table in DWARF
+
+        stream, size: 
+            A stream holding the .debug_aranges section, and its size
+
+        structs: 
+            A DWARFStructs instance for parsing the data
+    """
+    def __init__(self, stream, size, structs):
+        self.stream = stream
+        self.size = size
+        self.structs = structs
+
+        # Get entries of aranges table in the form of ARangeEntry tuples
+        self.entries = self._get_entries()
+
+        # Sort entries by the beginning address
+        self.entries.sort(key=lambda entry: entry.begin_addr)
+
+        # Create list of keys (first addresses) for better searching
+        self.keys = [entry.begin_addr for entry in self.entries]
+
+
+    def cu_offset_at_addr(self, addr):
+        """ Given an address, get the offset of the CU it belongs to, where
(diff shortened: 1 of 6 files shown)


elftools/dwarf/aranges.py
```diff
@@ -0,0 +1,88 @@
...
+        self.stream = stream
+        self.size = size
+        self.structs = structs
+
+        # entries is a sorted list of ARangeEntry tuples, 
+        # sorted by beginning address
+        self.entries = self._get_entries()
+        self.entries.sort(key=lambda entry: entry[0])
+        self.keys = [entry[0] for entry in self.entries]
+
```
eliben: here and elsewhere: please one blank line between class methods

elftools/dwarf/aranges.py
```diff
@@ -0,0 +1,88 @@
...
+import os
+from collections import namedtuple
+from ..common.utils import struct_parse
+from bisect import bisect_left
+import math
+
+ARangeEntry = namedtuple('ARangeEntry', 'begin_addr length info_offset')
+
+class ARanges(object):
+    """ ARanges table in DWARF
```
eliben: Lacks comments: how to use this class? Methods and ctor need usage comments too

elftools/dwarf/aranges.py
```diff
@@ -0,0 +1,88 @@
...
+                addr = struct_parse(addr_size('addr'), self.stream)
+                length = struct_parse(addr_size('length'), self.stream)
+                while addr != 0 and length != 0:
+                    entries.append(
+                        ARangeEntry(begin_addr=addr, 
+                            length=length, 
+                            info_offset=aranges_header["debug_info_offset"]))
+                    addr = struct_parse(addr_size('addr'), self.stream)
+                    length = struct_parse(addr_size('length'), self.stream)
+
```
eliben: no blank lines between if/elif bodies

elftools/dwarf/aranges.py
```diff
@@ -0,0 +1,88 @@
...
+            elif segment_size != 0:
+                raise NotImplementedError("Segmentation not supported")
+
+            offset = (offset 
+                + aranges_header.unit_length 
+                + self.structs.initial_length_field_size())
+
+        return entries
+
+    def _get_addr_size(self, addr_header_value):
```
eliben: I'd rename this to `_get_addr_size_struct` or something of the sort to make it clear it's a struct that should be parsed; same for the usages

eliben:
Thanks for the contribution!

I left some inline comments. Also, this is missing tests. I'm pretty sure many of the existing test objects/binaries for readelf-comparison tests already have aranges sections so you can enable this section in comparisons and make sure the tests exercise it and everything works.

dorothychen:
Noted! I've made the cosmetic changes and added aranges to the readelf tests. 

The formatting of readelf/greadelf might be slightly different on Mac vs linux (some of the original tests fail because of extra or too few zeros on my machine)

dorothychen:
Also, does it make sense to have a method in ARanges that returns a CU given an address? I am not quite sure if it's more consistent with the library to return an offset or the actual object.

eliben:
> Also, does it make sense to have a method in ARanges that returns a CU given an address? I am not quite sure if it's more consistent with the library to return an offset or the actual object.

Good question. I wonder how do you intend to use it? Are you going to call `DWARFInfo._parse_CU_at_offset` after getting the offset from `ARanges`?

dorothychen:
> Good question. I wonder how do you intend to use it? Are you going to call DWARFInfo._parse_CU_at_offset after getting the offset from ARanges?

Yes, as of now all my use cases involve getting the CU offset and immediately parsing the CU at that offset

eliben:
So I'd leave it as it is for now. In general, it probably makes sense to expose `_parse_CU_at_offset` as a public method of `DWARFInfo`.

You can go ahead and squash all commits in this PR into a single one, and I'll merge it.

dorothychen:
squashed/rebased

PR #109, opened 2016-07-15
Support SHT_NOTE sections

gbenson:
This commit adds a new NoteSection class with an iter_notes iterator which operates in much the same way as NoteSegment's iter_notes.

### DIFF
--- a/elftools/elf/notes.py
+++ b/elftools/elf/notes.py
@@ -1,0 +1,43 @@
+#-------------------------------------------------------------------------------
+# elftools: elf/notes.py
+#
+# ELF notes
+#
+# Eli Bendersky (eliben@gmail.com)
+# This code is in the public domain
+#-------------------------------------------------------------------------------
+from ..common.py3compat import bytes2str
+from ..common.utils import struct_parse, roundup
+from ..construct import CString
+
+
+def iter_notes(elffile, offset, size):
+    """ Yield all the notes in a section or segment.
+    """
+    end = offset + size
+    while offset < end:
+        note = struct_parse(
+            elffile.structs.Elf_Nhdr,
+            elffile.stream,
+            stream_pos=offset)
+        note['n_offset'] = offset
+        offset += elffile.structs.Elf_Nhdr.sizeof()
+        elffile.stream.seek(offset)
+        # n_namesz is 4-byte aligned.
+        disk_namesz = roundup(note['n_namesz'], 2)
+        note['n_name'] = bytes2str(
+            CString('').parse(elffile.stream.read(disk_namesz)))
+        offset += disk_namesz
+
+        desc_data = bytes2str(elffile.stream.read(note['n_descsz']))
+        if note['n_type'] == 'NT_GNU_ABI_TAG':
+            note['n_desc'] = struct_parse(elffile.structs.Elf_Nhdr_abi,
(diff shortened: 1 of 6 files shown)


elftools/elf/sections.py
```diff
@@ -174,3 +174,15 @@ def iter_symbols(self):
...
+    """ ELF NOTE section. Knows how to parse notes.
+    """
+    def __init__(self, header, name, stream, elffile):
+        super(NoteSection, self).__init__(header, name, stream)
+        self.elffile = elffile
+
+    def iter_notes(self):
+        """ Iterates the list of notes in the segment.
+        """
+        return self.elffile._iter_notes(self['sh_offset'], self['sh_size'])
```
eliben: Why not do the parsing here, in NoteSection?

elftools/elf/segments.py
```diff
@@ -102,36 +101,9 @@ class NoteSegment(Segment):
...
-                                              self.stream,
-                                              offset)
-            elif note['n_type'] == 'NT_GNU_BUILD_ID':
-                note['n_desc'] = ''.join('%.2x' % ord(b) for b in desc_data)
-            else:
-                note['n_desc'] = desc_data
-            offset += roundup(note['n_descsz'], 2)
-            note['n_size'] = offset - note['n_offset']
-            yield note
+        return self.elffile._iter_notes(self['p_offset'], self['p_filesz'])
```
eliben: Ah I see... I'd prefer to have it factored into a separate module, like elf/notes.py

eliben:
Thanks; I left some comments. Also, are there tests? Some of the binaries in the readelf-testing harness should have notes sections, maybe?

elftools/elf/segments.py
gbenson: Ok, I'll do that.

gbenson:
Thanks for your suggestions. I've refactored it into `elftools/elf/notes.py` as you suggested, and I've written an example which the testsuite executes.

elftools/elf/sections.py
```diff
@@ -174,3 +176,16 @@ def iter_symbols(self):
...
+
+class NoteSection(Section):
+    """ ELF NOTE section. Knows how to parse notes.
+    """
+    def __init__(self, header, name, stream, elffile):
+        super(NoteSection, self).__init__(header, name, stream)
+        self.elffile = elffile
+
+    def iter_notes(self):
+        """ Iterates the list of notes in the segment.
```
eliben: s/segment/section/ here

elftools/elf/notes.py
```diff
@@ -0,0 +1,43 @@
...
+# Eli Bendersky (eliben@gmail.com)
+# This code is in the public domain
+#-------------------------------------------------------------------------------
+from ..common.py3compat import bytes2str
+from ..common.utils import struct_parse, roundup
+from ..construct import CString
+
+
+def iter_notes(elffile, offset, size):
+    """ Iterates the list of notes in a section or segment.
```
eliben: Can this comment briefly state what is actually yielded on each iteration?

eliben:
Thanks, looks much better. I added a couple more minor comments and it's good to go (please squash again before I can pull)

gbenson:
Is this ok?

PR #115, opened 2016-08-29
Extend relocation patching in Relocationhandler

laanwj:
This adds two optional arguments to apply_section_relocations:
- symbol_mapping: The symbol mapping maps symbol ordinals to addresses. If no symbol mapping is provided, symbols are taken to be at absolute address st_value from their entry in the symbol table (as
  was the behavior before).
- section_base can be passed, which is the starting address of the section to be relocated in memory. This is used to correctly emit pc-relative relocations. If not passed this will default to 0.

These are necessary to relocate executable files with imports at a non-zero base address in memory.

### DIFF
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -113,17 +113,27 @@ class RelocationHandler(object):
                 return relsection
         return None
 
-    def apply_section_relocations(self, stream, reloc_section):
+    def apply_section_relocations(self, stream, reloc_section, symbol_mapping=None, section_base=0):
         """ Apply all relocations in reloc_section (a RelocationSection object)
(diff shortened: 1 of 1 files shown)


eliben:
It would be nice to see tests for this

laanwj:
Sure. Is it allowed to introduce a dependency on [unicorn](http://www.unicorn-engine.org/)? :)

eliben:
No :)

Ideally a real binary/object file that can go into readelf tests

eliben:
Any updates here?

laanwj:
Yes, I still intend to write a test here, but I've been very busy

eliben:
Closing. Please file a new PR with tests, if applicable

PR #118, opened 2016-09-25
Check for header attribute when comparing

tezeb:
When checking against None it throws excepiton, while it's quite easy to prevent(and check against None seems quite obvious).

### DIFF
--- a/elftools/elf/sections.py
+++ b/elftools/elf/sections.py
@@ -41,7 +41,11 @@ class Section(object):
         return self.header[name]
 
     def __eq__(self, other):
-        return self.header == other.header
+        try:
+            return self.header == other.header
+        except AttributeError:
+            return False
+    
     def __hash__(self):
         return hash(self.header)
 


elftools/elf/sections.py
```diff
@@ -41,7 +41,9 @@ def __getitem__(self, name):
         return self.header[name]
 
     def __eq__(self, other):
-        return self.header == other.header
+        if hasattr(other, 'header'):
```
eliben: It would be more pythonic to "try" this and have "except AttributeError".

elftools/elf/sections.py
```diff
@@ -41,7 +41,9 @@ def __getitem__(self, name):
...
+            return self.header == other.header
+        return False
```
eliben: Preserve free lines between methods

eliben:
Thanks! Squash into a single commit and I'll merge

tezeb:
Sorry, I was editing it through web interface and I thought it will get squashed automatically.
btw. according to following document, you can probably squash while merging:
https://github.com/blog/2141-squash-your-commits

Issue #71, opened 2015-04-29
Non-determinism of dict enumeration results in non-deterministic parsing


dcolascione:
If we enumerate the segments of an ARM EABI ELF file and print the p_type field of each, we print different results on different runs with the same input. Both PT_ARM_UNWIND and PT_AARCH64_UNWIND have the value 0x70000001; we decode this value using the Enum facility of the Construct library. Depending on the order in which we enumerate the ENUM_P_TYPE entries, Enum can prefer either PT_ARM_UNWIND or PT_AARCH64_UNWIND.

Enumeration mappings we provide to Construct should never be ambiguous!

eliben:
I forget, but is it actually correct for them to have the same value?

bigboze:
Yes, the values for PT_ARM_UNWIND and PR_AARCH64_UNWIND fall within the PT_LOPROC - PT_HIPROC range, which is an inclusive range reserved for processor-specific semantics. (0x70000000 - 0x7fffffff)

bigboze:
Same should ideally be applied for *_LOPROC - *_HIPROC and _*LOOS - *_HIOS ranges for other ELF fields.

eliben:
Also discussed in #121

eliben:
I believe this should be addressed by the latest commit (linked above ^^^).

PR #121, opened 2016-10-24
Handle ARM relocations

frederiksdun:
This PR adds support to handle ARM's R_ARM_ABS32 relocation.
Fixes issue #86


frederiksdun:
Any hints why https://travis-ci.org/eliben/pyelftools/builds/170086267 failed, but 202 built?

eliben:
Looking at the error log it says:

```

  File "scripts/readelf.py", line 1229, in <module>

    main()

  File "scripts/readelf.py", line 1189, in main

    show_heading=not do_file_header)

  File "scripts/readelf.py", line 256, in display_section_headers

    nsec, section.name, describe_sh_type(section['sh_type'])))

  File "./elftools/elf/descriptions.py", line 58, in describe_sh_type

    elif x >= ENUM_SH_TYPE['SHT_LOOS'] and x < ENUM_SH_TYPE['SHT_GNU_versym']:

TypeError: unorderable types: str() >= int()

@@ aborting - 'scripts/readelf.py' returned '1'
```
You could try re-running it locally on Python 3.3+ to reproduce

frederiksdun:
@eliben I think the current problem is that exidx and unwind are defined as 0x70000001 which is used as key in a dict(). Is unwind a valid sh_type?

eliben:
Not sure if just kicking them out is the right approach. These enum values are in "processor-specific" space which means many of them collide between processors. I think if you look in enums.py you'll see a bunch of such duplicated values, no?

frederiksdun:
Yes I see duplicates in enum.py which cause problems in description.py.
Even PT_ARM_EXIDX and PT_ARM_UNWIND have the same value which would conflict if descriptions are machine aware.

Do we need the descriptions for UNWIND?

eliben:
I think this is also related to #71

eliben:
Ugh, this is a nasty issue. I think that for any given processor, all enums should be unique. I found that on ARM, EXIDX and UNWIND are synonyms (http://infocenter.arm.com/help/topic/com.arm.doc.ihi0044f/IHI0044F_aaelf.pdf) so I removed UNWIND.

However, across different processors it's valid to have duplications in that range (above `0x70000000`). Here there's a deficiency in pyelftools parsing, in that it doesn't take the arch into account when parsing these fields. It should. This is also mentioned in #71.

eliben:
Presumably the test file you're adding hits this. Can you try to clean it up a bit (maybe just remove these segments) so it passes all tests? Then the PR can be pulled, since the parsing issue is unrelated

frederiksdun:
I rebased against master

eliben:
It seems like it's still failing on Travis for Python 3.4 and 3.5 -- maybe it's still the non-determinism?

eliben:
@frederiksdun I'll look into making the arch-specific field parsing work (as detailed here and in #71) -- it may unblock this PR. May take me a bit to find time for this, though

eliben:
A couple of recent commits made `p_type` and `sh_type` parsing dependent on `e_machine`, which should resolve the non-determinism issue in enums that was plaguing this.

I'll be happy to merge the PR if it's cleanly rebased on the latest master and all tests pass.

frederiksdun:
I want to update some more details before merging

frederiksdun:
@eliben Feel free to merge

Issue #123, opened 2016-11-01
fail to import elffile on windows (+patch)


screwer:
Due to missing 'getpagesize' field in 'resource':

> D:\GITREP\python\pyelftools\scripts>D:\GITREP\python\pyelftools\scripts\readelf.py
> Traceback (most recent call last):
>   File "D:\GITREP\python\pyelftools\scripts\readelf.py", line 23, in <module>
>     from elftools.elf.elffile import ELFFile
>   File "C:\tools\Python27\lib\site-packages\elftools\elf\elffile.py", line 15, in <module>
>     if not resource.getpagesize:
> AttributeError: 'module' object has no attribute 'getpagesize'

There is simple patch:
```
try:
    import resource
++    if "getpagesize" not in dir(resource):
++        raise ImportError
    PAGESIZE = resource.getpagesize()
except ImportError:
    # Windows system
    import mmap
    PAGESIZE = mmap.PAGESIZE
```

eliben:
Would you mind submitting a pull request? This way I can just merge it, Travis will run tests on it, etc.

PR #124, opened 2016-11-01
Fix for issue #123

screwer:
missing 'getpagesize' function in resource library on windows.

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -12,6 +12,8 @@ import zlib
 
 try:
     import resource
+    if "getpagesize" not in dir(resource):
+        raise ImportError
     PAGESIZE = resource.getpagesize()
 except ImportError:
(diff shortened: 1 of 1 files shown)


elftools/elf/elffile.py
```diff
@@ -12,6 +12,8 @@
 
 try:
     import resource
+    if "getpagesize" not in dir(resource):
```
eliben: I think it would be more Pythonic to check for AttributeError along with ImportError in the `except` clause. What do you think?

vadmium: I agree it would be simpler to catch AttributeError.

BTW hasattr(resource, "getpagesize") would have been a better than using dir().

eliben:
Any updates here?

eliben:
Closing due to inactivity. Reopen if needed

PR #127, opened 2016-12-20
Fix for padding encoding in Python 3.

mefistotelis:
The padding used a string literal instead of byte literal for the
pattern character, which caused invalid type error in Python 3.

This fix marks the pattern as byte literal, which does nothing on
Python >= 2.7 and creates bytes array instead of str in Python >= 3.

### DIFF
--- a/elftools/construct/adapters.py
+++ b/elftools/construct/adapters.py
@@ -380,12 +380,12 @@ class PaddingAdapter(Adapter):
     
     Parameters:
     * subcon - the subcon to pad
-    * pattern - the padding pattern (character). default is "\x00"
+    * pattern - the padding pattern (character as byte). default is b"\x00"
     * strict - whether or not to verify, during parsing, that the given 
       padding matches the padding pattern. default is False (unstrict)
     """
     __slots__ = ["pattern", "strict"]
-    def __init__(self, subcon, pattern = "\x00", strict = False):
+    def __init__(self, subcon, pattern = b"\x00", strict = False):
         Adapter.__init__(self, subcon)
         self.pattern = pattern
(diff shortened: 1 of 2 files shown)


eliben:
Thanks. Can you add a regression test, please?

mefistotelis:
Is there a simple tutorial/guidelines on how to create the test?

eliben:
There's a bunch of unit tests in the `test` directory. In your case it seems that you encountered some file on which parsing went bad? Can you reproduce with a minimal ELF file and add a simple unit test for reading it, making sure it doesn't blow up? Or however otherwise the bug manifested before your fix

mefistotelis:
I actually implemented a mechanism of modifying ELF header and sections. The bug manifested itself when I tried to save changes back to the ELF file. The write() method for binary files in Python 3 requires bytes and not str, so it caused invalid type exception.

eliben:
So this isn't something pyelftools uses currently? You're trying to use the included `construct` lib to write the ELF stuff back?

mefistotelis:
pyelftools uses its Padding class; whether it uses _encode() method from it - I don't know.
I described how I spotted the issue; I don't know whether it can be encountered in other scenarios.

I did used construct lib in my code; I made function struct_write() to save headers, similar the existing struct_parse().

Some of my low-level functions should be easily acceptable for you; but the higher level design should probably be discussed; my implementation has API which might be questionable. Also, it's WIP - some changes are not even commited yet.

eliben:
Got it, thanks for explaining. I see this change is similar to other modifications I made to construct in the past for Python 3 compatibility, so I feel it's ok submitting it.

PR #128, opened 2016-12-23
Added support for 8 & 16 bit addresses to DWARF Info

vanwinkeljan:


eliben:
Thanks. Is it possible to add a unit test for this?

PR #129, opened 2017-01-03
Ignore ascii decoding errors when creating special tag attributes.

destrius:
In the DynamicTag constructor, when _HANDLED_TAGS are being processed, stringtable.get_string() tries to ascii-decode the requested string from the string table. In some files the string can't be decoded, resulting in a UnicodeDecodeError. Ignore the error rather than fail to create the tag.

Specifically, I have a file whose DT_SUNW_FILTER tag whose d_val does not map to an ascii string.

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -66,7 +66,7 @@ class DynamicTag(object):
         return '<DynamicTag (%s): %r>' % (self.entry.d_tag, self.entry)
 
     def __str__(self):
-        if self.entry.d_tag in self._HANDLED_TAGS:
+        if (self.entry.d_tag in self._HANDLED_TAGS) and (self.entry.d_tag[3:].lower() in dir(self)):
             s = '"%s"' % getattr(self, self.entry.d_tag[3:].lower())
         else:
             s = '%#x' % self.entry.d_ptr
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -47,8 +47,15 @@ class DynamicTag(object):
             raise ELFError('Creating DynamicTag without string table')
         self.entry = entry
         if entry.d_tag in self._HANDLED_TAGS:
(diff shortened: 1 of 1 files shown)


eliben:
Can you provide a minimal file as a test?

I'm wondering if there's a larger issue here - what encoding *is* your file? Should a different decoding be used in `get_string`?

destrius:
I'll try to generate a test file; might take a while though as I'm currently busy with something else.

I poked around a bit at the particular file that gave me problems, and I think it might be malformed; the d_val of the tag is beyond the extents of the string table. So it's essentially returning garbage bytes. As far as I can tell, ascii should be the correct encoding for get_string, if the file is well formed. Unfortunately I can't disclose the file, but I can say that it is a standard library that came with a popular Android device.

My rationale for ignoring the decoding error is that _HANDLED_TAGS seems like a convenience feature, and so not being able to create an attribute for the malformed tag shouldn't be a deal breaker. In particular, the exception makes it less convenient to use iter_tags(); you'd have to explicitly call .next() inside a try-except block within the loop block.

If however you'd rather not mess up the code to handle a malformed file, I understand.

I also realised that my commit breaks DynamicTag.__str__(); I pushed up a fix for that in my fork.

eliben:
Thanks for the details. There's a ton of places where `pyelftools` would explode for malformed files. Due to the nature of its inputs (object files generated and consumed by tools), there's never been much emphasis on handling bad inputs in a tolerable way. The safest thing the library can do is explode with an exception

destrius:
Sure, thanks for looking at this!

rinon:
I'd like to reopen this pull request (or something like it) because Android uses OS-specific .dynamic tags 0x6000000f and 0x60000010 as `DT_ANDROID_REL` and `DT_ANDROID_RELSZ` respectively. This conflicts with the `DT_SUNW_FILTER` usage as a string table index. It is unsafe to assume that an OS-specific .dynamic entry must be a string table index when not all OSes comply with this usage.

PR #134, opened 2017-02-01
 Fix parsing string table

gcmoreira:
If for any reason parse_cstring_from_stream() cannot get a string in the given offset, it'll return None causing the following line raises an exception when it tries to get the decoded ascii string from the None object.


gcmoreira:
I'm sorry about the multiple commits, I had some issues with the temporary deprecation of unicode literals in Python 3.2, very annoying. I've tested the fix in 2.7,3.2,3.3,3.4 & 3.5.

PR #135, opened 2017-02-12
Enable parsing of relocations pointed to by DYNAMIC.

nneonneo:
In combination with DynamicSegment, this allows relocations to be processed without
a functional section table.

Example usage:

```python
from elftools.elf.dynamic import DynamicSegment
from elftools.elf.relocation import get_dynamic_reloc_tables

import sys

elff = ELFFile(open(sys.argv[1], 'rb'))

for seg in elff.iter_segments():
    if isinstance(seg, DynamicSegment):
        relos = get_dynamic_reloc_tables(elff.stream, elff, seg)
        for relname in relos:
            print relname
            for rel in relos[relname].iter_relocations():
                print rel
```


eliben:
Thanks for the contribution, and sorry about missing it for 2 weeks. My work on pyelftools comes in bursts.

I would much prefer to avoid multiple inheritance whenever possible. What do you think?

Also, this PR sorely misses tests

rupran:
I was about to implement something similar but then noticed this PR.

@eliben: would you prefer something like a class derived from RelocationSection which builds a 'mock' header from the information in .dynamic over multiple inheritance? To me that feels a bit unclean as we don't really have a Section backing the relocation table.

I saw that especially in dynamic.py we already use multiple inheritance to have a base class to share between Section and Segment objects so that's probably where the idea came from.

nneonneo:
I definitely let this fall off my radar - for two years! I'd love to see this still get into pyelftools. I can work up a few tests (mostly just files with bad or missing section tables) based on that example code I posted above.

Re multiple inheritance: I did base that off patterns I already saw in the code, so I was hoping it'd fit in with the existing code style.

eliben:
Some tests and usage samples would definitely be useful here to assess the actual design. I'd want to see how the code to use this looks.

nneonneo:
I've just added a short unit test, which should serve as a usage sample.

eliben:
@rupran @rhelmot would you like to help reviewing this?

elftools/elf/relocation.py
```diff
@@ -89,6 +90,54 @@ def iter_relocations(self):
...
+            self['sh_offset'], self['sh_size'], header['sh_type'] == 'SHT_RELA')
+
+        elf_assert(header['sh_type'] in ('SHT_REL', 'SHT_RELA'),
+            'Unknown relocation type section')
+        elf_assert(header['sh_entsize'] == self.entry_size,
+            'Expected sh_entsize of %s section to be %s' % (
+                header['sh_type'], self.entry_size))
+
+
+def get_dynamic_reloc_tables(stream, elffile, dynamic):
```
rupran: Is there a case where we need a different `stream` than the one from the `elffile` parameter? If not, it would probably be cleaner to just pass `elffile` and access the stream within this function.

I'm inclined to say this would also be the case for the `dynamic` parameter, so we could do all checking inside this function and just pass in an `ELFFile`.

nneonneo: I think previously `stream` was a separate parameter to `RelocationSection`'s constructor. At some point this changed, so I corrected `RelocationSection`'s signature but left `stream` as a separate parameter. I can correct everything to just remove the `stream` parameter and rely on `elffile`.

`dynamic`, on the other hand, might be either a section or a segment, and I don't want to preclude the API user from specifying it. If it helps, maybe I could make it default to `None` which triggers some automatic behaviour (e.g. default to the DYNAMIC segment).

rupran: It's still not fully consistent between segments and sections (the individual  `Segment` constructors get passed a `stream`, and some an `ELFFile` as well - `DynamicSegment` is one of those) but I'd prefer dropping the `stream` parameter here.

For `dynamic`, you are right, with either a segment or a section passed in this is more flexible. Adding a default parameter would probably only worsen readability so maybe it's best to just leave it as a mandatory parameter.

nneonneo:
Updated: removed the `stream` parameter to rely on the `elffile`'s own stream where possible/meaningful.

rhelmot:
`get_dynamic_reloc_tables` seems like something which should really be a method on the Dynamic base class. That would make the stream and the elffile both available, unless I'm missing something very big.

rupran:
Oh, indeed, I somehow overlooked the fact that `DynamicSegment` and `DynamicSection` have a common base class... So yes, it's probably best to put this into the `Dynamic` class.

nneonneo:
Happy to oblige - latest patch set moves `get_dynamic_reloc_tables` into Dynamic as `Dynamic.get_relocation_tables`.

eliben:
Thanks for the contribution @nneonneo, and for the reviews @rupran and @rhelmot

PR #140, opened 2017-03-28
Remove construct from elftools and use it as a dependency

rupran:
This removes the old version of `construct` - which was copied into this repository years ago - from `elftools` and ports all required files to work with the most recent version (2.8.10) from https://github.com/construct/construct

The syntactical changes to files in the `elftools/{common,dwarf,elf}` directories (found in the third commit of the series) are caused by
- renames of datatypes (e.g., `ULInt8` -> `Int8ul`) and classes (e.g., `Embed` -> `Embedded`) in `construct`, see http://construct.readthedocs.io/en/latest/transition.html for an overview
- the nameless initialization of Structs/Enums and datatypes by using the `/` operator
- changes to flag propagation in `If()` (it doesn't carry embedded flags anymore), so we now need to wrap the `If()` inside `Embedded()` instead of the other way around.
- `If()` lost its `elsevalue` parameter, so we need to use `IfThenElse` with the correct alternative for the else case.

I'm not sure about the correct `setup.py` integration because the currently released version 2.8.10 of `construct` has a python2 compatibility issue (see https://github.com/construct/construct/commit/98d6c5fb50bc9a3a13107071f0c870c29d7d1500, which was only merged after 2.8.10 was released). The last commit of this series *tries* to use `construct`'s GitHub repository (which already includes the fix) as a dependency but this requires `pip install` to be called with `--process-dependency-links`... I'm not totally happy with that and am happy to take comments on that.

Additionally, the just merged commit https://github.com/construct/construct/commit/4836722526cb9bf0de655176c6dc6bf3bc177669 changes the interface of `RepeatUntil` (more specifically, the `predicate` parameter takes another parameter - the current state of the list) which we 
- use in https://github.com/eliben/pyelftools/blob/master/elftools/dwarf/structs.py#L311 and 
- should probably take into `RepeatUntilExcluding` as well.

So maybe we should wait until 2.8.11 (or whatever the next version of `construct` will be called) is released... then we could just use that `construct` release from PyPI and get around `dependency_links`.


eliben:
What is the motivation for this change?

I ended up doing the copy back in the day because construct became unmaintained and I had to fix bugs and port it to Python 3. It may become unmaintained again, and I like the dependency-less situation more, to be honest.

rupran:
Well, it looked like development upstream had picked up again (with new releases and such) so I thought I'd give it a try as I was playing around with elftools anyway and it was in the TODO file. :-)

That said, I'm not aware of any bugs that would have bitten elftools by not upgrading the copy of construct.

rupran:
By dropping the last commit (and renaming the `construct28` folder and all its imports back to `construct`) we could also basically perform an update of the integrated version, thus staying dependency-less "on the outside". Not sure if that's really necessary or desired, that just came to my mind.

eliben:
Is there a specific bug this update would address?

rupran:
Not that I'm aware of.

eliben:
I prefer not to merge this. The API changes seem considerable and gratuitous, and I don't see anything this would fix.

... To be honest, I'm considering getting rid of the `construct` - I'm running more and more into its limitations, and complicating those struct definitions more and more. #71 is the latest instance where this would need to be done.

PR #146, opened 2017-05-28
NULL termination with IAR Embedded Workbench compiler

gerthmar:
The IAR Embedded Workbench compiler inserts a zero termination byte after some dies (eg after DW_TAG_subrange_type @0x2885 in the attached .elf file). This leads to parsing error of the subsequent die(s), since null is assumed to be the first byte of it.

This workaround consumes all null bytes that follow a complete die.
[iar_ewb.zip](https://github.com/eliben/pyelftools/files/1034393/iar_ewb.zip)

### DIFF
--- a/elftools/dwarf/die.py
+++ b/elftools/dwarf/die.py
@@ -191,18 +191,19 @@ class DIE(object):
                 value=value,
                 raw_value=raw_value,
                 offset=attr_offset)
-		
-        # some compilers add a zero termination after the die -> consume the
-        # zero termination to avoid wrong die size calculation
-        zero_term = False
+
+        # count and then consume any null termination bytes to avoid wrong die size calculation
+        zero_term = 0
         with preserve_stream_pos(self.stream):
-            v = self.stream.read(1)
-            if(v[0]==0):
-                zero_term = True
-        
-        if(zero_term):
-            # there was a zero termination -> consume it
-            self.stream.read(1)
+            while True:
+                b = self.stream.read(1)
+                if b == 0:
+                    zero_term += 1
+                else:
+                    break
+        if zero_term > 0:
+            # there was at least one zero termination -> consume all of them
+            self.stream.read(zero_term)
 
         self.size = self.stream.tell() - self.offset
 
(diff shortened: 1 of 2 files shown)


gerthmar:
I don't understand why this pull request failed to test. All tests are running fine on my machine. Any ideas?

eliben:
Added some review comments.

Another question: can a test be added? Do you have a simple test file produced by this compiler?

The PR seems to fail on the Python 3 builders -- see Travis logs in the "Details" link

gerthmar:
Couldnt find your review comments...

The elf file is attached in my first comment to this pull request as zip file. I will add it to git once the test is pass.

I am now able to reproduce the failed test on python3 (was using python2 on my machine). Looks like there is a more complicated issue with this pull request. Will have a look at it.

gerthmar:
Applied your review findings. The tests are now pass. 

However with the iar_ewb input file, there are still other issues, including formatting differences between pyelftools and readelf, not related to this pull request.

So I think you could pull the PR.

eliben:
I'm just curious about the other issues you're seeing with the test file. `run_readelf_tests.py` has a bunch of logic to deal with formatting differences, for example.

gerthmar:
The .zip in this post contains the test-script output, as well as the temporary files generated by readelf, so should be pretty straight forward to see the differences.

Some examples:

**for --debug-dump=info:** 
- readelf also dumps .debug_types section

**for --debug-dump=frames:**
pyelf:
```
  DW_CFA_advance_loc: 2 to 0000000008000b82
  DW_CFA_def_cfa_offset_sf: <??>
```
readelf:
  ```
DW_CFA_advance_loc: 2 to 08000b82
  DW_CFA_def_cfa_offset_sf: 8
```


[iar_ewb_test.zip](https://github.com/eliben/pyelftools/files/1059270/iar_ewb_test.zip)

eliben:
It seems like it could help to fix up the comparison to readelf (making sure pyelftools emits the same output), but this is less critical. If you have time for this (since the diff manifests for files you care about), I'll be happy for a separate PR that also adds the file as a test.

For now I can pull this PR

Issue #93, opened 2015-12-11
Incorrect parsing of CORE types


zachriggle:
CORE files use notes to store register information.  In particular, NT_GNU_ABI_TAG and NT_PRSTATUS are the same value.

Currently, `pyelftools` assumes the former and discards the [data stream](https://github.com/eliben/pyelftools/blob/d7aec6aba2cfed93e69dfe177cc161ccab42f15c/elftools/elf/segments.py#L127).

This should not be done when `elftype` is `"CORE"`.

zachriggle:
As an example, here's one that I finagled so as not to go through the GNU ABI tag parsing;

```
Container({'n_desc': '\x0b\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x0b\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xc9\xbc\x00\x00u\xbc\x00\x00\xc9\xbc\x00\x00\xc9\xbc\x00\x00\x00\x00\x00\x00@\x9c\x00\x00\x00\x00\x00\x00\xe0.\x00\x00\x00\x00\x00\x00%\x05\x00\x00\x00\x00\x00\x00\x9c&\x00\x00\xf4\x7frU*`WU\xb8\x98rU\x00\x00\x00\x00\x00\x00\x00\x00\xf4\x7frU\x00\x00\x00\x00+\x00\x00\x00+\x00\x00\x00\x00\x00\x00\x00c\x00\x00\x00\xff\xff\xff\xff\x00\x00tj#\x00\x00\x00\x86\x02\x01\x00*`WU+\x00\x00\x00\x00\x00\x00\x00', 'n_type': 'NT_GNU_ABI_TAG', 'n_name': 'CORE', 'n_offset': 564, 'n_descsz': 144, 'n_size': 164, 'n_namesz': 5})
```

eliben:
Could you please try to submit a pull request for this?

zachriggle:
After my previous three pull requests were abandoned or rejected? No thanks.  However, this is still a valid bug report since the tag parsing is b0rken.

You can see how I fixed it here: https://github.com/binjitsu/binjitsu/blob/master/pwnlib/elf/corefile.py#L22-L48 and how I parsed the data here: https://github.com/binjitsu/binjitsu/blob/master/pwnlib/elf/corefile.py#L152-L176 

https://github.com/eliben/pyelftools/pull/64 <-- Rejected
[...]

PR #147, opened 2017-06-14
Better support for core dumps (in particular, fix issue #93)

zlobober:
Hi Eli,

This PR fixes the main part of issue #93. 

Working with core dumps was important for our internal use of your library in Yandex, so I did the following steps:

- I splitted struct initialization and ELF parsing into two steps. 
- First of them is for parsing of only the ELF file header, and in particular, the information about ELF type that defines if we are working with a regular ELF file or a core dump. 
- On second step we build the remaining ELF structures and parse them. In particular, we build correct note types when working with core dumps on this phase.
- I added a new test binary (x64 core dump) into "testfiles_for_readelf". Some cases were not working as intended for it, I fixed all of them but one.

What I didn't do:

- The test with option "-n" doesn't work for core dumps properly. Mainly, because output format of vanilla readelf for core dumps notes is surprisingly different from the one for usual notes, e.g.:

```
$ test/external_tools/readelf -n test/testfiles_for_readelf/core_simple64.elf  

Displaying notes found at file offset 0x00000430 with length 0x000008b0:
  Owner                 Data size       Description
  CORE                 0x00000150       NT_PRSTATUS (prstatus structure)
  CORE                 0x00000088       NT_PRPSINFO (prpsinfo structure)
  CORE                 0x00000080       NT_SIGINFO (siginfo_t data)
  CORE                 0x00000130       NT_AUXV (auxiliary vector)
  CORE                 0x000002af       NT_FILE (mapped files)
    Page size: 4096
                 Start                 End         Page Offset
    0x0000000000400000  0x0000000000401000  0x0000000000000000
        /home/max42/pyelftools/test/testfiles_for_readelf/coredump_self
    0x0000000000401000  0x0000000000402000  0x0000000000000000
        /home/max42/pyelftools/test/testfiles_for_readelf/coredump_self
    0x0000000000402000  0x0000000000403000  0x0000000000000001
        /home/max42/pyelftools/test/testfiles_for_readelf/coredump_self
    0x00007ffff7a1c000  0x00007ffff7bd0000  0x0000000000000000
        /lib/x86_64-linux-gnu/libc-2.15.so
    0x00007ffff7bd0000  0x00007ffff7dcf000  0x00000000000001b4
        /lib/x86_64-linux-gnu/libc-2.15.so
    0x00007ffff7dcf000  0x00007ffff7dd3000  0x00000000000001b3
        /lib/x86_64-linux-gnu/libc-2.15.so
    0x00007ffff7dd3000  0x00007ffff7dd5000  0x00000000000001b7
        /lib/x86_64-linux-gnu/libc-2.15.so
    0x00007ffff7dda000  0x00007ffff7dfc000  0x0000000000000000
        /lib/x86_64-linux-gnu/ld-2.15.so
    0x00007ffff7ffc000  0x00007ffff7ffd000  0x0000000000000022
        /lib/x86_64-linux-gnu/ld-2.15.so
  CORE                 0x00000200       NT_FPREGSET (floating point registers)

```

You see, the note lines are joined together into a single block, and the NT_FILE-typed note has lots of extra information.

I didn't have much time for fixing it, so I just explicitly skipped this test in your testing suite.

I don't really want to fix it since it requires a lot of time to understand how the orignial readelf outputs information in this case and then careful repeating this behavior. Though, I still think that my contribution to the elf parsing code will be useful for many users of your library.

Let me know if there are any comments.


zlobober:
The build fails, I guess, because I used python2-specific "print >>sys.stderr" instead of "sys.stderr.write". Let me fix that.

eliben:
Thanks - I hope to find time to review this in the next few days

elftools/elf/structs.py
```diff
@@ -43,9 +43,33 @@ def __init__(self, little_endian=True, elfclass=32):
         assert elfclass == 32 or elfclass == 64
         self.little_endian = little_endian
         self.elfclass = elfclass
-        self._create_structs()
 
-    def _create_structs(self):
+    def create_basic_structs(self):
```
eliben: I'm not sure the split to _create_scalar_structs really helps and "scalar" is confusing in this context (at least to me). I suggest just inline it in here: create_basic_structs will set up all the word size stuff and call _create_ehdr.

scripts/readelf.py
```diff
@@ -568,7 +572,9 @@ def display_hex_dump(self, section_spec):
         """
         section = self._section_from_spec(section_spec)
         if section is None:
-            self._emitline("Section '%s' does not exist in the file!" % (
+            # readelf prints the warning to stderr. Even though stderrs are not compared
+            # in tests, we comply that behavior.
```
eliben: we comply with that behavior.

zlobober:
Sorry for the delay. I think I fixed everything you mentioned.
I also introduced a separate structure for NT_PRPSINFO core note (it contains the most important meta-information about process that produced the core dump) and added unittest that checks all the values for an attached core file.

PR #148, opened 2017-06-19
Update reserved machine arch idents

SkyLeach:
It seems to have been a while since these were updated.

I updated the dict from here: http://www.sco.com/developers/gabi/latest/ch4.eheader.html

Please look this over and merge if you approve.


eliben:
This looks good but the build failed - can you please fix?

SkyLeach:
repeated a keyword value, tests should complete now.  :-)

PR #149, opened 2017-07-05
Add OrderedDict implementation for seamless Python < 2.7 "support"

mzpqnxow:
The ```OrderedDict``` class was added to ```collections``` in Python 2.7 which means pyelftools will not work on older versions of Python. My target was 2.6.

By adding this small ```OrderedDict``` implementation, the majority (possibly all?) of pyelftools works on Python 2.6, and possibly earlier versions.

An alternative is to pull in ordereddict from PyPi and have that as a dependency but this seems simpler and shouldn't be burdensome to maintain, should you want to accept it.

Merge this if you'd like, otherwise I will maintain my own fork as I currently need this supported on 2.6

Thanks, I appreciate pyelftools quite a bit. If you'd like anything tested I'd be happy to help

### DIFF
--- a/elftools/common/legacy.py
+++ b/elftools/common/legacy.py
@@ -1,0 +1,128 @@
+# Copyright (c) 2009 Raymond Hettinger
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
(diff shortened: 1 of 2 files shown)


eliben:
Thanks! I'd prefer to keep it 2.7+ at this point, but this PR can keep open for a while if folks want to apply it in their forks

mzpqnxow:
I figured as much- thanks, hope it can be helpful to someone, though
hopefully anyone working low level binaries probably can figure out in
their own how to fix the error thrown on 2.5/2.6. You can close it now if
you'd like.

PR #152, opened 2017-08-12
Compressed sections

pmderodat:
Hello,

Assemblers from recent binutils releases compress debug sections, using the `SHF_COMPRESSED` section flag (see http://www.sco.com/developers/gabi/latest/ch4.sheader.html). When pyelftools tries to parse DWARF info in such a section, it crashes in a mysterious way:

```
  File "/usr/lib/python3.6/site-packages/elftools/elf/elffile.py", line 174, in get_dwarf_info
    relocate_dwarf_sections)
  File "/usr/lib/python3.6/site-packages/elftools/elf/elffile.py", line 402, in _read_dwarf_section
    section_stream, reloc_section)
  File "/usr/lib/python3.6/site-packages/elftools/elf/relocation.py", line 124, in apply_section_relocations
    self._do_apply_relocation(stream, reloc, symtab)
  File "/usr/lib/python3.6/site-packages/elftools/elf/relocation.py", line 177, in _do_apply_relocation
    stream_pos=reloc['r_offset'])
  File "/usr/lib/python3.6/site-packages/elftools/common/utils.py", line 34, in struct_parse
    raise ELFParseError(str(e))
elftools.common.exceptions.ELFParseError: expected 4, found 0
```

This set of commits enhances sections handling to match compressed sections and decompress them when reading their data, fixing crashes such as the above.

Thank you in advance for reviewing, and thank you for this great project! :-)


elftools/elf/sections.py
```diff
@@ -19,16 +22,83 @@ class Section(object):
          > sec = Section(...)
          > sec['sh_type']  # section type
     """
-    def __init__(self, header, name, stream):
+    def __init__(self, header, name, stream, structs):
```
eliben: Many section types already take `elffile`, why not do the same?

pmderodat: Indeed, why not? :-) I’ll do that, thanks!

pmderodat: I went even further, removing the `stream` constructor argument, which can be replaced by `self.ellfile.stream`. I hope that’s fine by you.

elftools/elf/sections.py
```diff
@@ -19,16 +22,84 @@ class Section(object):
...
+            self._compression_type = header['ch_type']
+            self._decompressed_size = header['ch_size']
+            self._decompressed_align = header['ch_addralign']
+        else:
+            self._decompressed_size = header['sh_size']
+            self._decompressed_align = header['sh_addralign']
+
+    @property
+    def compressed(self):
+        """ Is this section is compressed?
```
eliben: Double "is"

elftools/elf/structs.py
```diff
@@ -153,6 +154,16 @@ def _create_shdr(self):
             self.Elf_xword('sh_entsize'),
         )
 
+    def _create_chdr(self):
+        fields = [
```
eliben: Can you add a comment here that points to the best documentation you know of about this compression?

I don't see this documented in any of the ELF specs I have

test/test_compressed_support.py
```diff
@@ -0,0 +1,97 @@
+#-------------------------------------------------------------------------------
+# elftools tests
```
eliben: Explain what this test is doing here

test/testfiles_for_unittests/compressed.c
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
```
eliben: Add a comment here saying how you produce the .o files from this file (exact gcc/compiler command line)

test/test_compressed_support.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# Pierre-Marie de Rodat (pmderodat@kawie.fr)
+# This code is in the public domain
```
eliben: Also, does readelf support this? I wonder if it's possible to have a readelf-comparison test for a file with compressed sections

eliben:
Thanks, this looks very good overall - just a few questions

elftools/elf/structs.py
pmderodat: I used my distrib’s /usr/include/elf.h, which seems to come from glib: https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/elf.h;h=3900b4c9f0cad606a7beb2511c0bb6d603e885ee;hb=HEAD#l489 Is this good enough? I can also find http://www.sco.com/developers/gabi/latest/ch4.sheader.html, but there’s no anchor link for `SHF_COMPRESSED` and I’m not sure about the perenity of the URL.

test/test_compressed_support.py
pmderodat: I think it’s possible, as though you need to have a fairly recent readelf version: it looks like this feature has been added in readelf in May 2015: https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=commit;f=binutils/readelf.c;h=0e602686df5677fee06cbd1718b4a7aa5379cd2a

pmderodat: I just noticed that the repository embeds a pre-built readelf program. Unfortunately, it seems too old to handle compressed sections:

```shell
$ ../external_tools/readelf --version|head -n1
GNU readelf (GNU Binutils) 2.23.52.20130918

$ ../external_tools/readelf -wi compressed_64.o.elf 
readelf: Warning: skipping invalid relocation offset 0x1af in section .debug_info
readelf: Warning: skipping invalid relocation offset 0x1bc in section .debug_info
[…]
Contents of the .debug_info section:

  Compilation Unit @ offset 0x0:
   Length:        0x1 (32-bit)
   Version:       0
   Abbrev Offset: 0x0
   Pointer Size:  0
readelf: Warning: CU at offset 0 contains corrupt or unsupported version number: 0.
```

Could we install a newer version?

elftools/elf/structs.py
pmderodat: Ok, I also found this: <https://docs.oracle.com/cd/E53394_01/html/E54813/section_compression.html> I see this documentation is already referenced in `structs.py:_create_gnu_verneed`, although it seems that the section numbering scheme have changed: the Object File Format is now Chapter 13 whereas it used to be Chapter 7. I’ll add a commit to update these.

pmderodat:
Thank you for the review! I’ve fixed as much as I could: what remains is the pending question on whether `test/external_tools/readelf` could be updated so we can add a readelf testcase.

eliben:
Something went weird in the commits - I can no longer see the original changes of removing structs and having elffile as an argument, can you please check?

Regarding readelf - yep, sounds like it's time to update the version of the bundled binary. I think this can be done in a separate PR, actually. If you want to do this, that'd be great :) Since readelf annoyingly changes formatting from release to release it may take a bit of tweaking in pyelftools's own readelf and the testing harness but it shouldn't be a big deal.

eliben:
Yeah, it's all messed up :/ I see the commit link you're posted but don't see where I can mix it with other latests commits on GH

Can you try to squash your whole branch into a single commit so I can give this a final go-over before merging?

pmderodat:
Hello Eli,

Looking at the page that shows the squashed changes (https://github.com/eliben/pyelftools/pull/152/files), I realized that I somehow messed with my own commits: there was some part of the `Section` refactoring that was not in the PR. I just re-made them and pushed it.

I’m not sure what do you mean when you say mix it with other latests commits: do the page for the above link give you the information you were looking for?

Now moving to the readelf upgrade… sorry for the noise!

eliben:
Thanks, merged now!

Looking forward to the readelf upgrade PR :)

pmderodat:
Great, thank you! :-)

I’m making progress on the readelf upgrade, but there’s at least one thing not trivial: we need to handle at least one more section (`.eh_frame`) which, if http://www.airs.com/blog/archives/460 is to be believed, is not exactly the same as `.debug_frames`. Hopefully both will share most code.

eliben:
Handling `.eh_frame` would be great but it's a whole new can of worms :) Why is this necessary for upgrading readelf?

pmderodat:
With recent GNU binutils releaseses, `readelf --debug-dump=frames` now displays the content of the `.eh_frame` section, so I thought that in order to keep the diff between this output and `readelf.py`'s we would need to handle them as well. I might miss something, but I don’t think it’s that much a can of worm. Anyway, I’ll keep you updated. :-)

PR #155, opened 2017-08-25
Add parsing and readelf dumping for .eh_frame

pmderodat:
Hello Eli,

Unlike what I thought on #152, reading and dumping `.eh_frame` is not necessary in order to update the embedded `readelf`.

The diff I got while updating was due to some minor change in output, not to some new dump for `.eh_frame` itself. I noticed that while removing the `.eh_frame` filter in `run_readelf_tests.py`, but by that time I had already implemented most of it, so… here it is, better not throw that away. ;-)

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -9,7 +9,9 @@
 import copy
 from collections import namedtuple
 from ..common.utils import (struct_parse, dwarf_assert, preserve_stream_pos)
-from ..common.py3compat import iterkeys
+from ..common.py3compat import iterbytes, iterkeys
+from ..construct import Struct, Switch
+from .enums import DW_EH_encoding_flags
 from .structs import DWARFStructs
 from .constants import *
 
@@ -17,9 +19,17 @@ from .constants import *
 class CallFrameInfo(object):
     """ DWARF CFI (Call Frame Info)
 
+    Note that this also supports unwinding information as found in .eh_frame
+    sections: its format differs slightly from the one in .debug_frame. See
+    <http://www.airs.com/blog/archives/460>.
+
         stream, size:
             A stream holding the .debug_frame section, and the size of the
             section in it.
+
+        address:
+            Virtual address for this section. This is used to decode relative
+            addresses.
 
         base_structs:
             The structs to be used as the base for parsing this section.
@@ -34,9 +44,11 @@ class CallFrameInfo(object):
             such as guessing which CU contains which FDEs (based on their
             address ranges) and taking the address_size from those CUs.
     """
-    def __init__(self, stream, size, base_structs):
+    def __init__(self, stream, size, address, base_structs,
+                 for_eh_frame=False):
         self.stream = stream
         self.size = size
+        self.address = address
         self.base_structs = base_structs
         self.entries = None
 
@@ -44,6 +56,11 @@ class CallFrameInfo(object):
         # offset. Useful for assigning CIE to FDEs according to the CIE_pointer
         # header field which contains a stream offset.
         self._entry_cache = {}
+
+        # The .eh_frame and .debug_frame section use almost the same CFI
+        # encoding, but there are tiny variations we need to handle during
+        # parsing.
+        self.for_eh_frame = for_eh_frame
 
     def get_entries(self):
         """ Get a list of entries that constitute this CFI. The list consists
@@ -74,6 +91,10 @@ class CallFrameInfo(object):
 
         entry_length = struct_parse(
             self.base_structs.Dwarf_uint32(''), self.stream, offset)
+
+        if self.for_eh_frame and entry_length == 0:
+            return ZERO(offset)
+
         dwarf_format = 64 if entry_length == 0xFFFFFFFF else 32
 
         entry_structs = DWARFStructs(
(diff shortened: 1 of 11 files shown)


elftools/dwarf/callframe.py
```diff
@@ -10,17 +10,27 @@
...
+from .enums import DW_EH_encoding_flags
 from .structs import DWARFStructs
 from .constants import *
 
 
 class CallFrameInfo(object):
     """ DWARF CFI (Call Frame Info)
 
+    Note that this also supports unwinding information as found in .eh_frame
+    sections: its format differs lightly from the one in .debug_frame. See
```
eliben: s/lightly/slightly/ ?

elftools/dwarf/callframe.py
```diff
@@ -193,6 +231,171 @@ def _parse_instructions(self, structs, offset, end_offset):
...
+                    enc: fld_cons('function')
+                    for enc, fld_cons
+                    in self._eh_encoding_to_field(entry_structs).items()})),
+        }
+
+        # Build the Struct we will be using to parse the augmentation data.
+        # Stop as soon as we are not able to match the augmentation string.
+        fields = []
+        aug_dict = {}
+        for i in range(len(augmentation)):
```
eliben: Any reason to not just iterate over elements of augmentation here?

elftools/dwarf/callframe.py
```diff
@@ -193,6 +231,171 @@ def _parse_instructions(self, structs, offset, end_offset):
...
+                break
+
+            if fld is True:
+                aug_dict[fld] = True
+            else:
+                fields.append(fld)
+
+        # Read the augmentation twice: once with the Struct, once for the raw
+        # bytes. Read the raw bytes last so we are sure we leave the stream
+        # pointing right after the augmentation.
```
eliben: Can the comment further explain this logic? Won't reading with Struct leave it in the right place? You need both anyway, right?

elftools/dwarf/callframe.py
```diff
@@ -193,6 +231,171 @@ def _parse_instructions(self, structs, offset, end_offset):
...
+        if encoding_modifier == 0:
+            pass
+
+        elif encoding_modifier == DW_EH_encoding_flags['DW_EH_PE_pcrel']:
+            # Start address is relative to the address of the
+            # "initial_location" field.
+            result['initial_location'] += (
+                self.address + initial_location_offset
+            )
+
```
eliben: extra newline

elftools/dwarf/callframe.py
```diff
@@ -225,13 +428,16 @@ class CFIEntry(object):
         offset: the offset of this entry from the beginning of the section
         cie: for FDEs, a CIE pointer is required
```
eliben: Can you document the augmentation_* fields here? With a link to the best online resource, if possible

scripts/readelf.py
```diff
@@ -1064,12 +1102,18 @@ def _dump_debug_frames_interp(self):
                         fullhex=True, lead0x=False)))
                 ra_regnum = entry.cie['return_address_register']
 
+            else: # ZERO terminator
+                assert isinstance(entry, ZERO)
```
eliben: I'm not sure about the name ZERO here; maybe END_MARKER? Is this dictated by the output of readelf?

pmderodat:
> Thanks, really nice to have this! It has always irritated me how poorly documented this eh_frame business is.

Sure! I never had to deal with `.eh_frame` nor `.debug_frames`, so that was instructive. Indeed we’re lucky Ian Lance Taylor wrote these blog posts…

> What about testing? Does removing the eh_ filter in the test runner actually make it run on several eh_frame sections in the test files? Can you please verify, and do you think additional files are needed to exercise this functionality?

Without the change in `run_readelf_tests.py`, the testsuite fails quite explicitely on multiple test:
```sh
$ git checkout HEAD^ -- test/run_readelf_tests.py
$ test/run_readelf_tests.py -k
Test file 'test/testfiles_for_readelf/core_simple64.elf'
Test file 'test/testfiles_for_readelf/exe_compressed64.elf'
.......................FAIL
....for option "--debug-dump=frames"
@@ Number of lines different: 46 vs 108
@@ Output #1 dumped to file: /tmp/out1_ajpu0y47.stdout
@@ Output #2 dumped to file: /tmp/out2_rg84x7ad.stdout
[…]
Conclusion: FAIL (10/18)
```

So I don’t think we need extra testcases: the already present ELF files already have a bunch of `.eh_frame` sections of their own. My work was based on these, actually. Now I’m no expert in this area, so I’m not sure if we have realistic format coverage, here: maybe more `DW_EH_PE_*` constants appear “in the nature”, but it’s hard to know.

elftools/dwarf/callframe.py
pmderodat: Yes, will fix, thanks!

elftools/dwarf/callframe.py
pmderodat: I did it that way to avoid Python2/3 compat issues: iterating on `str` in Python2 gives 1-sized strings whereas iterating on `bytes` in Python3 gives integers. So I went to the most simple alternative. I hesitated with adding some helper in `py3compat.py`: I can do that if you think it’s better.

elftools/dwarf/callframe.py
pmderodat: First, we indeed need both the bytes for readelf dump and the parsed struct to extract information (such as the `FDE_encoding` field). Besides, if the augmentation is not totally supported (unknown char in it), then the Struct will be incomplete: trailing fields will be missing and thus we need to read raw bytes to get past the augmentation data. I will add these details to the comment.

elftools/dwarf/callframe.py
pmderodat: I guess you mean the newline before the closing paren. Will fix!

elftools/dwarf/callframe.py
pmderodat: Ok. I can’t find much information about augmentation in the DWARF spec, so I’ll refer to Ian’s blog.

scripts/readelf.py
pmderodat: ZERO is indeed what `readelf` displays:

```sh
$ readelf --debug-dump=frames /bin/ls|tail -n7
  DW_CFA_nop
  DW_CFA_nop
  DW_CFA_nop

00002cf0 ZERO terminator
```

As it’s not present in DWARF and as there is no documentation for this, I would suggest we stick to `readelf`’s naming, to reduce the possible confusion. What do you think?

elftools/dwarf/callframe.py
eliben: Got it, thanks

elftools/dwarf/callframe.py
eliben: Also after; in general in Python there's no need to have newlines before `elif` and `else`, unless that really makes code unreadable. The default is less whitespace in these location. Not a big deal either way

scripts/readelf.py
eliben: Sounds good, but please beef up the comment in the place where it's defined to explain the name choice

eliben:
SGTM re the testing. Also note a test is failing on Python 2 now

pmderodat:
Thank you for the Python2 heads-up; now tested with both Python2.7 and Python3! As the bytes iteration discrepancy bit me twice in `readelf.py`, I went ahead and introduced a new helper in `py3compat.py`: `iterbytes`.

PR #158, opened 2017-10-02
Wheel support

dsch:
Advantages of wheels
- Faster installation for pure python packages.
- Avoids arbitrary code execution for installation. (Avoids setup.py)
- Allows better caching for testing and continuous integration.
- Creates .pyc files as part of installation to ensure they match the python interpreter used.
- More consistent installs across platforms and machines.

### DIFF
--- a/TODO
+++ b/TODO
@@ -17,14 +17,14 @@ do so in the future.
 Distribution
 ------------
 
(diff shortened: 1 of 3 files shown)


eliben:
Thanks!

Issue #138, opened 2017-03-16
Inconsistent condition between #address_offsets in elffile.py and in readelf.c


david942j:
In method [ELFFile#address_offsets](https://github.com/eliben/pyelftools/blob/8cd87bec6f69102726a50fecbd8f5748a7cd3a75/elftools/elf/elffile.py#L131)
```python
if (start >= seg['p_vaddr'] and
   end <= seg['p_vaddr'] + seg['p_filesz']):
```

According to binutils/readelf.c#offset_from_vma:
```cpp
static long
offset_from_vma (FILE * file, bfd_vma vma, bfd_size_type size)
{
  Elf_Internal_Phdr * seg;

  if (! get_program_headers (file))
  {
      warn (_("Cannot interpret virtual addresses without program headers.\n"));
      return (long) vma;
  }

  for (seg = program_headers;
       seg < program_headers + elf_header.e_phnum;
       ++seg)
  {
      if (seg->p_type != PT_LOAD)
        continue;

      if (vma >= (seg->p_vaddr & -seg->p_align)
          && vma + size <= seg->p_vaddr + seg->p_filesz)
        return vma - seg->p_vaddr + seg->p_offset;
  }

  warn (_("Virtual address 0x%lx not located in any PT_LOAD segment.\n"),
        (unsigned long) vma);
  return (long) vma;
}
```

So seems the condition in `address_offsets` should be
```python
if (start >= seg['p_vaddr'] & -seg['p_align'] and
   end <= seg['p_vaddr'] + seg['p_filesz']):
```
Not sure what this difference affects, just figure this out.

david942j:
In addition, `address_offsets` should add condition that only consider `LOAD` segment:
```python
for seg in self.iter_segments():
    if seg['p_type'] != 'PT_LOAD':
        continue
    if (start >= seg['p_vaddr'] and
        end <= seg['p_vaddr'] + seg['p_filesz']):
        yield start - seg['p_vaddr'] + seg['p_offset']
```

eliben:
Thanks. It would be great if you had a test case that demonstrates this

david942j:
I sent a pr for fixing the second comment.

For the original issue, I don't know why readelf did that (and for me that behavior seems no sense)

This issue can be closed if #159 got merged. 😄

PR #159, opened 2017-10-10
Fix - ELFFile#address_offsets should consider PT_LOAD only

david942j:
As mentioned in #138 (the second comment)
`ELFFile#address_offsets` should consider PT_LOAD segments only,
otherwise wrong addresses might be yielded.

For example, the readelf output by ```$ readelf -l `which cat` ```:
```
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000000040 0x0000000000000040
                 0x00000000000001f8 0x00000000000001f8  R E    0x8
  INTERP         0x0000000000000238 0x0000000000000238 0x0000000000000238
                 0x000000000000001c 0x000000000000001c  R      0x1
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x000000000000787c 0x000000000000787c  R E    0x200000
  LOAD           0x0000000000007a88 0x0000000000207a88 0x0000000000207a88
                 0x0000000000000638 0x00000000000007d8  RW     0x200000
```

and python script:
```python
import elftools.elf.elffile
f = elftools.elf.elffile.ELFFile(open("/bin/cat"))
print(list(f.address_offsets(0x40)))
```
Expect: `[64]`
Output: `[64, 64]`

This is because `address_offsets` take the `PHDR` segment into consideration (shouldn't be).
This pr fixed this behavior.

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -136,7 +136,9 @@ class ELFFile(object):
         """
         end = start + size
         for seg in self.iter_segments():
-            if (seg['p_type'] != 'PT_LOAD'): continue
+            # consider LOAD only to prevent same address being yielded twice
+            if seg['p_type'] != 'PT_LOAD':
+                continue
             if (start >= seg['p_vaddr'] and
                 end <= seg['p_vaddr'] + seg['p_filesz']):
                 yield start - seg['p_vaddr'] + seg['p_offset']
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -136,6 +136,7 @@ class ELFFile(object):
         """
         end = start + size
         for seg in self.iter_segments():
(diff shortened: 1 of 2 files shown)


elftools/elf/elffile.py
```diff
@@ -136,6 +136,7 @@ def address_offsets(self, start, size=1):
         """
         end = start + size
         for seg in self.iter_segments():
+            if (seg['p_type'] != 'PT_LOAD'): continue
```
eliben: Please move `continue` to the next line; parens around condition aren't necessary; and also add a comment explaining this condition

david942j: Commited :)

Issue #75, opened 2015-05-31
Ability to read the full list of ARM build attributes such as Tag_CPU_arch Tag_CPU_name from the .ARM.attributes section


chr15m:
There doesn't seem to be a way to access these attributes for ARM ELF binaries.

http://llvm.org/docs/doxygen/html/ARMBuildAttributes_8h_source.html

The readelf tool does the correct thing here:
`readelf -A ...` where ... is some ARM ELF binary will show these tags.

chr15m:
Here is the patch to the 'readelf' command that has the details on how they do this:

http://www.sourceware.org/ml/binutils/2005-09/msg00131.html

eliben:
I suppose you could still use pyelftools to get the raw section from the ELF file; it's just that pyelftools wouldn't decode it for you?

I'd accept a patch/pull request that implements this.

chr15m:
That's what I've done for now, thank you. Will submit a pull request if I get around to patching pyelftools itself. Thanks for the excellent software!

sevaa:
@chr15m is this still an issue?

chr15m:
I no longer work on the software where this was required. Maybe @umlaeute knows.

sevaa:
Well, the ability to parse `.ARM.attributes` has been in the library for a while, so barring a more definitive answer I'd suggest this issue has been fixed...

umlaeute:
i don't fully remember the details, but afaik we have been able to extract all the details we are interested in from ARM binaries for ages (but probably we've just been happy with the manual decoding). it's code that we seldom touch as it  *just works*.

sevaa:
Good to know. Let's close the issue. I can't.

PR #160, opened 2017-11-16
ARM build attributes section parsing

tyb0807:
Hi Eli,

This answers https://github.com/eliben/pyelftools/issues/75, please tell me if it's ok.

Thank you


elftools/elf/enums.py
```diff
@@ -16,20 +16,23 @@
     ELFCLASS64=2
 )
 
+
```
eliben: These added newlines don't look necessary...

elftools/elf/structs.py
```diff
@@ -347,3 +358,62 @@ def _create_stabs(self):
...
+        )
+
+        # Structure of a build attribute tag.
+        self.Elf_Attribute_Tag = Struct('Elf_Attribute_Tag',
+                                        Enum(self.Elf_uleb128('tag'),
+                                             **ENUM_ATTR_TAG_ARM)
+        )
+
+
+def _LEB128_reader():
```
eliben: I'm not sure I understand why this is needed. Can you point to a place where this format is documented?

tyb0807:
Hello,

Sorry I was in a rush yesterday, here is the doc I've used to implement this http://infocenter.arm.com/help/topic/com.arm.doc.ihi0045e/IHI0045E_ABI_addenda.pdf. I also had a look at `readelf` source code.

elftools/elf/enums.py
tyb0807: From PEP 8,
> Surround top-level function and class definitions with two blank lines.

So I think of these `Enum` as a 'top-level definition'. I remove these then?

tyb0807:
I just pushed some tests. I know that I'm a beginner at Python, so please tell me if the code design is ok. Also, does our `readelf.py` have to produce the exact output as binutils' `readelf` ? I have slightly modified the output to be (I hope) clearer though.

elftools/elf/enums.py
eliben: Yeah, it's not very consistent throughout the code. In this particular file it was all one line. I suggest you don't modify them in your PR as it's unrelated. I can make it more consistent separately later

elftools/elf/structs.py
eliben: OK, now I see in the spec you provided that LEBs are needed -- but why duplicate all this code with the dwarf files? I suggest placing them in common/construct_utils.py

eliben:
> Also, does our readelf.py have to produce the exact output as binutils' readelf

It's be best this way, otherwise the readelf comparison tests fail. There are some exclusions in run_readelf_tests.py, but in general the output should match

tyb0807:
Is it better now?

elftools/elf/descriptions.py
```diff
@@ -144,6 +164,34 @@ def describe_note(x):
...
+    d_entry = _DESCR_ATTR_VAL_ARM[idx]
+
+    if d_entry is None:
+        if tag == 'TAG_COMPATIBILITY':
+            return (_DESCR_ATTR_TAG_ARM[tag]
+                    + 'flag = %d, vendor = %s' % (val, extra))
+
+        elif tag == 'TAG_ALSO_COMPATIBLE_WITH':
+            if val.tag == 'TAG_CPU_ARCH':
+                return _DESCR_ATTR_TAG_ARM[tag] + d_entry[val]
```
pmhahn, line 179: I noticed the following bug while working on #514: `d_enty is None` in line 172, so this will fails:
> Value of type "None" is not indexable  [index]

PR #176, opened 2018-01-05
Don't include 'meta-tag' markers in descriptions of ENUM_D_TAG

### LINKED ISSUE #150: Remove DT_ENCODING from the dynamic tags list
According to [here](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-42444.html) and [here](http://www.sco.com/developers/gabi/2000-07-17/ch5.dynamic.html), DT_ENCODING is not a real value, but rather a marker indicating something about the semantics of all the values lower than it. However, it shares a value with `DT_PREINIT_ARRAY`, which means that when performing a lookup for the tag name of a section, we're at the mercy of the hash function in determining which value we see first. The result is that on some machines, `d_tag` for these entries will be `DT_ENCODING`, while some will be `DT_PREINIT_ARRAY`. This is really bad!

Another possible solution would be to change the "tag lookup" function to just blacklist all the "semantic tags" like DT_ENCODING, DT_LOWOS, etc.

rhelmot:
fixes #150

### DIFF
--- a/elftools/elf/descriptions.py
+++ b/elftools/elf/descriptions.py
@@ -463,30 +463,28 @@ _DESCR_NOTE_ABI_TAG_OS = dict(
     ELF_NOTE_OS_SYLLABLE='Syllable',
 )
 
+def _reverse_dict(d, low_priority=()):
+    out = {}
+    for k, v in iteritems(d):
+        if v in out and k in low_priority:
+            continue
+        out[v] = k
+    return out
 
-_DESCR_RELOC_TYPE_i386 = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_i386))
-
-
-_DESCR_RELOC_TYPE_x64 = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_x64))
-
-
-_DESCR_RELOC_TYPE_ARM = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_ARM))
-
-
-_DESCR_RELOC_TYPE_AARCH64 = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_AARCH64))
-
-
-_DESCR_RELOC_TYPE_MIPS = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_MIPS))
-
-
-_DESCR_D_TAG = dict(
-        (v, k) for k, v in iteritems(ENUM_D_TAG))
-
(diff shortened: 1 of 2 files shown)


elftools/elf/descriptions.py
```diff
@@ -463,30 +463,28 @@ def describe_attr_tag_arm(tag, val, extra):
...
-
-
-_DESCR_RELOC_TYPE_MIPS = dict(
-        (v, k) for k, v in iteritems(ENUM_RELOC_TYPE_MIPS))
-
-
-_DESCR_D_TAG = dict(
-        (v, k) for k, v in iteritems(ENUM_D_TAG))
-
+def _reverse_dict(d, low_priority=()):
```
eliben: Please add a docstring to this function explaining the arguments

elftools/elf/descriptions.py
```diff
@@ -463,30 +463,28 @@ def describe_attr_tag_arm(tag, val, extra):
...
+        out[v] = k
+    return out
+
+_DESCR_RELOC_TYPE_i386 = _reverse_dict(ENUM_RELOC_TYPE_i386)
+_DESCR_RELOC_TYPE_x64 = _reverse_dict(ENUM_RELOC_TYPE_x64)
+_DESCR_RELOC_TYPE_ARM = _reverse_dict(ENUM_RELOC_TYPE_ARM)
+_DESCR_RELOC_TYPE_AARCH64 = _reverse_dict(ENUM_RELOC_TYPE_AARCH64)
+_DESCR_RELOC_TYPE_MIPS = _reverse_dict(ENUM_RELOC_TYPE_MIPS)
+_DESCR_D_TAG = _reverse_dict(ENUM_D_TAG, (
+    # these are 'meta-tags' marking semantics of numeric ranges of the enum
```
eliben: Can you link to the relevant documentation pages here (as you did in the issue)?

rhelmot:
It is, in fact, not possible to add tests to verify that this is working, since the bug it quashes is based on nondeterminsm in python's hash algorithm.

eliben:
@rhelmot I see. I think a test would still be useful, in case this code changes in the future in a way that may break some current use case. WDYT?

rhelmot:
There is now a test, but it is longer than the code it is testing.

eliben:
Yeah, this isn't a great test because it's more aimed at how things are done than why they're done.

Since the change is simply and it helps unblock you, I'll merge this. However, consider submitting a better test at some future point.

The best kind of test is a full readelf-comparison test; the second best kind of test is a file in `testfiles_for_unittests` with specific unit testing (like the stuff `test_dynamic.py` is doing)

PR #177, opened 2018-01-19
Do not crash if CIE doesn't define CFA

emersion:
```
Contents of the .eh_frame section:

00000000 0000000000000010 00000000 CIE "zR" cf=1 df=-8 ra=16
   LOC           CFA      
0000000000000000Traceback (most recent call last):
  File "scripts/readelf.py", line 1377, in <module>
    main()
  File "scripts/readelf.py", line 1356, in main
    readelf.display_debug_dump(options.debug_dump_what)
  File "scripts/readelf.py", line 737, in display_debug_dump
    self._dump_debug_frames_interp()
  File "scripts/readelf.py", line 1222, in _dump_debug_frames_interp
    self._dwarfinfo.EH_CFI_entries())
  File "scripts/readelf.py", line 1204, in _dump_frames_interp_info
    self._emit(' %-9s' % describe_CFI_CFA_rule(line['cfa']))
  File "./elftools/dwarf/descriptions.py", line 129, in describe_CFI_CFA_rule
    if rule.expr:
AttributeError: 'NoneType' object has no attribute 'expr'
```


eliben:
Do you happen to have a test case (a simple file) where this crash manifests?

emersion:
Yeah, do you want me to add a test?

Also note that parsing such a file will add an empty line in the CIE. Would you prefer not adding the empty line? (That's what other tools like `readelf` do)

eliben:
Yes, a test would be great. As simple as possible to reproduce the issue (since it will have to pass all the comparisons with readelf).

Not sure I understand the question about the empty line, but in general we want to maintain compatibility with readelf's output to make strong testing possible.

emersion:
Added a new test file and modified the parser to match `readelf`'s behavior.

PR #182, opened 2018-02-22
Convert all ascii decoding to utf-8 decoding

### LINKED ISSUE #173: Crash when non-ascii characters are used in symbols or sonames
At the bottom of every read from memory is the conversion from bytes to string with `.decode('ascii')`. This fails extremely loudly when there's a character >0x7e. Strings like this can occur naturally in e.g. elf files found on android systems. To reproduce just copy libc.so.6 and replace the libc.so.6 soname text with libc\xffso.6 or whatever.

Two possible solutions:
- return bytes instead of string
- replace `s.decode('ascii')` with `''.join(chr(c) for c in s)`

rhelmot:
Closes #173 

Adds a testcase asserting that we can handle a symbol named Δ


eliben:
Looks good, thanks! Seems like the build is failing on 3.2 though, due to unicode literals being added in 3.3 (https://www.python.org/dev/peps/pep-0414/) -- can you fix that please?

rhelmot:
fixed via force push

PR #183, opened 2018-02-22
Choose members of the dynamic tag enum based on the current machine

### LINKED ISSUE #175: Only use os- and processor- specific enum values on the appropriate os and processor
So this is the _actual_ cause that was behind yesterday's issue, though that is probably still a problem.

If you look at how this project tries to load the file linked from this issue https://github.com/angr/cle/issues/103#issuecomment-355405436 it crashes because it thinks the dynamic tag 0x6000000f is `DT_SUNW_FILTER`, and assumes that the corresponding d_val points to a string. However, this is not solaris, and 0x6000000f is not a filter here, so the pointer is not an index into a string table, and so the lookup crashes.

This seems like a very complicated and invasive change, and I don't know anything about the architecture of this project, so I'd appreciate a response other than "patches welcome".

rhelmot:
closes #175 

Adds a testcase and a 35k android marshmallow library (libsoundtrigger.so) for the testcase. The testcase also reuses an existing solaris test binary.


elftools/elf/elffile.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, stream):
 
         self.structs.create_basic_structs()
         self.header = self._parse_elf_header()
-        self.structs.create_advanced_structs(self['e_type'], self['e_machine'])
+        self.structs.create_advanced_structs(self['e_type'], self['e_machine'], self['e_ident']['EI_OSABI'])
```
eliben: Please wrap this for 80-col lines

elftools/elf/enums.py
```diff
@@ -573,9 +582,13 @@
...
-    DT_AUXILIARY=0x7ffffffd,
-    DT_FILTER=0x7fffffff,
-    _default_=Pass,
+)
+
+ENUMMAP_EXTRA_D_TAG_MACHINE = dict(
+    EM_MIPS=ENUM_D_TAG_MIPS,
+    EM_MIPS_RS3_LE=ENUM_D_TAG_MIPS,
+    # TODO: add the rest
+    # solaris doesn't go here because what matters is e_ident[EI_OSABI] == ELFOSABI_SOLARIS
```
eliben: I'm not sure I follow this comment. What are you trying to say? It may be obvious that we now have a separate section for the solaris tags, so maybe it's superfluous?

test/test_dynamic.py
```diff
@@ -68,6 +68,26 @@ def test_reading_symbols(self):
...
+                            tag.entry.d_tag.startswith("DT_SUNW"):
+                        seen.add(tag.entry.d_tag)
+
+            return seen
+
+        f1 = extract_sunw(os.path.join('test', 'testfiles_for_unittests',
+            'exe_solaris32_cc.sparc.elf'))
+        f2 = extract_sunw(os.path.join('test', 'testfiles_for_unittests',
+            'android_dyntags.elf'))
+        self.assertEqual(f1, {'DT_SUNW_STRPAD', 'DT_SUNW_LDMACH'})
```
eliben: Nice testing!

Do you think it would be possible to add readelf-comparison tests for these files?

eliben:
Thanks for the contribution, this looks great overall. I left some (minor) comments in the review.

rhelmot:
Fix pushed. I tried running readelf.py on the test files and I actually found a bug with the PR, so that's been resolved (along with the other changes).

However, when I tried adding a readelf-comparison test, it failed extremely badly for a lot of reasons unrelated to this PR, notably the naming of solaris-specific sections and `readelf -d` outputting a second hunk of section data after the listing of the dynamic tags.

PR #192, opened 2018-07-20
readelf: support --wide option

frederiksdun:
I had some issues for readelf with testfiles filename > 76 chars

What do you think?


eliben:
Does this match readelf behavior?

frederiksdun:
Yes it does: https://github.com/eliben/pyelftools/pull/192/commits/17874463f969cb8fbe36f6c10202dabdbcb03f4e#diff-7629bf4e191a8003a69d29ebe7604cf4R73

But some whitespace issue still need to be addressed in output

I just wanted to have your sight on this first.

eliben:
I don't object if the change doesn't end up to be too complicated since it's a fringe use case and mostly outside the core capability of `pyelftools`.

eliben:
Closing due to inactivity. Reopen PR if needed.

PR #194, opened 2018-08-17
Call relocation for ARM

dimas3452:
Call instruction relocating for ARM requires a specific calculation function.
This patch implements it.
I tested this on ARMv5 architecture.

Signed-off-by: Koltunov Dmitry <koltunov@ispras.ru>

### DIFF
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -219,10 +219,16 @@ class RelocationHandler(object):
     def _reloc_calc_sym_plus_addend_pcrel(value, sym_value, offset, addend=0):
         return sym_value + addend - offset
 
+    def _arm_reloc_calc_sym_plus_value_pcrel(value, sym_value, offset, addend=0):
+        return sym_value / 4 + value - offset / 4
+
     _RELOCATION_RECIPES_ARM = {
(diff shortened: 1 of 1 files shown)


dimas3452:
Existing test files do not cover it.
I tested it by relocating the relocatable object file for ARM using `RelocationHandler`. I extracted the contents of the sections that need relocation and relocated it. The contents of the relocated sections were loaded into the memory of the QEMU emulator and launched into QEMU with the debugger connected.
I think, i can write a usage example.

dimas3452:
## V2
* Add test to verify support for 'R_ARM_CALL' relocation type

dimas3452:
## V3
* Add `common.py3compat` module usage
* Move test ELF files into `test/testfiles_for_unittests`
* Add a comment for `test/testfiles_for_unittests/simple.c` that explains how to compile it to get the included output files
* Add a comment for `test/run_arm_reloc_call_test.py ` that explains what this test does
* Change extension for test ELF files

eliben:
@dimas3452 I don't think your test actually ran in CI because it was not discovered. I reorganized the code a bit to be more consistent with other tests and your test is now in https://github.com/eliben/pyelftools/blob/master/test/test_arm_call_reloc.py -- it fails when run with Python 3, can you please check?

dimas3452:
@eliben I found the cause of failure. In Python 3 `/` operator does a floating-point division.

PR #196, opened 2018-09-05
Fixup for arm call relocation

dimas3452:


eliben:
This works, thanks for the quick fix!

Issue #189, opened 2018-05-07
Expand DT_FLAGS or provide flags enum


yugr:
`DT_FLAGS` dynamic tag is a bitmask which may contain any combination of `DF_ORIGIN`, `DF_SYMBOLIC`, etc. flags. Would be nice if pyelftools expanded `DT_FLAGS` to a set of flags or at least provided enum with values of flags.

PR #200, opened 2018-10-02
Provide enums for DT_FLAGS and DT_FLAGS_1

rupran:
This change adds two enums with the name to value mappings
for the two flags fields in the dynamic section. The values
and corresponding names are taken from the elf/elf.h file
in the most recent glibc version.

The enums are also used to print the names instead of the
raw hex values for DT_FLAGS and DT_FLAGS_1 in
scripts/readelf.py.

Fixes: #189


scripts/readelf.py
```diff
@@ -425,6 +426,10 @@ def display_dynamic_tags(self):
                     parsed = 'Library soname: [%s]' % tag.soname
                 elif tag.entry.d_tag.endswith(('SZ', 'ENT')):
                     parsed = '%i (bytes)' % tag['d_val']
+                elif tag.entry.d_tag == 'DT_FLAGS':
+                    parsed = describe_dt_flags(tag.entry.d_val)
```
eliben: No "Flags: " needed here?

eliben:
Thanks for the contribution! It'd be great to have a test here

scripts/readelf.py
rupran: That's what my system's `readelf` prints, so I figured I'd add it here as well.

rupran:
I can add a test case, that would mean adding a binary with some of the flags enabled, right?

scripts/readelf.py
rupran: Ah, sorry, I just realized I read your comment wrong. For DT_FLAGS, `readelf` only shows the respective flag names, while for DT_FLAGS_1, the additional "Flags: " is printed... This is also the case in the `readelf` binary shipped with the tests.

PR #203, opened 2018-10-24
ARMAttribute: fix access to structs, stream and nul

rupran:
The __init__ function of ARMAttribute has two parameters
structs and stream through which the caller can pass in the
relevant objects (ARMAttributesSubsubsection does that after
seeking to the right position in stream).

The accesses for TAG_SECTION and TAG_SYMBOL, however, were
referring to non-existing members instead of the parameters.

Additionally, one assertion tries to access an undefined
'null_byte' variable which should be 'nul' instead.

Unfortunately, I don't have an ARM system to generate a
testcase binary, but maybe somebody else can contribute that.

Fixes: #160

### DIFF
--- a/elftools/elf/sections.py
+++ b/elftools/elf/sections.py
@@ -287,14 +287,12 @@ class ARMAttribute(object):
 
             if self.tag != 'TAG_FILE':
                 self.extra = []
-                s_number = struct_parse(self.structs.Elf_uleb128('s_number'),
-                                        self.stream
-                           )
+                s_number = struct_parse(structs.Elf_uleb128('s_number'), stream)
 
                 while s_number != 0:
                     self.extra.append(s_number)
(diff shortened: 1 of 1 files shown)


eliben:
Thanks. It would be really great to have some tests for this. Can you compile a simple binary to ARM (cross-compile, I don't believe you need an actual ARM system) that would manifest these issues?

rupran:
According to the ARM ABI specification [[0]], Section 2.2.4, the Tag_Section and Tag_Symbol attributes are deprecated and optional since the 2.09 release which was nearly 6 years ago. I haven't figured out if a modern gcc can still emit those attributes but I haven't found an obvious command line switch or similar documentation.

[0]: http://infocenter.arm.com/help/topic/com.arm.doc.ihi0045e/IHI0045E_ABI_addenda.pdf

PR #204, opened 2018-10-26
Remove unnecessary 'preserve_stream_pos'

andersdellien:
The stream position in the .debug_info stream can't change when
reading from the .debug_abbrev stream.


andersdellien:
(Very) minor performance improvement, around 2-3% of execution time is spent in that context manager.

Tested with the unit test files as well as some larger files.

What do you think?

PR #206, opened 2018-11-21
Implemented ELFFile.get_machine_arch for the remaining architectures.

gahag:
Added all architectures according to the ENUM_E_MACHINE.


elftools/elf/elffile.py
```diff
@@ -211,18 +211,194 @@ def get_dwarf_info(self, relocate_dwarf_sections=True):
...
+            'EM_Z80'           : 'Zilog Z80',
+            'EM_VISIUM'        : 'VISIUMcore',
+            'EM_FT32'          : 'FTDI Chip FT32 32-bit RISC',
+            'EM_MOXIE'         : 'Moxie',
+            'EM_AMDGPU'        : 'AMD GPU',
+            'EM_RISCV'         : 'RISC-V'
+        }
+
+        if self['e_machine'] in architectures:
+            return architectures[self['e_machine']]
```
eliben: It could be a bit cleaner to use the dict .get method with a default of '<unknown>' here?

gahag: Thanks! My python is a bit rusty.

PR #207, opened 2018-12-06
Fix parsing issues in .ARM.attributes section

tbp1776:
* Fix seeking to the start of each subsection and subsubsection
* Fix concurrency issue when the position of the cursor in the stream
  is changed during the "yield"


eliben:
Closing due to inactivity. Reopen PR if needed.

PR #209, opened 2018-12-15
Simplify handling of null DIEs

andersdellien:
The code that is intended to coalesce null DIEs into the DIE that
precedes them does not do that and is actually not needed as the
'unflattening' procedure takes care of any unexpected null DIEs.

As the behavior will be the same both with and without the code
that handles the null DIEs we can remove it.

Also added a unit test that verifies the DIE size calculation.


andersdellien:
Sure, let me try to explain in a bit more detail.

The intention behind the part I removed is that it should take null DIEs and merge them into the previous entry. However, I don't think it works properly (I suppose 'if self.stream.read(1) == 0' should be 'if self.stream.read(1) == b'\00' instead) - so right now the output of the parser when it encounters a sequence of \000 bytes in the stream is a corresponding sequence of null DIEs.

This might potentially be a problem, but actually unflatten_tree handles this situation:

```
                # parentstack should not be really empty here. However, some
                # compilers generate DWARF that has extra NULLs in the end and
                # we don't want pyelftools to fail parsing them just because of
                # this.
                if len(parentstack) > 0:
                    # end of children for the current parent
                    parentstack.pop()
```

As unflatten_tree takes care of any 'extra' null DIEs my reasoning is that there is no need for the code in _parse_DIE that was intended to remove these. So rather than trying to fix it, I simply removed it. 

Does it make sense?

eliben:
Thanks, merging

PR #210, opened 2018-12-16
Bugfix: dwarf_decode_address.py [ Need to reset prevstate when the line number sequence ends]

rvijayc:
There is a minor bug in the dwarf_decode_address.py example. "prevstate" needs to be reset when the line number sequence ends. It doesn't show up with the sample file as it likely has only one compilation unit.


rvijayc:
Looks like I didn't branch at the right point for this pull request. I'll close this and open a new one.

PR #212, opened 2018-12-22
More efficient AbbrevDecl handling

andersdellien:
Create all the AbbrevDecl objects during parsing and later return
references to them - this gives a small performance gain.


eliben:
Can you give more details about the performance gain? What file did you observe it on, and what was the gain?

andersdellien:
The performance gain is small  (but measurable) - about 300 ms out of a total time of 40 seconds. My test file contains about 350.000 DIEs, which also is the number of AbbrevDecl objects created before this patch is applied. With this patch, the number of allocations decreases to about 10.000.

PR #214, opened 2019-01-27
Support for DWARFv4 location lists in dwarf_location_lists.py

andersdellien:
In DWARFv4 the location lists are referenced with the 'sec_offset'
attribute form instead of 'data4' or 'data8'.

### DIFF
--- a/examples/dwarf_location_lists.py
+++ b/examples/dwarf_location_lists.py
@@ -58,7 +58,7 @@ def process_file(filename):
                 # AttributeValue object (from elftools.dwarf.die), which we
                 # can examine.
                 for attr in itervalues(DIE.attributes):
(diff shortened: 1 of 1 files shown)


eliben:
Thanks for the contribution.

In the follow-up comment, are you suggesting updating the example? Or the library logic for reading location lists?

andersdellien:
When I wrote that comment I was mainly thinking of the example code - but it would be nice to refactor the library code to also handle 'DW_FORM_exprloc' attributes (I haven't really thought it through, but it seems doable). So, both. 

Would you be interested in a patch that updates the example/library code along these lines?

eliben:
Yes, that sounds useful

PR #215, opened 2019-01-29
Fixup error on empty .debug_pubtypes section

laerreal:
Sometimes, `-gpubnames` option for gcc results in empty ".debug_pubtypes". section.
That section only contains single zero entry for a CU. This entry is incorrectly handled by
current `NameLUT` implementation.


elftools/dwarf/namelut.py
```diff
@@ -160,9 +160,8 @@ def _get_entries(self):
         offset = 0
         entry_struct = Struct("Dwarf_offset_name_pair",
                 self._structs.Dwarf_offset('die_ofs'),
-                CString('name'))
-        die_ofs_struct = self._structs.Dwarf_offset('die_ofs')
-                
+                If(lambda ctx: ctx['die_ofs'],CString('name')))
```
eliben: Can you add a comment above `entry_struct =` explaining why this conditional parsing exists and what the expected result is when the offset is 0 or not 0

laerreal: Done.

Issue #83, opened 2015-08-25
Parsing 12 Mb of DIE's takes 1 minute


nklacar:
Hi , 

Performance seems pretty poor when parsing an entire DWARF section for an elf. The DWARF size is about 12 meg and parsing takes 1 minute. 

Attaching perf call graph. Commercial debuggers are able to parse the dwarf in < 2 seconds.  

RepeatUntil parse method seems to be eating most the time. Is it the object copying that is taking too long? 

Any suggestion on how to speed up performance? 

![output2](https://cloud.githubusercontent.com/assets/6211792/9456199/4e608e30-4a89-11e5-8818-5eb9753db46b.png)

eliben:
Performance is a known problem with pyelftools, if you need to read the whole info. The binary parsing library pyelftools is built on (construct) isn't designed for speed :-/

nklacar:
If I was to implement construct in C , would thing speed up? I'm trying to isolate where the biggest bottlenecks are... I like this project since its in python and it would be great if I could make it work for my requirements.

eliben:
Reimplementing construct in C would be a huge undertaking.

I'd invest into more laziness in how pyelftools parses DWARF. You very rarely need ALL OF IT, so it makes sense to parse only the parts you *do* need. pyelftools tries to be lazy, but perhaps more laziness is required.

nklacar:
I need to be able to parse all the variables. Il try and make it only parse the variable names and not the types and use lazy parsing for only the variables I need.

eliben:
There've been several changes to optimize parsing and make it lazier over the past few years. Since this issue is very old I'm going to close it. Feel free to open new issues

PR #216, opened 2019-02-01
Lazy parsing

laerreal:
Recently, I used `pyelftools` in a debugger. During my work I faced a problem.
`pyelftools` parses much of debug info when the debugger just looks for a
few things. E.g. when the debugger looks for a DIE describing a variable,
entire CU (all its DIEs) is parsed. When working with a big binary (like
[Qemu](https://www.qemu.org/) emulator) debugger loading consumes about
5 minutes which is very annoying to a user.

The solution I suggest is lazy parsing. This patch series reworks some internals
of the library so it only parses requested things.

Also, because an ELF file is considered constant, there is no a reason to parse
things twice. So, I also added some caches for parsed entities.

### DIFF
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -55,6 +55,8 @@ class CompileUnit(object):
 
         # A list of DIEs belonging to this CU. Lazily parsed.
         self._dielist = []
+        # A list of corresponding DIE offsets.
+        self._diemap = []
 
     def dwarf_format(self):
         """ Get the DWARF format (32 or 64) for this CU
@@ -112,12 +114,14 @@ class CompileUnit(object):
 
         # First pass: parse all DIEs and place them into self._dielist
         die_offset = self.cu_die_offset
+        dm = self._diemap
         while die_offset < cu_boundary:
             die = DIE(
                     cu=self,
                     stream=self.dwarfinfo.debug_info_sec.stream,
                     offset=die_offset)
             self._dielist.append(die)
+            dm.append(die_offset)
             die_offset += die.size
 
         # Second pass - unflatten the DIE tree
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -6,6 +6,7 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from bisect import bisect_left
 from .die import DIE
 
 
@@ -83,6 +84,64 @@ class CompileUnit(object):
         """
         self._parse_DIEs()
         return iter(self._dielist)
+
+    def iter_DIE_children(self, die):
+        """ Given a DIE, yields either its children, without null DIE list
+            terminator, or nothing, if that DIE have no children.
+
+            The null DIE terminator is saved in that DIE when iteration ended.
+        """
+        if not die.has_children:
+            return
+
+        dm = self._diemap
+        dl = self._dielist
+        s = die.stream
+        cu_off = self.cu_offset
+
+        cur_offset = die.offset + die.size
+
+        while True:
+            i = bisect_left(dm, cur_offset)
+            # Note that `dm` cannot be empty because a `die`, the argument,
+            # is already parsed.
+            if i < len(dm) and cur_offset == dm[i]:
+                child = dl[i]
+            else:
+                child = DIE(
+                        cu = self,
+                        stream = s,
+                        offset = cur_offset)
+                dl.insert(i, child)
+                dm.insert(i, cur_offset)
+
+            child.set_parent(die)
+
+            if child.is_null():
+                die._terminator = child
+                return
+
+            yield child
+
+            if not child.has_children:
+                cur_offset += child.size
+            elif "DW_AT_sibling" in child.attributes:
+                sibling = child.attributes["DW_AT_sibling"]
+                cur_offset = sibling.value + cu_off
+            else:
+                # If no DW_AT_sibling attribute is provided by the producer
+                # then the whole child subtree must be parsed to find its next
+                # sibling. There is one zero byte representing null DIE
+                # terminating children list. It is used to locate child subtree
+                # bounds.
+
+                # If children are not parsed yet, this instruction will manage
+                # to recursive call of this function which will result in
+                # setting of `_terminator` attribute of the `child`.
+                if child._terminator is None:
+                    for _ in self.iter_DIE_children(child): pass
+
+                cur_offset = child._terminator.offset + child._terminator.size
 
     #------ PRIVATE ------#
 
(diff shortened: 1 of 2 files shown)


eliben:
This could help address #83

elftools/dwarf/callframe.py
```diff
@@ -148,89 +150,35 @@ def _parse_entry_at(self, offset):
...
+                header=header,
+                stream=self.stream,
+                instructions_start=instructions_start,
+                instructions_end=end_offset,
+                offset=offset,
                 augmentation_bytes=aug_bytes,
                 structs=entry_structs, cie=cie)
         return self._entry_cache[offset]
 
-    def _parse_instructions(self, structs, offset, end_offset):
```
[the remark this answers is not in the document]
laerreal: Probably, because this function has been moved to another class. For lazy instruction parsing it's is required the function to be accessible from `CFIEntry`. But `CFIEntry` object has no reference to `CallFrameInfo`. Actually, that function has not been changed.

laerreal:
Also, I noticed that because of commit rebasing during preparation, GitHub confuses order of commits. So, it's better to use a history view tool that does not rely on timestamps, like `gitk`.

laerreal:
I reworked several comments.

elftools/dwarf/dwarfinfo.py
```diff
@@ -164,31 +164,42 @@ def has_CFI(self):
...
         """
-        cfi = CallFrameInfo(
+        return CallFrameInfo(
             stream=self.debug_frame_sec.stream,
             size=self.debug_frame_sec.size,
             address=self.debug_frame_sec.address,
             base_structs=self.structs)
-        return cfi.get_entries()
+
+    def CFI_entries(self):
```
eliben: It's not clear to me what is gained from this new laziness of CFI entries, for example. If a callsite calls `CFI_entries`, the parsing happens only on the first call, and isn't repeated? But why would `CFI_entries` be called multiple times in the first place?

elftools/dwarf/die.py
```diff
@@ -116,10 +118,26 @@ def get_full_path(self):
...
-        """ Yield all children of this DIE
+        """ Iterates all children of this DIE
+
+            If you sure that all children will be processed then better
+            iterate `children` attribute.
         """
-        return iter(self._children)
+        try:
+            # Try to use lazy `children`.
+            return iter(self.__dict__["children"])
```
eliben: If you have to do this manually anyway, did the `@lazy` decorator help here really?

elftools/common/utils.py
```diff
@@ -102,6 +102,41 @@ def roundup(num, bits):
     """
     return (num - 1 | (1 << bits) - 1) + 1
 
+
+class lazy(object):
```
eliben: I'm not convinced that this magic is really needed, and I would very much prefer to keep the library simple. Lazy parsing of specific constructs can be done with simple code in the specific places where it's required. I left some specific questions in other parts of the code, but the larger design question is - what gets re-parsed unnecessarily and why, and what is the easiest way to prevent that?

elftools/dwarf/dwarfinfo.py
laerreal: That patch only prevents creation of multiple `CallFrameInfo` instances.
Entries parsing is already lazy and its result cached internally in `CallFrameInfo`.
Because previous `*CFI_entries` method created new `CallFrameInfo` instance each time, the caching has not been used actually.

elftools/common/utils.py
laerreal: Yes, lazy parsing with caching can be implemented in any place where it is needed. But it creates some similar code in each the place. I think, it's better to keep such an auxillary code in one place. And then to attach it to a special code using a decorator.

elftools/dwarf/die.py
laerreal: Lazy `children` attribute is mainly added for a user who wants exactly all children of a DIE.
This code serves to another user who wants only few children of a DIE. Because `iter_DIE_children` is slower that a `tuple` iteration, `iter_children` first tries to use lazily constructed `children` but does not forces its construction. It's an optimization only.

eliben:
Unfortunately, I'm not very happy with the PR as it is right now. IMHO it does several slighly related things and adds unnecessary magic.

I would happily accept a patch that adds lazy parsing of DIEs, without magical decorators. The other things can be discussed and added in separate PRs later on

laerreal:
Done.

elftools/dwarf/compileunit.py
```diff
@@ -73,14 +76,84 @@ def get_top_DIE(self):
         """ Get the top DIE (which is either a DW_TAG_compile_unit or
             DW_TAG_partial_unit) of this CU
         """
-        return self._get_DIE(0)
+        dm = self._diemap
```
eliben: Please don't do this, here and elsewhere. Just use `self._diemap` (and `self._dielist`) directly

elftools/dwarf/compileunit.py
```diff
@@ -73,14 +76,84 @@ def get_top_DIE(self):
...
+            terminator, or nothing, if that DIE have no children.
+
+            The null DIE terminator is saved in that DIE when iteration ended.
+        """
+        if not die.has_children:
+            return
+
+        dm = self._diemap
+        dl = self._dielist
+        s = die.stream
```
eliben: Don't use these abbreviations, please

elftools/dwarf/compileunit.py
```diff
@@ -55,6 +56,8 @@ def __init__(self, header, dwarfinfo, structs, cu_offset, cu_die_offset):
 
         # A list of DIEs belonging to this CU. Lazily parsed.
         self._dielist = []
+        # A list of corresponding DIE offsets.
+        self._diemap = []
```
eliben: Can you please add a substantial comment here explaining how the laziness works. What are the invariants for the relevant method calls, and how _dielist and _diemap interplay?

elftools/dwarf/compileunit.py
laerreal: Accessing attributes in [slower][1] than locals. It's significant in a hot code. However, I can use full variable names.

[1]: https://stackoverflow.com/questions/28597014/python-why-is-accessing-instance-attribute-is-slower-than-local

eliben: Yep, I know that. Does it actually affect the profile of DWARF parsing with pyelftools? I find that extremely unlikely given all the other things this code is doing

msuiche:
Out of curiosity, what is the current status on this PR?

laerreal:
I have no time to refactor this now.

woodruffw:
@laerreal @eliben If there aren't any objections, I'd be happy to perform the requested refactors and bring this up-to-date. I can either do that on a new branch + PR or be given push access here, whatever's easier/more comfortable.

eliben:
@woodruffw migrate the subset we agreed upon during the review into a fresh PR

woodruffw:
Will do, thanks!

woodruffw:
Opened #249 based on these changes, with additional feedback addressed.

PR #217, opened 2019-02-15
Also decode strings in _DynamicStringTable.get_string()

rupran:
StringTableSection.get_string() returns an UTF-8 decoded
string (or '' if fetching the string failed) since #182
but the code in _DynamicStringTable was never updated to
decode anything at all so it just returns a bytes sequence
in Python 3.

Let's convert the string there as well to be able to use
both string tables the same way without having to worry
about decoding. Adapt the test cases accordingly.


eliben:
Thanks!

Issue #218, opened 2019-02-21
DynamicSegment.iter_symbols() might terminate early


rupran:
The following code in DynamicSegment.iter_symbols() might return a too low number of entries for the dynamic symbol table.
 https://github.com/eliben/pyelftools/blob/60319cbb9d469e8d7c56e8f85dd55294c0ce4f62/elftools/elf/dynamic.py#L204-L224

It tries to find the next pointer bigger than the beginning of the symbol table (indicated by the 'DT_SYMTAB' dynamic tag) and uses it as the end of the table. For shared libraries or PIE executables with a load offset of 0, **pointers** to e.g. 'DT_SYMTAB' are something like 0x380 which is in a range where tags like 'DT_STRSZ' or 'DT_RELASZ' have their **value**. However, this code does not make any distinction between value and pointer tags (and it seems there is no generic way to do so anyway) so it might use a size value as the (wrong) end of the symbol table.

Other projects (such as https://github.com/chromium/crashpad) use the information in 'DT_HASH' or 'DT_GNU_HASH' to compute the number of elements in the symbol table as every symbol has a hash associated to it (see the change set introducing this at https://github.com/chromium/crashpad/commit/1f1657d573c789aa36b6022440e34d9ec30d894c)

Should I implement something similar here as well? Additionally, we might also want to provide a DynamicSegment.get_symbol(n) function similar to the regular symbol table.

eliben:
Sure, go for it

PR #219, opened 2019-02-21
Improve symbol table handling in DynamicSegment

rupran:
In ultra-stripped binaries the only information about a (dynamic) symbol table comes from the dynamic segment. The DT_SYMTAB tag holds the base address of the table but no tag exists for the number of symbols in the table. The current approach tries to guess this number by looking for the nearest pointer bigger than the DT_SYMTAB pointer in all dynamic tags. However, some of these entries are size values (DT_STRSZ, DT_RELASZ,...) and the heuristic may wrongly interpret them as pointers, which makes DynamicSegment.iter_symbols() return a too low number of symbols.

This change improves the detection by gathering information from the DT_GNU_HASH or DT_HASH symbol hash tables, relying on the fact that every symbol must have a corresponding hash in order for the dynamic loader to properly resolve symbols during loading. The change is inspired by the code used by https://github.com/chromium/crashpad (introduced in [this change]).

Additionally, DynamicSegment only had the iter_symbols() so far, whereas the regular SymbolTableSection also has access to the number of symbols, a specific symbol by index and lookup for symbols by name. I ported these features to DynamicSegment as well.

I noticed that SymbolTableSection.iter_symbols() deliberately leaves out the first entry of the symbol table (which is a STT_NOTYPE entry anyway) while DynamicSegment.iter_symbols() doesn't... should we make this consistent as well? Or is the risk of breaking existing tools using DynamicSegment too high?

Fixes: #218

[this change]: https://github.com/chromium/crashpad/commit/1f1657d573c789aa36b6022440e34d9ec30d894c

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -8,6 +8,7 @@
 #-------------------------------------------------------------------------------
 import itertools
 
+from collections import defaultdict
 from .hash import HashSection, GNUHashSection
 from .sections import Section, Symbol
 from .enums import ENUM_D_TAG
@@ -227,6 +228,36 @@ class DynamicSegment(Segment, Dynamic):
                 break
         Segment.__init__(self, header, stream)
         Dynamic.__init__(self, stream, elffile, stringtable, self['p_offset'])
+        self._symbol_list = None
+        self._symbol_name_map = None
+
+    def num_symbols(self):
+        """ Number of symbols in the table recovered from DT_SYMTAB
+        """
+        if self._symbol_list is None:
+            self._symbol_list = list(self.iter_symbols())
+        return len(self._symbol_list)
+
+    def get_symbol(self, index):
+        """ Get the symbol at index #index from the table (Symbol object)
+        """
+        if self._symbol_list is None:
+            self._symbol_list = list(self.iter_symbols())
+        return self._symbol_list[index]
+
+    def get_symbol_by_name(self, name):
+        """ Get a symbol(s) by name. Return None if no symbol by the given name
+            exists.
+        """
+        # The first time this method is called, construct a name to number
+        # mapping
+        #
+        if self._symbol_name_map is None:
+            self._symbol_name_map = defaultdict(list)
+            for i, sym in enumerate(self.iter_symbols()):
+                self._symbol_name_map[sym.name].append(i)
+        symnums = self._symbol_name_map.get(name)
+        return [self.get_symbol(i) for i in symnums] if symnums else None
 
     def iter_symbols(self):
         """ Yield all symbols in this dynamic segment. The symbols are usually
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -8,6 +8,7 @@
 #-------------------------------------------------------------------------------
 import itertools
 
+from .hash import HashSection, GNUHashSection
 from .sections import Section, Symbol
 from .enums import ENUM_D_TAG
 from .segments import Segment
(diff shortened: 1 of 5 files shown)


rupran:
I've marked this as 'work in progress' in order to get feedback, especially concerning the last paragraph (returning the first STT_NOTYPE element or not). I'm also not sure if computing the number of entries (and thus constructing/parsing the hash tables) should be done in DynamicSegment.get_number_of_symbols() which would then be called by DynamicSegment.iter_symbols().

eliben:
To be honest, I'm having a hard time keeping up with this dynamic table related work in pyelftools. @vapier added this code originally, and other folks worked on it - @rhelmot and others who can be found in the commit logs.

I'd be happy for someone else to review this in detail - I simply don't have the time to do this now.

rhelmot:
Yes, this is important work for me and I'll happily step in and take up some of your work. I can get to this later today.

eliben:
@rhelmot Thank you!

elftools/elf/dynamic.py
```diff
@@ -189,6 +250,39 @@ def __init__(self, header, stream, elffile):
...
+        return self._num_symbols
+
+    def get_symbol(self, index):
+        """ Get the symbol at index #index from the table (Symbol object)
+        """
+        if self._symbol_index_map is None:
+            self._symbol_index_map = {}
+            for idx, symbol in enumerate(self.iter_symbols()):
+                self._symbol_index_map[idx] = symbol
+        return self._symbol_index_map[index]
```
rhelmot: If symbol_index_map is strictly a map from index to symbol, it is probably better as a list - then this whole thing can be a list comprehension

elftools/elf/dynamic.py
```diff
@@ -189,6 +250,39 @@ def __init__(self, header, stream, elffile):
...
+        """
+        # The first time this method is called, construct a name to number
+        # mapping
+        #
+        if self._symbol_name_map is None:
+            self._symbol_name_map = defaultdict(list)
+            for i, sym in enumerate(self.iter_symbols()):
+                self._symbol_name_map[sym.name].append(i)
+        symnums = self._symbol_name_map.get(name)
+        return [self.get_symbol(i) for i in symnums] if symnums else None
```
rhelmot: why None instead of an empty list?

elftools/elf/dynamic.py
```diff
@@ -202,33 +296,56 @@ def iter_symbols(self):
...
+        # If DT_GNU_HASH did not exist, maybe we can use DT_HASH
+        if end_ptr is None:
+            _, hash_offset = self.get_table_offset('DT_HASH')
+            if hash_offset is not None:
+                hash_section = _DynamicHashSection(self.stream, hash_offset,
+                                                   self.elffile)
+                end_ptr = tab_ptr + \
+                    hash_section.get_number_of_symbols() * symbol_size
+
+        if end_ptr is None:
```
rhelmot: now that we have the better mechanisms to do this, should this arm perhaps log a warning that it is using a heuristic to detect the number of symbols?

for that matter, is it legal to have a dynamically linked object without a hashtable of some sort?

elftools/elf/dynamic.py
```diff
@@ -189,6 +250,39 @@ def __init__(self, header, stream, elffile):
...
         Dynamic.__init__(self, stream, elffile, stringtable, self['p_offset'])
+        self._symbol_index_map = None
+        self._symbol_name_map = None
+        self._num_symbols = None
+
+    def num_symbols(self):
+        """ Number of symbols in the table recovered from DT_SYMTAB
+        """
+        if self._num_symbols is None:
+            self._num_symbols = len(list(self.iter_symbols()))
```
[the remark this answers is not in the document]
rupran: Personally, I find that less readable, and the biggest overhead is in iter_symbols() anyway, so I'd prefer to leave it my way.

elftools/elf/dynamic.py
rupran: Oh, indeed, I had written it differently earlier and changed it without properly rethinking... I'll do that, thanks!

elftools/elf/dynamic.py
rupran: That's what SymbolTableSection does (https://github.com/eliben/pyelftools/blob/master/elftools/elf/sections.py#L174) so I figured this should behave the same way.

elftools/elf/dynamic.py
rupran: Good idea to add a warning message, I can do this. Do you think that logic should rather be in num_symbols() as it's actually about finding the number of symbols in the table?

For the second part I'm not sure, I'll dive into the binutils code tomorrow and check. My gut feeling would be that we need one but with ELF you never know...

rupran:
> This is fantastic! This is a set of functionality I've been missing from pyelftools for years.
> 
> If you want to add symbol lookup functionality to this code, I give you permission to use my implementation: https://github.com/angr/cle/blob/master/cle/backends/elf/hashtable.py. If this is integrated, I can probably remove my implementation.

Thanks :)

For lookup functionality it would probably be nice to also expose that as a HashTableSection class of some sort (if section headers are there) but then we'll get into the same discussion as on #135 about multiple inheritance...

I can implement that in the _DynamicHashSection objects now and we can build a good solution for actual sections later, what do you think?

rupran:
@rhelmot any comments on the new version and the comment above?

elftools/elf/dynamic.py
rhelmot: I think keeping it in iter_symbols is fine, considering the way the code is architected right now.

rhelmot:
- the easy way to deal with multiple inheritance is to escape the chains of object orientation and just write a function that duck types a hash table. I'm not too opinionated on any specific way to do that as long as it's documented :P
- I always consider the null symbol to be a real symbol, so we should return it.

elftools/elf/dynamic.py
rupran: Alright, sounds good.

About the warning message, there is no logging in pyelftools anywhere right now, so I don't think it makes sense to add it just for this case.

I glanced over the glibc code today and from the code inside dl-lookup.c (https://github.com/lattera/glibc/blob/master/elf/dl-lookup.c#L384 and following) it looks like there must be either a DT_HASH or DT_GNU_HASH table.

rupran:
> * the easy way to deal with multiple inheritance is to escape the chains of object orientation and just write a function that duck types a hash table. I'm not too opinionated on any specific way to do that as long as it's documented :P

Okay, this will be a follow-up feature anyway, so we can discuss this in a separate PR.

> * I always consider the null symbol to be a real symbol, so we should return it.

Sounds good, so it will stay like it is.

I've removed the 'draft' restriction from the pull request so that we could decide on merging the current state. Is there any way to trigger Travis without re-pushing?

rupran:
@rhelmot would you please review the newest version once more?

@eliben anything from your side you'd rather have changed?

rupran:
For the second commit, all added functions are checked on the `aarch64_super_stripped.elf` file which was already present in the `tests/testfiles_for_unittests/` directory. This also checks the logic in `iter_symbols()` with a `DT_HASH`-style table (as the added functionality depends on `iter_symbols()`).

The first commit tests `HashSection.get_number_of_symbols()` for both types, so the only thing not exercised at the moment is `DynamicSegment.iter_symbols()` with `DT_GNU_HASH` which would look very similar to the `test_reading_symbols` test in `test/test_dynamic.py`. I'll add one which uses another already existing binary.

PR #220, opened 2019-03-04
Add ability to parse the NT_FILE note found in core files

zephyrj:
Hi, 

I've been having a play with this library to poke around some core files to learn about their structure. I've found it very useful and so wanted to contribute in a small way. I've put together a pull request that adds the ability to read the NT_FILE part of the PT_NOTE that can be found  within core files. I've added some structs to parse it when the it detects that it has come across the NT_FILE note.

My research into the structure of the note has come from reading:
- https://www.gabriel.urdhr.fr/2015/05/29/core-file/ <- an excellent introduction to core file structure
- https://chromium.googlesource.com/native_client/nacl-binutils/+/upstream/master/binutils/readelf.c

I have added a test in test/test_core_notes.py that checks that the parsed note matches the output of the eu-readelf tool for NT_FILE.

Python is very much a language I'm still learning but I have tried to adhere to PEP8 guidelines in the added code. I noticed that the 79 char limit isn't strictly followed in all of the code but I have stuck to it in my changes; personally I am a PyCharm IDE user and prefer having longer lines but I'd rather not cause problems for hardcore terminal users. I'd appreciate any feedback on the changes

tox ouput looks good on my machine
```
Conclusion: SUCCESS
summary
  py27: commands succeeded
  py35: commands succeeded
  congratulations :)
```

### DIFF
--- a/elftools/elf/structs.py
+++ b/elftools/elf/structs.py
@@ -372,6 +372,17 @@ class ELFStructs(object):
                 String('pr_psargs', 80),
             )
 
+        self.Elf_Nt_File = Struct('Elf_Nt_File',
+                                  self.Elf_xword("num_map_entries"),
+                                  self.Elf_xword("page_size"),
+                                  Array(lambda ctx: ctx.num_map_entries,
+                                        Struct('Elf_Nt_File_Entry',
+                                             self.Elf_addr('vm_start'),
+                                             self.Elf_addr('vm_end'),
+                                             self.Elf_offset('page_offset'))),
+                                  Array(lambda ctx: ctx.num_map_entries,
+                                        CString('filename')))
+
     def _create_stabs(self):
         # Structure of one stabs entry, see binutils/bfd/stabs.c
         # Names taken from https://sourceware.org/gdb/current/onlinedocs/stabs.html#Overview
--- a/elftools/elf/structs.py
+++ b/elftools/elf/structs.py
@@ -372,6 +372,10 @@ class ELFStructs(object):
                 String('pr_psargs', 80),
             )
 
+        # A PT_NOTE of type NT_FILE matching the definition in
+        # https://chromium.googlesource.com/
+        # native_client/nacl-binutils/+/upstream/master/binutils/readelf.c
+        # Line 15121
         self.Elf_Nt_File = Struct('Elf_Nt_File',
                                   self.Elf_xword("num_map_entries"),
                                   self.Elf_xword("page_size"),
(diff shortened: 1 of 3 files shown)


test/test_core_notes.py
```diff
@@ -1,47 +1,188 @@
-#------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
```
eliben: What's changing in these comments?

test/test_core_notes.py
```diff
@@ -1,47 +1,188 @@
...
+
 class TestCoreNotes(unittest.TestCase):
     """ This test makes sure than core dump specific
         sections are properly analyzed.
     """
 
-    def test_core_prpsinfo(self):
-        with open(os.path.join('test',
+    def setUp(self):
+        self._core_file = open(os.path.join('test',
```
eliben: Why the setUp?

elftools/elf/structs.py
```diff
@@ -372,6 +372,17 @@ def _create_note(self, e_type=None):
                 String('pr_psargs', 80),
             )
 
+        self.Elf_Nt_File = Struct('Elf_Nt_File',
```
eliben: Can you please add a comment here explaining where this structure came from (link to some documentation)

test/test_core_notes.py
zephyrj: I tried to follow PEP8 to the letter as it was mentioned in the Hacking Guide.
https://www.python.org/dev/peps/pep-0008/#id31
PEP8 states that block comments should be a "# " (a hash symbol followed by a space) but the code didn't adhere to that so I changed it. I then removed a single "-" char from the end of the line to keep to the PEP8 line length https://www.python.org/dev/peps/pep-0008/#id19

Basically, the actions of a nervous first time contributor trying to stick to the letter of the law :)

test/test_core_notes.py
zephyrj: The opening of the test core file was needed in my test as well as the existing test in the file - it seemed to me that it was a case of duplicate code so I moved that task into a testSetup function given that it is common to both tests.

Browsing through the documentation it seems what I actually want is setUpClass() as there is no need for this to be executed before every test

test/test_core_notes.py
eliben: No worries.
Let's keep them consistent with the other header comments in pyelftools

elftools/elf/structs.py
zephyrj: Sorry for the delay; been very busy recently. I have added a link

test/test_core_notes.py
zephyrj: I've reverted the header style to the way it was previously

PR #225, opened 2019-05-01
Improved handling of location information

andersdellien:
This commit moves some of the location-handling code from the examples
to a new class (LocationParser) in order to make it more reusable.

Also adds two test files containing location information.


andersdellien:
Yes, that makes sense. I'll modify the pull request.

PR #231, opened 2019-06-20
Fix deprecation warning in Python 3.7

scottj97:
```
$SITE_PYTHON/lib/python3.7/site-packages/elftools/construct/lib/container.py:5
 Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

This change is compatible with Python 3.3 and up, when the ABCs were moved to collections.abc.

### DIFF
--- a/elftools/construct/lib/container.py
+++ b/elftools/construct/lib/container.py
@@ -2,8 +2,8 @@
 Various containers.
 """
 
-from collections import MutableMapping
 from pprint import pformat
+from .py3compat import MutableMapping
 
 def recursion_lock(retval, lock_name = "__recursion_lock__"):
     def decorator(func):
--- a/elftools/construct/lib/py3compat.py
+++ b/elftools/construct/lib/py3compat.py
@@ -6,6 +6,11 @@
 import sys
 PY3 = sys.version_info[0] == 3
 
+try:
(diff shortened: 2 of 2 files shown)


scottj97:
Force-pushed a better implementation that retains backward compatibility.

scottj97:
It's not quite that simple though...this moved in Python 3.3, so if we make it dependent on PY3 then 3.2 will be broken.

I guess in py3compat we could import MutableMapping using these 4 lines of code, then in container.py import from py3compat? Is that an improvement?

eliben:
> It's not quite that simple though...this moved in Python 3.3, so if we make it dependent on PY3 then 3.2 will be broken.
> 
> I guess in py3compat we could import MutableMapping using these 4 lines of code, then in container.py import from py3compat? Is that an improvement?

That's how it works for the other stuff. It's just useful to collect all these compatibility imports in a single place

scottj97:
Updated, though I put it into `elftools/construct/lib/py3compat.py` instead of `elftools/common/py3compat.py`. Confused me for a while that there's two modules with the same name.

scottj97:
I forgot to test python3. Fixed that now.

eliben:
Thanks!

PR #239, opened 2019-09-17
dwarf_expr: Add DW_OP_{lo,hi}_user

woodruffw:
Adds `DW_OP_lo_user` and `DW_OP_hi_user`.

From DWARFv4 7.1 and 7.7.1:

> Values in the range between prefix_lo_user and prefix_hi_user inclusive, are reserved for vendor specific extensions. Vendors may use values in this range without conflicting with current or future system-defined values. All other values are reserved for use by the system.


woodruffw:
Just realized that it doesn't make a ton of sense to decode just the start and end opcodes here, so feel free to close this if you feel it isn't appropriate.

I could use `_generate_dynamic_values` instead, but I don't know if there's any precedent for representing vendor operations as e.g. `DW_OP_lo_user0` and onwards.

woodruffw:
> There's a precedent to having such constants, look for example at `DW_ATE_lo_user` and `DW_TAG_lo_user`

Ah, good to know!

> How do you envision using these? Do you have a test or a use case in mind?

My plan was to use them as a bounds check, i.e. just to have something more symbolic than the opcode literals in my codebase.

PR #240, opened 2019-09-17
dwarf_expr: Add DWARFv5 OPs

woodruffw:
Adds more `DW_OP_*` constants added in version 5 of the DWARF spec.

See DWARFv5, 7.7.1.


eliben:
Thanks for the contribution

PR #242, opened 2019-10-01
dwarf/enums Add GNU parameter tags

woodruffw:
Adds `DW_TAG_GNU_template_template_param`, `DW_TAG_GNU_template_parameter_pack`, and `DW_TAG_GNU_formal_parameter_pack`, which both GCC and Clang emit to help identify template template parameters and variadic template/formal parameters.

`DW_TAG_GNU_template_parameter_pack` and `DW_TAG_GNU_formal_parameter_pack` were slated for DWARFv5 according to [this](http://wiki.dwarfstd.org/index.php?title=C%2B%2B0x:_Variadic_templates), but it looks like work has stalled and the vendor-prefixed names/opcodes remain in use.

Also reformats the table to fit the new tag names.


elftools/dwarf/enums.py
```diff
@@ -11,79 +11,82 @@
...
+    DW_TAG_namespace                   = 0x39,
+    DW_TAG_imported_module             = 0x3a,
+    DW_TAG_unspecified_type            = 0x3b,
+    DW_TAG_partial_unit                = 0x3c,
+    DW_TAG_imported_unit               = 0x3d,
+    DW_TAG_mutable_type                = 0x3e,
+    DW_TAG_condition                   = 0x3f,
+    DW_TAG_shared_type                 = 0x40,
+    DW_TAG_type_unit                   = 0x41,
+    DW_TAG_rvalue_reference_type       = 0x42,
```
eliben: Maybe it's the review tool, but the `=`s don't seem all aligned now. Why did you reformat the ones above if the ones below are on a different offset anyway?

woodruffw: Hmm, I think it might be the review tool -- `_default_` is now unaligned (I'll fix that), but the others were aligned before and after my changeset. In any case, I can drop the alignment changes if you'd like.

woodruffw: Weird. `_default_` is aligned in my local editor, but not on GitHub's review tool 😕

woodruffw:
Thanks!

PR #244, opened 2019-10-04
dwarf/enums: More attributes, tags, and forms

woodruffw:
Another medley of enum additions, sourced from DWARFv5 changes as well as vendor-specific values.

Sources:

* https://sourceware.org/elfutils/DwarfExtensions
* https://github.com/llvm-mirror/llvm/blob/master/include/llvm/BinaryFormat/Dwarf.def


elftools/dwarf/enums.py
```diff
@@ -251,35 +260,55 @@
...
+    DW_FORM_strx2               = 0x26,
+    DW_FORM_strx3               = 0x27,
+    DW_FORM_strx4               = 0x28,
+    DW_FORM_addrx1              = 0x29,
+    DW_FORM_addrx2              = 0x2a,
+    DW_FORM_addrx3              = 0x2b,
+    DW_FORM_addrx4              = 0x2c,
 
-    DW_FORM_GNU_strp_alt    = 0x1f21,
+    DW_FORM_GNU_addr_index  = 0x1f01,
```
eliben: Again, can you please just double check the formatting here? Seems like indentation changed by from line 308 the `=` is at a different column

woodruffw: Yeah, that was intentional in this case -- I noticed that some of the groupings were only locally aligned, not aligned as a whole. For 308 onwards I just made the new `DW_FORM_GNU` enums align with the existing ones, and didn't make them align with the non-vendor-specific enums. I can change that, though.

eliben: Thanks for checking. I'll just merge it and fix it up if it looks off

woodruffw:
Thanks!

PR #245, opened 2019-10-04
dwarf/constants: More DW_LANG, DW_ATE constants

woodruffw:
Most of these were added in DWARFv5.

Reference:

* https://github.com/llvm-mirror/llvm/blob/master/include/llvm/BinaryFormat/Dwarf.def


woodruffw:
Thanks!

PR #246, opened 2019-10-15
dwarf/die: Handle DW_FORM_flag_present in value translation

woodruffw:
When an attribute has form DW_FORM_flag_present it is implicitly
indicated as present, with no actual value.

Ref. DWARFv4, section 7:

> A flag is represented explicitly as a single byte of data (DW_FORM_flag) or implicitly
(DW_FORM_flag_present). In the first case, if the flag has value zero, it indicates the
absence of the attribute; if the flag has a non-zero value, it indicates the presence of the
attribute. In the second case, the attribute is implicitly indicated as present, and no value is
encoded in the debugging information entry itself.


woodruffw:
Alright, here's a small sample:

```c++
#include <cstdlib>
#include <iostream>

#define COINFLIP() (rand() % 2)

int main(int argc, char const *argv[]) {
  int x = 100;

  if (COINFLIP()) {
    auto lambda = [x]() {
      std::cout << "x=" << x << '\n';
      if (COINFLIP()) {
        std::cout << "foo" << '\n';
      } else {
        std::cout << "bar" << '\n';
      }
    };

    lambda();
  }
  return 0;
}
```

compiled with Clang 7:

`clang -g -grecord-gcc-switches -O0 -std=c++11 -o lambda lambda.cpp`

You can see the `DW_FORM_flag_present` forms on `DW_AT_artificial` with `llvm-dwarfdump -show-form lambda`, e.g.:

```
0x00000078:           DW_TAG_formal_parameter
                        DW_AT_type [DW_FORM_ref4]       (0x00000865 "Init*")
                        DW_AT_artificial [DW_FORM_flag_present] (true)

0x0000007d:           NULL
```

eliben:
Yes, a unit test would be great!

woodruffw:
> Yes, a unit test would be great!

Awesome, I'll make one tomorrow.

eliben:
LGTM. Seems like you have failures on Travis with this PR, though?

woodruffw:
> LGTM. Seems like you have failures on Travis with this PR, though?

Yep, some dumb mistakes on my part. Should be good to go now!

Issue #113, opened 2016-08-18
CompileUnit: Track DIEs by offset


JonathonReinhart:
DIEs refer to one another by their offset in the stream.

For example, a [`DW_TAG_typedef`](http://www.dwarfstd.org/doc/DWARF4.pdf#page=96) DIE will have  "a `DW_AT_type` attribute whose value is a reference to the type named by the typedef."

So to get the actual type of a typedef DIE, I currently use something like this in my monkey-patched `CompileUnit` class:

``` python
import elftools.dwarf.compileunit
class MyCompileUnit(elftools.dwarf.compileunit.CompileUnit):

    def get_DIE_at_offset(self, offset):
        for die in self.iter_DIEs():
            if die.offset == offset:
                return die 
        return None

elftools.dwarf.compileunit.CompileUnit = MyCompileUnit
```

This O(n) lookup is inefficient.

Currently, `CompileUnit` [stores all of the DIEs in a simple list](https://github.com/eliben/pyelftools/blob/60319cbb9d469e8d7c56e8f85dd55294c0ce4f62/elftools/dwarf/compileunit.py#L120).

I propose instead (or also) storing them in a dictionary, using their `offset` as the key. This will be unique for all DIEs, and provide fast lookup via a new function like:

``` python
def _get_DIE_by_offset(self, offset):
    return self._diedict[offset]
```

Since the existing `_get_DIE` could be used by existing code (even though it is "private"), it makes sense to leave the existing infrastructure in place, and maintain this offset-lookup side-by-side.

eliben:
Sure, I don't see why not (PRs welcome :)), but this isn't high priority. Compared to time needed to parse the DWARF, building a dict outside CompileUnit should be trivial (time-wise). It gives you all the DIEs, so if you anticipate frequent lookups you can create the offset->DIE mapping yourself.

JonathonReinhart:
Yes, I do hope to submit a PR, it's just a matter of when.  I opened this issue to a) advertise my thoughts in case there were any conflicts with existing efforts, and b) remind myself what the hell my plan was in the future :smiley:

eliben:
Let me know if you plan to be working on this. Since this is something that may be relatively easy to do outside pyelftools, I'm not super-inclined to add it to the library.

georgemorgan:
This would be substantially helpful for me. I spent a while looking around the source code trying to find this until I discovered I needed to implement it myself.

dgvogol:
[...]

PR #249, opened 2019-10-30
Lazy DIE parsing

woodruffw:
Supersedes/closes #216.

### DIFF
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -55,6 +55,8 @@ class CompileUnit(object):
 
         # A list of DIEs belonging to this CU. Lazily parsed.
         self._dielist = []
+        # A list of corresponding DIE offsets.
+        self._diemap = []
 
     def dwarf_format(self):
         """ Get the DWARF format (32 or 64) for this CU
@@ -112,12 +114,14 @@ class CompileUnit(object):
 
         # First pass: parse all DIEs and place them into self._dielist
         die_offset = self.cu_die_offset
+        dm = self._diemap
         while die_offset < cu_boundary:
             die = DIE(
                     cu=self,
                     stream=self.dwarfinfo.debug_info_sec.stream,
                     offset=die_offset)
             self._dielist.append(die)
+            dm.append(die_offset)
             die_offset += die.size
 
         # Second pass - unflatten the DIE tree
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -6,6 +6,7 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from bisect import bisect_left
 from .die import DIE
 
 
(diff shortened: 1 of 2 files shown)


elftools/dwarf/compileunit.py
```diff
@@ -53,8 +54,13 @@ def __init__(self, header, dwarfinfo, structs, cu_offset, cu_die_offset):
         # requested.
         self._abbrev_table = None
 
-        # A list of DIEs belonging to this CU. Lazily parsed.
+        # A list of DIEs belonging to this CU.
+        # This list is lazily constructed as DIEs are iterated over.
         self._dielist = []
+        # A list of corresponding DIE offsets.
+        # This list is used to determine where to insert a DIE
+        # within `self._diemap` during iteration for future memoization.
```
eliben: Within `_dielist`?

elftools/dwarf/compileunit.py
```diff
@@ -73,14 +79,81 @@ def get_top_DIE(self):
         """ Get the top DIE (which is either a DW_TAG_compile_unit or
             DW_TAG_partial_unit) of this CU
         """
-        return self._get_DIE(0)
+
+        # Note that a top DIE always has minimal offset, no bisect required.
+        if self._diemap and self._diemap[0] == self.cu_die_offset:
```
eliben: This code implies `_diemap` could be non-empty but the `[0]` element is not at the current die offset, is that true?

I'd like the invariants to be really well documented

eliben:
Thanks for taking this on! I'll be happy to review just the lazy parsing part

elftools/dwarf/compileunit.py
woodruffw: Yep, thanks! Will fix.

elftools/dwarf/compileunit.py
woodruffw: Yeah, I think the structure here is a little roundabout -- it *should* be sufficient to just check whether `_diemap` is empty and, if so, parse and memoize the DIE at `cu_due_offset`. Will test.

eliben:
> > Thanks for taking this on! I'll be happy to review just the lazy parsing part
> 
> No problem, and thank you!

Let me know when it's ready for another round of review

elftools/dwarf/compileunit.py
```diff
@@ -53,8 +54,13 @@ def __init__(self, header, dwarfinfo, structs, cu_offset, cu_die_offset):
         # requested.
         self._abbrev_table = None
 
-        # A list of DIEs belonging to this CU. Lazily parsed.
+        # A list of DIEs belonging to this CU.
+        # This list is lazily constructed as DIEs are iterated over.
         self._dielist = []
+        # A list of corresponding DIE offsets.
```
eliben: Can you explain a bit more what each element of `_diemap` is (e.g. its type) and semantics? Is each element an offset from which the corresponding DIE was parsed in the file? So it's the same as DIE.offset? But I guess in a separate list to allow binary searches?

Sicne diemap is a key data structure for the whole approach, it's worth documenting it well. With a good mental model of what diemap is doing, reading the rest of the code becomes easier. "... Show me your tables ..." etc.

woodruffw: Sure:

Each element in `_diemap` is an integer, corresponding to a DIE's offset within the file. You're exactly right re: binary searches: the only reason it isn't a `dict` or other type (more closely paring it with `_dielist` is so that `bisect_left` can be used on it.

I'll update the docs to (hopefully) make that more clear.

woodruffw:
Okay, I've updated the documentation for the `_diemap` attribute. Should be good for another round of review.

elftools/dwarf/compileunit.py
```diff
@@ -73,14 +82,83 @@ def get_top_DIE(self):
...
+    def iter_DIE_children(self, die):
+        """ Given a DIE, yields either its children, without null DIE list
+            terminator, or nothing, if that DIE have no children.
+
+            The null DIE terminator is saved in that DIE when iteration ended.
+        """
+        if not die.has_children:
+            return
+
+        cu_off = self.cu_offset
```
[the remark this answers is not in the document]
woodruffw: Yep, looks like I missed this when removing the other abbreviated attributes. Will fix.

elftools/dwarf/compileunit.py
```diff
@@ -73,14 +82,83 @@ def get_top_DIE(self):
...
+
+        cur_offset = die.offset + die.size
```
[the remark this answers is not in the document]
woodruffw: Added a comment explaining `cur_offset`'s usage as a pivot -- let me know what you think!

eliben:
The PR description still says "in progress". Is there anything else you were meaning to add here, or can I go ahead and merge this?

woodruffw:
> The PR description still says "in progress". Is there anything else you were meaning to add here, or can I go ahead and merge this?

Nope, should be good to merge. I'll remove that note.

JonathonReinhart:
@eliben @woodruffw: @mdmillerii mentioned this in issue #113 which I opened in 2016. Without having dug back into the problem myself, would you agree that this closes #113?

woodruffw:
Yes, I think so -- these changes mean that DIEs are cached and lookup is no longer O(n).

PR #267, opened 2020-01-27
Feature expressions

sevaa:
Support for several DWARF opcodes that occur in the wild.

The way DWARF expressions work, unknown opcodes throw off the parser completely, since the structure of the arguments can't be derived from any outside metadata. Had to fix that.

DW_OP_GNU_entry_value introduces the concept of nested expressions - an expression blob as an argument of an opcode. The visitor pattern is a poor fit for that, so I've introduced the GenericExprDumper - a generic expression to command list converter, with a hook for derived classes to provide a single operation formatter function. Nested expressions fit more naturally into that.

Feel free to redo or throw away the DIE.resolve_reference(). It was a quick and dirty fix.


sevaa:
Checked failed - closing, will redo

PR #264, opened 2020-01-23
Cached random access to CUs and DIEs

mdmillerii:
Here is a series that refactors the DIE cache to cache entries without walking from the top of a CU,
and also one that adds a similar cache to the CU lookup in the DWARFInfo.

PR #268, opened 2020-01-27
for sibling of form ref_addr, only sibling value should be used

sagiben:
Hi, 

cur_offset  calculation might be wrong if the sibling is DW_FORM_ref_addr, because it holds an reference from the beginning of the .debug_info.

>     This type of reference (DW_FORM_ref_addr) is an offset from the
>     beginning of the .debug_info section of the target executable or
>     shared object file, or, for references within a supplementary object file,
>     an offset from the beginning of the local .debug_info section;

this fix is related to : 
https://github.com/eliben/pyelftools/commit/670079afe5a472aab74cec74acc15ae7c50cdb83#r36974522


eliben:
@sagiben thanks for the report
@woodruffw could you please review it, as it relates to code you've committed for lazy DIE parsing?

woodruffw:
Yeah, I can take a look this evening.

woodruffw:
Yeah, looks like @sagiben is right -- this calculation will be wrong for that `DW_FORM`.

@sagiben are you still willing to do a PR for this? If not I can take it (and if so please feel free to ping me at review time!)

mdmillerii:
This works, or we could use `.get_DIE_from_attribute()` in #264, which would check the form explicitly.  

The sig8, sup4 and sup8 forms don't make sense for a sibling so this covers the likely cases through existing dwarf5.

sagiben:
thanks @woodruffw.

Regarding the PR, whatever works for you.
This PR is ready for a merge, unless I missed something.

woodruffw:
> This PR is ready for a merge, unless I missed something.

Sorry, I totally missed that this was a PR and not just an issue! My bad.

sagiben:
@eliben, can you merge this PR ?

eliben:
> @eliben, can you merge this PR ?

Any way you could add a test that exercises this (fails without the patch, passes with)?

sevaa:
Doesn't #273 address the underlying issue?

sagiben:
> Doesn't #273 address the underlying issue?

unfortunately no.
I have an ELF (which I can't share) which reproduce the issue I mentioned.
I ran the following command with the latest commit to master (a70ccd3e163d2ee04bed68212654f8843b90300d)


```
$ /tmp/pyelftools/scripts/readelf.py --debug-dump=info /path/to/my.elf
...
  File "/tmp/pyelftools/scripts/readelf.py", line 1482, in <module>
    main()
  File "/tmp/pyelftools/scripts/readelf.py", line 1458, in main
    readelf.display_debug_dump(args.debug_dump_what)
  File "/tmp/pyelftools/scripts/readelf.py", line 780, in display_debug_dump
    self._dump_debug_info()
  File "/tmp/pyelftools/scripts/readelf.py", line 989, in _dump_debug_info
    for die in cu.iter_DIEs():
  File "/tmp/pyelftools/elftools/dwarf/compileunit.py", line 179, in _iter_DIE_subtree
    for d in self._iter_DIE_subtree(c):
  File "/tmp/pyelftools/elftools/dwarf/compileunit.py", line 179, in _iter_DIE_subtree
    for d in self._iter_DIE_subtree(c):
  File "/tmp/pyelftools/elftools/dwarf/compileunit.py", line 178, in _iter_DIE_subtree
    for c in die.iter_children():
  File "/tmp/pyelftools/elftools/dwarf/compileunit.py", line 132, in iter_DIE_children
    offset=cur_offset)
  File "/tmp/pyelftools/elftools/dwarf/die.py", line 94, in __init__
    self._parse_DIE()
  File "/tmp/pyelftools/elftools/dwarf/die.py", line 173, in _parse_DIE
    abbrev_decl = self.cu.get_abbrev_table().get_abbrev(self.abbrev_code)
  File "/tmp/pyelftools/elftools/dwarf/abbrevtable.py", line 36, in get_abbrev
    return self._abbrev_map[code]
KeyError: 118

```

sevaa:
The issue I've addressed was specific to DWARF 2 targeting a 64-bit architecture. Was that your case, too?

If the binary is not to be shared, maybe we could work towards sharing a couple of debug sections alone? This way, the **code** won't leave the premises :)

mdmillerii:
It looks like your toolchain generates these; perhaps you can create a generic testcase (that would be much smaller)?  Maybe just a short test program?

sevaa:
I've observed this in an iOS binary made by Xcode on Mac. It reproduces on the latest Xcode. In my own projects, iOS binaries are not ELF files. I use `pyelftools` to parse the DWARF sections of those, but file parsing per se is done by another library, `filebytes`. I don't think @eliben would appreciate bringing third party dependencies into the test suite.

What's *your* toolchain?

sagiben:
Hi, I was able to reproduce this issue with DIAB WindRiver compiler. I can try to generate test case, would you like that I will upload the ELF ?

sevaa:
Please share a binary, if possible. Now I'm quite ticked off that my fix doesn't apply. :)

sagiben:
I uploaded the ELF to send.firefox 
[test1.elf](https://send.firefox.com/download/414a6a89b8da9856/#bNZHE-xPdxuMJJI1q80SuA)
available for download for a week.
pyelftools crash is not the same as in the original bug report but it fails at the same location but with my fix pyelftools dump all the dwarf info

sevaa:
Got the file, reproduced the crash. The issue is definitely distinct from what I had, since this is 32-bit  code.

sagiben:
ELF test case was added to PR.

eliben:
Note: I've reverted this commit since it breaks readelf tests. The breakage log is pasted in the revert commit.

PR #269, opened 2020-01-29
Cache machine_arch in order to speed up _do_apply_relocation()

mephi42:
When loading linux kernel debug info under pypy3, according to
cProfile, get_machine_arch() calls from _do_apply_relocation()
constitute 30% of the run time. Avoid unnecessary calls by caching the
return value in the new RelocationHandler.machine_arch field.

### DIFF
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -110,6 +110,8 @@ class RelocationHandler(object):
     """
     def __init__(self, elffile):
         self.elffile = elffile
+        # Cache machine_arch in order to speed up _do_apply_relocation()
+        self.machine_arch = self.elffile.get_machine_arch()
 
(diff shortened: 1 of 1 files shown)


mephi42:
It appears that my linux kernel debug info (Fedora 31,
5.4.13-201.fc31.x86_64) must not be relocated, since its `e_type` is
`ET_EXEC`, not `ET_REL`. Despite that and for whatever reason, the
relocation section for `.debug_info` is present and contains completely
bogus information.

So I'm using `get_dwarf_info(relocate_dwarf_sections=False)` now and
everything (including performance) is great. I'd rather have the
`relocate_dwarf_sections` default value be based on `e_type` value
rather than being always `True` though - I will create an issue to
discuss this.

Sorry for the noise.

PR #273, opened 2020-01-31
ref_addr size changed between v2 and v3 - take 2

sevaa:
In DWARF 2, the DW_FORM_ref_addr format matches the target address size, while in DWARF3+ it matches the bitness of the CU record. Here are the relevant fragments from the spec, part 7:

v2:

> The second type of reference is the address of any debugging information entry within the same executable or shared object; it may refer to an entry in a different compilation unit from the unit containing the reference. This type of reference (DW_FORM_ref_addr) **is the size of an address on the target architecture**; it is relocatable in a relocatable object file and relocated in an executable file or shared object. 

v3:

> The second type of reference can identify any debugging information entry within a program; in particular, it may refer to an entry in a different compilation unit from the unit containing the reference, and may refer to an entry in a different shared object. This type of reference (DW_FORM_ref_addr) is an offset from the beginning of the .debug_info section of the target executable or shared object; it is relocatable in a relocatable object file and frequently relocated in an executable file or shared object. For references from one shared object or static executable file to another, the relocation and identification of the target object must be performed by the consumer. **In the 32-bit DWARF format, this offset is a 4-byte unsigned value; in the 64-bit DWARF format, it is an 8-byte unsigned value** (see Section 7.4).

If `elftools` encounters 32-bit DWARF v2 targeting a 64-bit architecture, it will misparse DW_FORM_ref_addr and crash downstream.

I encountered this in an iOS binary from 2017, built with Xcode several versions ago for ARM64. This probably never came up before because by the time 64 bit code became relevant, most toolchains would generate DWARF 3 or newer.


sagiben:
Hi,
Also look at https://github.com/eliben/pyelftools/pull/268 
I had an issue that readelf crashed in case that a sibling contain DW_FORM_ref_addr

eliben:
Thanks, the change looks fine. But could you add a test? Any chance to add the test file you encountered it with? Or some minimized version thereof?

sevaa:
I'll try. It was from a 2017 version of the toolchain. The bug is specific to DWARFv2; I'll see if I can make the modern version of LLVM (or any other compiler) emit specifically v2.

eliben:
gcc has `-gdwarf-version`

sevaa:
Provided. Sorry for the wacky format of the sample data files; I couldn't get GCC to emit it, used Xcode/Clang instead. pyelftools doesn't take Mach-O files, so I've dumped a handful of relevant sections from the DWARF bundle.

elftools/dwarf/dwarfinfo.py
```diff
@@ -290,19 +290,23 @@ def _parse_CU_at_offset(self, offset):
         # with a default address_size=4. If, after parsing the header, we
         # find out address_size is actually 8, we just create a new structs
         # object for this CU.
+        # Same logic for DWARF version. Some structs, notably DW_FORM_ref_addr,
```
eliben: This whole comment block is out of date now.

I propose to just unconditionally create a new `cu_structs` - it's cheap. And modify the comment accordingly. WDYT?

test/test_refaddr_bitness.py
```diff
@@ -0,0 +1,49 @@
...
+# elftools tests
+#
+# Seva Alekseyev (sevaa@sprynet.com)
+# This code is in the public domain
+#------------------------------------------------------------------------------
+
+import unittest
+import os, sys, io
+
+# sys.path.insert(1, os.getcwd())
```
eliben: Remove?

test/test_refaddr_bitness.py
```diff
@@ -0,0 +1,49 @@
...
+
+from elftools.dwarf.dwarfinfo import DWARFInfo, DebugSectionDescriptor, DwarfConfig
+
+class TestRefAddrOnDWARFv2With64BitTarget(unittest.TestCase):
```
eliben: Please add a substantial comment explaining what this test is testing/doing, and how.

eliben:
Did another round of review and assigned to issue to you. Please reassign back to me once you're done and it's ready for another round

test/test_refaddr_bitness.py
sevaa: Commented

test/test_refaddr_bitness.py
sevaa: That was a debug helper line. Removed.

elftools/dwarf/dwarfinfo.py
sevaa: Made the structs creation unconditional, explained in the comments.

sevaa:
I don't think I can assign PRs to people around here.

eliben:
> I don't think I can assign PRs to people around here.

Oops, sorry about that. I guess unassigning yourself is a reasonable signal too :)

Issue #274, opened 2020-02-03
False positive on LocationParser.attribute_has_location


sevaa:
LocationParser.attribute_has_location() returning `True` doesn't guarantee that the attribute has a location that can be parsed with LocationParser. I'm currently looking at a v2 file from Android NDK with the following DIE:

>  <2><a73e>: Abbrev Number: 75 (DW_TAG_member)
>     <a73f>   DW_AT_name        : (indirect string, offset: 0x394c): FL_FOO
>     <a743>   DW_AT_decl_file   : 23
>     <a744>   DW_AT_decl_line   : 22
>     <a745>   DW_AT_type        : <0x5256>
>     <a749>   DW_AT_external    : 1
>     <a74a>   DW_AT_declaration : 1
>     <a74b>   **DW_AT_const_value (DW_FORM_data4) : 0x40000000**
> 

`attribute_has_location()` returns `true` for the `DW_AT_const_value` attribute. It's clearly not a location pointer.

The parent of that goes:

>  <1><a720>: Abbrev Number: 3 (DW_TAG_structure_type)
>     <a721>   DW_AT_name        : (indirect string, offset: 0x34f0): CMyClass
>     <a725>   DW_AT_byte_size   : 40
>     <a726>   DW_AT_decl_file   : 23
>     <a727>   DW_AT_decl_line   : 17
>     <a728>   DW_AT_sibling     : <0xa7b5>

The corresponding code item in a C++ source is a class level static const int:

struct CMyClass
{
    static const int FL_FOO = 0x40000000;
}

0x4 is clearly not a location code. Maybe some careful parsing of the spec is in order?

eliben:
PRs welcome

sevaa:
I'm not yet sure what's the correct condition there. I mean, the existing code was probably born from experience; sometimes DW_FORM_data4 **does** contain an offset into debug_loc. Do you have any files where this happens? Which DIE tag(s) is it under? I'll search in my own binary corpus, but the wider the net, the better.

Even if I take the easy way out and assume that `DW_FORM_data4` is never a location list if used within a `DW_TAG_member`, that would still require that the tag (or a DIE reference) is passed to the function. There's no DIE reference in the attribute object. That means - the fix will be opt-in; a library consumer will only be taking advantage of the fix if they change their invocation of attribute_has_location().

eliben:
It'd be difficult for me to dig up this knowledge at this point since I haven't actually been *using* this library for quite a while.

[...]

PR #276, opened 2020-02-04
DW_AT_const_value of form data4/8 is not a location

sevaa:
Crude fix for #274.


sevaa:
Something went wrong, will fix

PR #284, opened 2020-02-07
GNU expressions

sevaa:
Support for some GNU opcodes for DWARF expressions.

One of them, `DW_OP_GNU_entry_value`, threw a monkey wrench into the whole expression parsing scheme. As an argument, it contains a variable length binary blob **with another expression**. So the expression parsing logic should recurse. Now, expression parsing in pyelftools was originally designed with progressive parsing in mind, thus the visitor logic. There is no out out the box logic for "parse this expression into a single data structure".

So I came up with one. In order to keep things compatible, I've split the `ExprDumper` class from `descriptions.py` into two parts - the "unite operations into a list" bit and the "format an operation into a string" bit. The former resides under `dwarf_expr` now, as `GenericExprDumper`, and the parsing logic of `DW_OP_GNU_entry_value` relies on that to combine the nested expression into a list.

There's a unit test, too.


eliben:
There's not much information online about `DW_OP_GNU_entry_value` and I can't find it in any of the DWARF specs. Can you point to a good reference doc explaining it?

sevaa:
Here: http://www.dwarfstd.org/ShowIssue.php?issue=100909.1

eliben:
> Here: http://www.dwarfstd.org/ShowIssue.php?issue=100909.1

OK, I see that op in the DWARF v5 spec is well, but the link doesn't mention GNU -- do these ops have the same semantics as the non-GNU ones?

sevaa:
The binary where I saw this was DWARF 4, not 5. My theory is that they've blessed the GNU extension for DWARF 5, but in DWARF 4 file it still counts as a user extension.

Supporting GNU-less `DW_OP_entry_value` *as introduced in DWARF 5* is not the point of this PR. For one thing, I have nothing to test it on.

elftools/dwarf/dwarf_expr.py
```diff
@@ -169,6 +175,29 @@ def _visit_OP_addr(self, opcode, opcode_name):
...
                 struct_parse(self.structs.Dwarf_target_addr(''), self.stream)]
 
+    # ULEB128, then an expression of that length
+    def _visit_OP_nestedexpr(self, opcode, opcode_name):
+        size = struct_parse(self.structs.Dwarf_uleb128(''), self.stream)
+        nested_expr_blob = self.stream.read(size)
+        if isinstance(nested_expr_blob, str): # Python 2 patch. In Py3, it comes as bytes.
+            nested_expr_blob = [ord(b) for b in nested_expr_blob]
+        dumper = GenericExprDumper(self.structs)
+        # If used from a dumper context, reuse the outer operation formatter
```
eliben: Ugh, this seems extremely hacky, sorry. The circular dependency between visitor and dumper, the magic discovery of what "context" you're used in. I wish there was a cleaner way to design this. DWARF has about 17000 dark corners and gotchas, if we implement each with a cut-through-all-layers hack, pyelftools will become completely unusable very soon. I'd rather we not support this opcode for now if there's no better solution.

test/test_dwarf_expression_opcodes.py
```diff
@@ -0,0 +1,59 @@
+#------------------------------------------------------------------------------
+# elftools tests
```
eliben: Unit tests are good, but I'm also worried about the proliferation of these small obscure `.o` files with extracted sections, which I don't know how to reproduce.

One of the keys to the library's stability is readelf-comparison tests. It would be much preferred if one could find a relevant readelf-comparison test for new added features.

eliben:
I concede that a confounding factor here is tha the `GenericExprVisitor` class is difficult to extend for the recursive case. This class is only used in a single place in pyelftools, and it's not clear how widely used it is externally. I'd be OK with modifying this interface from `dwarf_expr` as long as converting existing uses of `GenericExprVisitor` to the new thing would be trivial.

One idea would be a simpler API in `dwarf_expr` that takes an expr and returns a nested Python representation (simply nested lists of structs?), which could then be easily converted to an appropriate string in `descriptions`.

test/test_dwarf_expression_opcodes.py
sevaa: Extracted sections was another PR :) That's a discussion for elsewhere. This one has .o files, perfectly ELF. What's wrong with them?

sevaa:
Well, I always saw the visitor as pointless, even before the `entry_value` debacle. Who needs progressive parsing of a data structure that's typically well under 50 bytes?

Now, speaking of a reasonable path forward. Would you mind if I:
 - move the expression parsing logic from `process_expr` to another function (e. g. `parse_expr`), that would emit a list of (opcode, opcode_name, args) tuples? That would recurse naturally. Create another `GenericExprVisitor`, parse the nested expr, and use that as the first argument.
 - reimplement `process_expr` in terms of that - parse into a list, then call _after_visit for each operation so that ExprDumper stays happy?

This will not affect those consumers that rely on reimplementing `_after_visit` *alone*, such as the current version of your own `ExprDumper`. More involved subclasses might be out of luck. For nested expressions, there will be no iterated visiting. Notably, those consumers who reimplement `_default_visitor` won't see their custom visitor called for nested expression commands.

Then, if we're aiming to test against `readelf --debug-dump=info`, I guess I'll have to tweak ExprDumper too, to emit something readelf compatible.

As a process note, since the library moved forward since I started, I think I should pull the current master and start from scratch.

eliben:
> Well, I always saw the visitor as pointless, even before the `entry_value` debacle. Who needs progressive parsing of a data structure that's typically well under 50 bytes?
> 

The goal, if I recall correctly, wasn't progressive parsing for performance. It was simply a convenience API that would consumers would use to process exprs without writing "type switches". Such visitor APIs are common in Python (there are a few in the stdlib), especially for evaluation-related APIs.

But it turned out that `ExprDumper` doesn't actually use this as planned...

> Now, speaking of a reasonable path forward. Would you mind if I:
> 
>     * move the expression parsing logic from `process_expr` to another function (e. g. `parse_expr`), that would emit a list of (opcode, opcode_name, args) tuples? That would recurse naturally. Create another `GenericExprVisitor`, parse the nested expr, and use that as the first argument.
> 
>     * reimplement `process_expr` in terms of that - parse into a list, then call _after_visit for each operation so that ExprDumper stays happy?
> 
> 
> This will not affect those consumers that rely on reimplementing `_after_visit` _alone_, such as the current version of your own `ExprDumper`. More involved subclasses might be out of luck. For nested expressions, there will be no iterated visiting. Notably, those consumers who reimplement `_default_visitor` won't see their custom visitor called for nested expression commands.
> 
> Then, if we're aiming to test against `readelf --debug-dump=info`, I guess I'll have to tweak ExprDumper too, to emit something readelf compatible.
> 
> As a process note, since the library moved forward since I started, I think I should pull the current master and start from scratch.

That sounds like a decent idea, but I'm OK with something even more dramatic, now that I think of it. Maybe we can just replace `GenericExprVisitor` with a function that parses into a nested list, and rewrite `descriptions` to use that. Basically your `parse_expr`.

Then (possibly as a later step, and not mandatory) we can recreate `GenericExprVisitor` in an example (i.e. in `examples`) using this function.

sevaa:
Do we want to keep the `_default_visitor`-like hook for consumers to implement argument parsing for those opcodes that are not supported by pyelftools out of the box?

Do we want to cache the dispatch table for opcodes?

eliben:
> Do we want to keep the `_default_visitor`-like hook for consumers to implement argument parsing for those opcodes that are not supported by pyelftools out of the box?
> 

We can do that, it should be relatively easy. `parse_expr` could have an optional `default_op_handler` function parameter that's invoked on opcodes that aren't in the dispatch table.

> Do we want to cache the dispatch table for opcodes?

Yes. Generally, much of the code in `dwarf_expr.py` can be reused, similarly to your original PR.

eliben:
BTW, sorry for dumping this redesign on you. I can do it myself, but it may take me a few days to get to it. Feel free to just wait for me or if you're motivated, you can go ahead and do it.

sevaa:
> parse_expr could have an optional default_op_handler function parameter that's invoked on opcodes that aren't in the dispatch table.

That was my idea.

>>Do we want to cache the dispatch table for opcodes?
>Yes.

Where? We can't use one global dispatch table, since its contents depend on the structs (note the DWARF32/64 distinction on `DW_OP_call_ref`). Maybe in the CU? Or we can have *two* global dispatch tables - 32 and 64 bit.

Since the expression parser I have in mind is stateless, I'd rather make it a standalone function (maybe with a helper method in the CU) rather than a class.

Sorry for making you design it here :) I don't mind working on it, but you've turned down my previous design, so there...

eliben:
Let me think about this for a bit, I'll get back to you.

eliben:
See #299

PR #289, opened 2020-02-17
readelf.py: adapt section mapping output for .tbss sections

rupran:
GNU readelf does not show the .tbss section as part of the loaded data segment when listing the section to segment mappings, using the ELF_TBSS_SPECIAL macro in include/elf/internal.h to skip printing the section name.

From `include/elf/internal.h` [[0]]: 
```C
/* .tbss is special.  It doesn't contribute memory space to normal
   segments and it doesn't take file space in normal segments.  */
#define ELF_TBSS_SPECIAL(sec_hdr, segment)			\
  (((sec_hdr)->sh_flags & SHF_TLS) != 0				\
   && (sec_hdr)->sh_type == SHT_NOBITS				\
   && (segment)->p_type != PT_TLS)
```
and `binutils/readelf.c` [[1]]:
```C
  for (j = 1; j < filedata->file_header.e_shnum; j++, section++)
    {
      if (!ELF_TBSS_SPECIAL (section, segment)
	  && ELF_SECTION_IN_SEGMENT_STRICT (section, segment))
	printf ("%s ", printable_section_name (filedata, section));
    }

```
Implement the same logic in readelf.py. 

This is related to #275 where Segment.section_in_segment() was fixed for SHF_TLS-flagged sections.

[0]: https://github.com/bminor/binutils-gdb/blob/be70ff166e683fea2d78097d378020a8d3259189/include/elf/internal.h#L305
[1]: https://github.com/bminor/binutils-gdb/blob/be70ff166e683fea2d78097d378020a8d3259189/binutils/readelf.c#L5389


eliben:
Thanks for the contribution!

As I've mentioned elsewhere, I've been really swamped recently and couldn't find much time for pyelftools. I still care! I'm hopefully looking at my schedule clearing at some point not too far in the future, and expect to sweep over the open PRs and issues within the next few weeks.

rupran:
I just added a very simple test file which results in an error without the patch applied.

```
.......................FAIL
....for option "-e"
@@ Mismatch on line #75:
>>   03     .init_array .fini_array .dynamic .got .data .bss <<
>>   03     .tbss .init_array .fini_array .dynamic .got .data .bss <<
 ([('equal', 0, 9, 0, 9), ('insert', 9, 9, 9, 15), ('equal', 9, 59, 15, 65)])
```

With the patch, the test suite passes.

PR #292, opened 2020-03-07
readelf: print addend for RELA relocations without symbol

rupran:
When processing relocations from a SHT_RELA type section, GNU readelf displays the value of the 'r_addend' field if no symbol index is given (that is, 'r_info_sym' is 0).

By also implementing this we can better test the output for 64-bit binaries which commonly use SHT_RELA relocations.

The included test file is the same as tls.elf but compiled for x86_64. Its code is the following:

```C
  __thread int i;

  int main(){}
```

and it is compiled using the following command line:

```
$ gcc -m64 -o tls64.elf tls.c
```


rupran:
Sure, I just added the original tls.c file with comments for the compiler flags.

Issue #288, opened 2020-02-15
off-by-one parsing eh_frame


rhelmot:
When enumerating `elffile.get_dwarf_info().EH_CFI_entries()` for 
[this file](https://github.com/eliben/pyelftools/files/4207796/in.acng.gz), the following error occurs:

```
apt-cacher-ng_2-2_mips64el.deb.data/usr/lib/apt-cacher-ng/in.acng
Traceback (most recent call last):
  File "/home/audrey/angr/pyelftools/elftools/construct/core.py", line 351, in _parse
    return self.packer.unpack(_read_stream(stream, self.length))[0]
  File "/home/audrey/angr/pyelftools/elftools/construct/core.py", line 293, in _read_stream
    raise FieldError("expected %d, found %d" % (length, len(data)))
elftools.construct.core.FieldError: expected 4, found 3

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/audrey/angr/pyelftools/elftools/common/utils.py", line 40, in struct_parse
    return struct.parse_stream(stream)
  File "/home/audrey/angr/pyelftools/elftools/construct/core.py", line 190, in parse_stream
    return self._parse(stream, Container())
  File "/home/audrey/angr/pyelftools/elftools/construct/core.py", line 353, in _parse
    raise FieldError(ex)
elftools.construct.core.FieldError: expected 4, found 3

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./try_load.py", line 14, in <module>
    angr.Project(filename, auto_load_libs=False)
  File "/home/audrey/angr/angr/angr/project.py", line 129, in __init__
    self.loader = cle.Loader(self.filename, concrete_target=concrete_target, **load_options)
  File "/home/audrey/angr/cle/cle/loader.py", line 134, in __init__
    self.initial_load_objects = self._internal_load(main_binary, *preload_libs, *force_load_libs, preloading=(main_binary, *preload_libs))
  File "/home/audrey/angr/cle/cle/loader.py", line 655, in _internal_load
    obj = self._load_object_isolated(main_spec)
  File "/home/audrey/angr/cle/cle/loader.py", line 813, in _load_object_isolated
    return backend_cls(full_spec, is_main_bin=self.main_object is None, loader=self, **options)
  File "/home/audrey/angr/cle/cle/backends/elf/elf.py", line 146, in __init__
    self._load_function_hints_from_fde(dwarf)
[...]

PR #295, opened 2020-03-09
callframe.py: fix DW_EH_PE_absptr decoding

pmderodat:
Hello,

The second commit fixes the bug uncovered by #288 and includes the reproducer as a new testcase, and the first commit extends relocations parsing to handle the special ELF64 MIPS format… so that the second commit’s test works properly.

Thanks,


eliben:
Thanks for the fix! @rhelmot would you mind reviewing, since you opened #288?

rhelmot:
Incredible, the mips64 fixes actually let me remove some hacks from angr that were hacking around the relocation stuff. I can verify that everything loads correctly with this patch.

rhelmot:
Also, to provide attribution: the test file comes from apt-cacher-ng_2-2_mips64el in debian

pmderodat:
Glad to hear that, thank you for letting us know. :-) Do you think we should rename the test file?

rhelmot:
Probably? I don't really care about the name but there might be licensing requirements to distribute it

eliben:
Is there a chance to reproduce the issue in some clean-slate mips64 binary? E.g. concoct a simple C file that - when compiled - would manifest it?

PR #296, opened 2020-03-11
.stack_sizes section support

bigboze:
Just testing the waters here to see if this change is welcome. Added support for dumping the .stack_sizes sections recently added to LLVM. (See https://reviews.llvm.org/D39788). Have locally tested with 32-bit and 64bit ELF files in both endians, but wasn't sure if I should add the ELF and tests (about 8k per file).

Copied the uleb parsing from the DWARF implementations... Should probably refactor them to live elsewhere.


bigboze:
Test binaries are minimal test cases to demonstrate .stack_sizes in 32/64bit ELF files for little and big endian. They were generated using yaml2obj tool found in LLVM project.

eliben:
This new sections seems very LLVM specific, helping them measure/benchmark binary size. Is it supported by any other toolchains? Does readelf dump it?

bigboze:
I think it is currently specific to the LLVM toolchain. It has a range of users of this feature who are interested in the per function stack usage. llvm-readelf has `--stack-sizes - Display contents of all stack sizes sections` to dump it.

eliben:
I think I'd prefer to wait until it gains more usage then.

Many toolchains inject specific ELF sections for various needs, and much of these aren't standard. I wouldn't want the feature set of the library to balloon too much.

PR #297, opened 2020-03-12
Only byteswap the little endian version of mips64 r_raw_info

rhelmot:
Followup to #295. It turns out i was not thorough enough with my testing.

The only test file I have for this change is a libc, which is way bigger than any of the binaries currently in the test files archive. But I can confirm it works, and mirrors the way we hacked around this weird ELF edge case in angr before it was supported here.

### DIFF
--- a/elftools/elf/structs.py
+++ b/elftools/elf/structs.py
@@ -237,6 +237,9 @@ class ELFStructs(object):
 
             def compute_r_info(ctx):
                 raw = ctx['r_raw_info']
+                if not self.little_endian:
+                    return raw
+                # little endian requires an additional byteswap
                 return (((raw & 0xffffffff) << 32)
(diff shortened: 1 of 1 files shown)


eliben:
@pmderodat PTAL if you can, as it's related to #295

pmderodat:
This patch looks correct to me, thanks @rhelmot for submitting it! I had to write the structure layouts in the two endianities on paper to convince myself it was correct, and I’m still thinking about the most readable way to express this logic so that future readers don’t have to scratch their head as much as I am. ;-)

I’m also currently building MIPS64 cross toolchains to produce small testcases for both endianities: I’ll submit them when they are ready. So this patch is ok to merge for me. Thanks again!

eliben:
Thank you, @rhelmot and @pmderodat

PR #300, opened 2020-03-15
Enhance MIPS64 testing and simplify handling code for its peculiar relocations

pmderodat:
Hello  @eliben @rhelmot,

These commits add simple MIPS64 testcases as suggested in #295. Although these new tests cover the relocation decoding enhancements (#295 and #297), they don’t contain an `.eh_frame` section, and thus cannot reproduce the problem that @rhelmot reported initially in #288.

For the record that I managed to make GCC generate this section using the `-funwind-tables` compilation option… but then pyelftools’s readelf could not load these object files anymore, as our handling of MIPS relocations is incomplete. I see three ways out: 1) either somehow link these object file testcases (to an executable or a shared library, similar to the angr reproducer), but for that I’ll likely need to build a libc for MIPS64, 2) extend pyelftools’ support for MIPS relocations, which look quite involved from a quick glance at the spec, or 3) try to reproduce the original bug (#288) with my system compiler (after all, the bug is not MIPS dependent).

I don’t lose hope. ;-) 3) looks like the easiest approach: I’ll keep you updated. In the meantime, I’m submitting these minimal tests anyway, and also the relocations decoding simplification (feedback welcome on whether you find it useful). Thank you both!


eliben:
Thank you @pmderodat 

I'll wait for @rhelmot to review this

rhelmot:
I can confirm that this duplicates the old behavior, which I know is correct, but I don't have any domain-specific knowledge of mips relocations other than "do the tests pass".

eliben:
Thank you both! Yes, relocations are tricky :-)

Issue #302, opened 2020-03-16
Same attribute twice in the DIE


sevaa:
Take a look at the abbrev section of  test/testfiles_for_unittests/dwarf_gnuops1.o in master, declaration 272. According to readelf, it goes:

>    272      DW_TAG_subprogram    [no children]
>     DW_AT_external     DW_FORM_flag_present
>     DW_AT_declaration  DW_FORM_flag_present
>     **DW_AT_linkage_name DW_FORM_strp**
>     DW_AT_name         DW_FORM_strp
>     DW_AT_decl_file    DW_FORM_data1
>     DW_AT_decl_line    DW_FORM_data1
>     **DW_AT_linkage_name DW_FORM_strp**
>     DW_AT value: 0     DW_FORM value: 0

`DW_AT_linkage_name` is present twice. Found a couple of DIEs with that declaraion - the value of `linkage_name` is identical in both places.

`Readelf` has no problem with that, pyelftools presumes unique attribute names.

---

The DWARF in the file is v4. This is a violation of the DWARF standard v4 section 2.2:

>No more than one attribute with a given name may appear in any debugging information entry.

There was a similar bug recorded for GCC: https://gcc.gnu.org/legacy-ml/gcc-bugs/2016-10/msg02517.html . So this looks like a compiler bug too. Still, it would break the idea of testing against readelf. I'll see if I can create a bug for GCC.

EDIT: on a later version of Debian, it doesn't reproduce. Guess they've fixed some time in the middle.

Issue #27, opened 2014-02-26
Disagreement between objdump and pyelftools on DW_TAG_typedef::DW_AT_type


altendky:
I am trying to use pyelftools to parse both TriCore and C166 ELF files (just TriCore for now) to get a list of variables and addresses including being aware of structure members.  The application will be for a remote watch window for the embedded target.  I have [fiddled with the dwarf_die_tree.py example](http://tny.cz/c7174417) but ran into an issue where I am unable to connect between typedef’s and the unnamed structure definitions they reference.  Or so it seems to my DWARF-ignorant brain (in case my previous comments did not in some way make that obvious).

To avoid any issues associated with my particular architecture ELF I also tried my script against test/testfiles_for_unittests/sample_exe64.elf and observed the same thing.  When I run objdump -W as a reference I get, amongst other things:

```
<1><1d6>: Abbrev Number: 3 (DW_TAG_typedef)
    <1d7>   DW_AT_name        : (indirect string, offset: 0xcd): size_t
    <1db>   DW_AT_decl_file   : 2
    <1dc>   DW_AT_decl_line   : 214
    <1dd>   DW_AT_type        : <0x1e1>
```

My pyelftools script results in (again, just a snippet):

```
  DIE tag=DW_TAG_typedef
    Name: size_t
    Offset: 470
    File: 2
    Line: 214
    Type: 63
    Attributes: OrderedDict([('DW_AT_name', AttributeValue(name='DW_AT_name', form='DW_FORM_strp', value=b'size_t', raw_value=205, offset=471)), ('DW_AT_decl_file', AttributeValue(name='DW_AT_decl_file', form='DW_FORM_data1', value=2, raw_value=2, offset=475)), ('DW_AT_decl_line', AttributeValue(name='DW_AT_decl_line', form='DW_FORM_data1', value=214, raw_value=214, offset=476)), ('DW_AT_type', AttributeValue(name='DW_AT_type', form='DW_FORM_ref4', value=63, raw_value=63, offset=477))])
[...]

PR #303, opened 2020-03-18
GNU expressions, take 3

sevaa:
No tests so far.

I've produced a body of ELF files with GNU opcodes by compiling `gdb` on Debian Buster x86_64. It generates DWARF by default. I was going to set up a test against readelf, but then ran into #302 - multiple instances of the same attribute. Dear maintainers, please decide what to do about that.

Alternatively, I can publish a standalone unit test - go over a file or two, parse all expressions, make sure all the opcodes parse.

Alternatively, I can make a custom test where I would match the parsed expressions against the output of readelf, ignoring everything else.

EDIT: I've rebuilt GDB with the latest everything, the duplicate attribute issue is no longer there. HOWEVER, now there's an unsupported relocation type.


eliben:
Hold on, can you clarify the relation to #302. IIUC you closed that issue because it's a DWARF violation by the generating toolchain?

So the only way you can create a reproducer is using a toolchain version with this bug, and hence readelf comparisons fail?

sevaa:
Since then, I've found a repro with a later version of the GCC toolchain that doesn't have the duplicate attribute bug. I mean, the bug was out in the wild, probably for a while, and someone might have affected binaries. But for my particular issue (the GNU opcodes), there's no pressing need to work around the bug.

I'm still getting readelf mismatches, but this time it's pyelftools' fault - the toolchain emits an unsupported x86_64 relocation code :)

eliben:
The code looks reasonable on a first glance. I'd like to see a readelf-comparison test for this, if possible.

sevaa:
Working on that. Right now I'm getting mismatches on `readelf -r` and am not sure what to do about them.

```
.......................FAIL
....for option "-r"
@@ Mismatch on line #27:
>>000000000bc4  02ee0000002a r_x86_64_rex_gotp 0000000000000000 xmalloc - 4<<
>>000000000bc4  02ee0000002a <unknown>         0000000000000000 xmalloc - 4<<
```

eliben:
> Working on that. Right now I'm getting mismatches on `readelf -r` and am not sure what to do about them.
> 
> ```
> .......................FAIL
> ....for option "-r"
> ....Output #1 is readelf, Output #2 is pyelftools
> @@ Mismatch on line #27:
> >>000000000bc4  02ee0000002a r_x86_64_rex_gotp 0000000000000000 xmalloc - 4<<
> >>000000000bc4  02ee0000002a <unknown>         0000000000000000 xmalloc - 4<<
> ```

That seems like a relocation type pyelftool doesn't know about? It may be as easy as checking the code for this relocation and adding it to `ENUM_RELOC_TYPE_x64`? Though it possibly may be a bit more nuanced.

If you don't figure this out within some reasonable time, post that ELF file in a separate issue and I'll try to tackle it.

sevaa:
The `readelf` test is not a particularly good fit for expression parser testing, since a lot of expressions sit in the locations section and `readelf` doesn't dump those. **Ideally, it would help to implement `--debug-dump=loc` in readelf.py and test on that, too.** It might break things - expose gaps in functionality and/or incompatibilities with readelf that went undetected until now.

sevaa:
The tests cover 4 out of 7 GNU opcodes so far. I'll commit some more shortly.

eliben:
> The `readelf` test is not a particularly good fit for expression parser testing, since a lot of expressions sit in the locations section and `readelf` doesn't dump those. **Ideally, it would help to implement `--debug-dump=loc` in readelf.py and test on that, too.** It might break things - expose gaps in functionality and/or incompatibilities with readelf that went undetected until now.

That sounds like a good candidate for a new issue -- I'll be happy to accept contributions that implement this.

sevaa:
Fun fact: I can't readelf-test `loc` independently from this PR, because one of the test files, `improved-dwarfv4.o.elf`, has GNU opcodes under locations :)

eliben:
> Fun fact: I can't readelf-test `loc` independently from this PR, because one of the test files, `improved-dwarfv4.o.elf`, has GNU opcodes under locations :)

Can you please clarify? readelf comparisons of `loc` are necessary to test this PR?

sevaa:
> Oh, so loc tests depend on this PR to fully pass?

Precisely, because at least one file already under test/testfiles_for_readelf **has** the GNU opcodes. The only reason this didn't come up before, our test against readelf doesn't look at DWARF expressions that are hidden in the `locations` section.

But once `readelf.py` supports `--debug-dump=loc`, and `run_readelf_tests.py` includes that in the list of options, it will.

> Would you like to contribute loc comparison tests too?

The code is pretty much ready. But if I send a PR right now, it will fail the regression test, for the reason above.

elftools/dwarf/callframe.py
```diff
@@ -488,7 +490,8 @@ def _decode_CFI_table(self):
         line_stack = []
 
         def _add_to_order(regnum):
-            if regnum not in cur_line:
+            # DW_CFA_restore and others remove registers from cur_line, but they stay in reg_order. Avoid duplicates.
```
eliben: Nit: please reflow to 80 columns

elftools/dwarf/descriptions.py
```diff
@@ -132,7 +132,7 @@ def describe_CFI_CFA_rule(rule):
         return '%s%+d' % (describe_reg_name(rule.reg), rule.offset)
 
 
-def describe_DWARF_expr(expr, structs):
+def describe_DWARF_expr(expr, structs, cu_offset = None):
```
eliben: `cu_offset=None` (no space), and explain it in the comment

elftools/dwarf/descriptions.py
```diff
@@ -539,7 +540,7 @@ def __init__(self, structs):
         self.expr_parser = DWARFExprParser(self.structs)
         self._init_lookups()
 
-    def dump_expr(self, expr):
+    def dump_expr(self, expr, cu_offset = None):
```
eliben: Same comment as above for cu_offset formatting and comment

elftools/dwarf/descriptions.py
```diff
@@ -568,7 +569,10 @@ def _init_lookups(self):
         self._ops_with_hex_arg = set(
             ['DW_OP_addr', 'DW_OP_call2', 'DW_OP_call4', 'DW_OP_call_ref'])
 
-    def _dump_to_string(self, opcode, opcode_name, args):
+    def _dump_to_string(self, opcode, opcode_name, args, cu_offset = None):
```
eliben: `cu_offset=None`

elftools/dwarf/dwarf_expr.py
```diff
@@ -171,9 +178,33 @@ def parse_arg_struct(arg_struct):
     def parse_arg_struct2(arg1_struct, arg2_struct):
         return lambda stream: [struct_parse(arg1_struct, stream),
                                struct_parse(arg2_struct, stream)]
+ 
+    # ULEB128, then an expression of that length
+    def parse_nestedexpr():
+        def parse(stream):
+            size = struct_parse(structs.Dwarf_uleb128(''), stream)
+            nested_expr_blob = stream.read(size)
+            if isinstance(nested_expr_blob, str): # Python 2 patch. In Py3, it comes as bytes.
```
eliben: It's odd that you need this here. Could you use something from py3compat?

elftools/dwarf/dwarf_expr.py
```diff
@@ -171,9 +178,33 @@ def parse_arg_struct(arg_struct):
...
+                nested_expr_blob = [ord(b) for b in nested_expr_blob]
+            return [DWARFExprParser(structs).parse_expr(nested_expr_blob)]
+        return parse
+
+    # In Python 2, read returns a str, in Byton 3 - bytes.
+    # For Python 2, bytearray will do, since bytes is a synonym of str and isn't a collection of ints.
+    def to_blob(blob):
```
eliben: I'm pretty sure you could create a struct in `structs.py` to parse this. There are many examples in `elf/structs.py` and `dwarf/structs.py`. A positive fallout would be that construct is already ported for py2/py3 and you wouldn't need these conditions here.

Generally, in pyelftools all py2/py3 differences are stashed into py3compat or the construct parsing code (except some small things like displaying bytes/string)

elftools/elf/descriptions.py
```diff
@@ -177,9 +177,12 @@ def describe_note(x):
     n_desc = x['n_desc']
     desc = ''
     if x['n_type'] == 'NT_GNU_ABI_TAG':
-        desc = '\n    OS: %s, ABI: %d.%d.%d' % (
-            _DESCR_NOTE_ABI_TAG_OS.get(n_desc['abi_os'], _unknown),
-            n_desc['abi_major'], n_desc['abi_minor'], n_desc['abi_tiny'])
+        if x['n_name'] == 'Android': # Lame patch :(
```
eliben: Is this for readelf compatibility?

There's a lot of code like this in `descriptions` so I'd remove the comment.

test/run_readelf_tests.py
```diff
@@ -48,19 +48,20 @@ def discover_testfiles(rootdir):
             yield os.path.join(rootdir, filename)
 
 
-def run_test_on_file(filename, verbose=False):
+def run_test_on_file(filename, verbose, opt):
```
eliben: Add a comment explaining opt

elftools/elf/descriptions.py
sevaa: Yes, to keep up with readelf :( Removed.

elftools/dwarf/dwarf_expr.py
sevaa: You do it this way:
```
  [struct_parse(structs.Dwarf_uint8(''), stream) for i in range(len)]
```
Getting an array of ints (which I don't think is a good Pythonic data structure for a blob). But there's no helper function to that effect that I can see. I could create one...

sevaa:
How do you tell Github that the requested changes were provided?

eliben:
Just noticing that the new test files (`dwarf_gnuopsX.so.elf`) take a long while to run tests on. Any way they can be made smaller OR just left with the section we want to test here (removing everything else)

sevaa:
**gnuops2 and gnuops3 can safely go away**; between `info` and `loc`, gnuops4 has all 7 GNU opcodes that we now support (it was a HAIRY codebase :)).

That said, I found quite a few minor, unrelated issues in pyelftools by testing on real life binaries as opposed to toy examples. Do you know that readelf sometimes truncates superlong path names? I do now... Or look at that business with exported symbol size.

One more reason I'd like to keep gnuops4 around - I don't think you have Android SOs other than this one in the test file library. New platform, and all that. Android is the most popular Linux distro, technically speaking. In fact, I don't think you have a lot of SO's to begin with - your test library is mostly executables and .o files.

Still, I think I can trim some fluff from gnuops4. And I will do that - once I'm done with locations. Same PR or a new one?

eliben:
Thanks, trimming could help too. You can create a separate PR for all that (if 2 and 3 overlap with 4 in tested features, I'm OK with removing them)

PR #304, opened 2020-03-22
Support for --debug-dump=loc in readelf.py and in the test

sevaa:
The ultimate aim is better testing of the expression dumper.

Some DWARF expressions are not in `.debug_info`, they are in `.debug_loc`. Running `readelf --debug-dump=info` doesn't spell those out, only `readelf --debug-dump=loc` does. This change will make pyelftools' test suite run and compare that option, too.

This doesn't need a dedicated test - running the readelf test on the test files with DWARF in them will pick up this logic automagically.

Location entries of type `BaseAddressEntry` are not supported - we don't have a single test file with those.


scripts/readelf.py
```diff
@@ -1332,6 +1335,58 @@ def _dump_debug_frames_interp(self):
...
+            attr = top_die.attributes
+            if 'DW_AT_low_pc' in attr:
+                return attr['DW_AT_low_pc'].value
+            elif 'DW_AT_entry_pc' in attr:
+                return attr['DW_AT_entry_pc'].value
+            else:
+                raise ValueError("Can't find the base IP (low_pc) for a CU")                    
+
+        di = self._dwarfinfo
+        loc_lists = di.location_lists()
```
eliben: could you do something like 

```
loc_lists = list(di.location_lists() or [])
```
This would save a bunch of lines of logic

scripts/readelf.py
```diff
@@ -1332,6 +1335,58 @@ def _dump_debug_frames_interp(self):
...
+        if loc_lists:
+            loc_lists = [ll for ll in loc_lists.iter_location_lists()]
+            if len(loc_lists) > 0:
+                # To dump a location list, one needs to know the CU.
+                # Scroll through DIEs once, list the known location list offsets
+                CUs = [cu for cu in di.iter_CUs()]
```
eliben: ```CUs = list(di.iter_CUs())```? 

But do you actually use it for random access? Why not iterate over `iter_CUs` directly?

scripts/readelf.py
sevaa: Readelf distinguishes between the cases of no locations section and empty locations section - the latter produces a message.

scripts/readelf.py
sevaa: Good point.

scripts/readelf.py
eliben: OK, early returns are better then. Use `list` anyway instead of list comprehension here. Also, please add a comment explaining that we want to distinguish between these cases. Thanks!

eliben:
Removing two of the large files really helps. I run readelf tests in parallel mode, and with only a single large file its latency is hidden by the other parallel workers so the tests run not much slower than before now

sevaa:
Okay then, I'll leave gnuops4 alone for the time being.

PR #305, opened 2020-03-25
tox: sync python versions with travis settings

vapier:


eliben:
This is a good change to make, but I won't merge it for now because of my own setup. On my machine tox isn't friends with Pythons newer than 3.5, due to an older version of distutils and all that packaging jazz. I could probably figure this out but was postponing it out of lack of time. I need to update to a much newer Ubuntu version anyway.

Sorry for the unsatisfying reply. There's not much difference between 3.5 and 3.8 (that is, code running on the former will run on the latter) so this hasn't been high priority. I do use 3.8 for manual tests.

vapier:
distros aren't shipping python3.5 anymore, so it's not possible to run the tests against py3 out of the box anymore.  Debian Buster (current stable) only has python3.7, and Ubuntu Bionic LTS only has python3.6.

eliben:
I see. Then I'd recommend not to use `tox` at all at this point. I can remove the tox file from the repo if it's a bother for folks, but otherwise feel free to ignore it. It predates Travis tests anyway, and Travis is the definitive CI for pyelftools

vapier:
tox is extremely useful for creating reproducible virtualenvs for people.  it's less of an issue for tox as it doesn't have 3rd party deps, but it still helps with diff python versions.

i'm not sure how deleting tox.ini rather than keeping it in sync with the travis settings is an improvement.  people still want to run against multiple versions quickly.

vapier:
tox supports a `py3` keyword where it'll just select one random version

PR #265, opened 2020-01-23
dwarf .debug_types and .debug_macros

mdmillerii:
Here is a series that adds .debug_types with an api similar to random-die, and also a .debug_macros with an API based on location lists.

Both have readelf based testing, including a new assembly generated dwarf for some macinfo corners.

PR #307, opened 2020-04-09
A few additonal changes

mdmillerii:
Here are 4 additional commits on top of #265 .  The would likely work  against #264 .
* There are only 16 sizes reserved
* Cache the dwarf structs by arg key
-- Avoid generating the same set of struct parsers over and over - two for every CU
-- Saves 2.5% runtime on the full unit tests (ELF and DWARF) so even more for DWARF
* A commit that probably should be part of #264 to use that infrastructure in descriptions.py
* An optional commit that removes an unnecessary assignment.


eliben:
Could you please resolve the conflicts?

eliben:
Also, what would be a good title for this PR? What single piece of functionality is it adding/fixing? Should it be further split to separate PRs?

eliben:
@mdmillerii friendly ping

eliben:
Closing this PR for now. Please reopen if there's renewed interest in working on this.

PR #310, opened 2020-05-03
fix issue in aranges cu_offset_at_addr

paddygord:
if there are aranges for some parts of the binary but not others, incorrect aranges may be returned
I'm not confident in the inclusive-end check, please review carefully :smile:


eliben:
Thanks! And sorry for the delays. Do you have any tests to add here?

paddygord:
@eliben done, I think the tests are fairly thorough

PR #314, opened 2020-05-21
Fix determining PAGESIZE under Jython

mephi42:
Jython has neither `resource` nor `mmap`, therefore just use a
reasonable default.


eliben:
Thanks!

PR #315, opened 2020-05-21
dwarf/dwarf_expr: Add support for DW_OP_GNU_push_tls_address

woodruffw:
Adds support for `DW_OP_GNU_push_tls_address` to the DWARF expression parser.

`DW_OP_form_tls_address` uses `0xe0` (the same as `DW_OP_lo_user`, annoyingly) and has nearly identical semantics to the DWARFv3+ `DW_OP_form_tls_address`. Despite this, it looks like clang prefers to emit `DW_OP_form_tls_address` in some cases, probably because `gdb` has historically had better support for it.

Also fixes the operand decoding of `DW_OP_const8{u,s}`, which was incorrectly using two 32-bit operands instead of a single 64-bit operand (ref: DWARFv4 2.5.1.1).

### DIFF
--- a/elftools/dwarf/descriptions.py
+++ b/elftools/dwarf/descriptions.py
@@ -559,16 +559,15 @@ class ExprDumper(object):
     def _init_lookups(self):
         self._ops_with_decimal_arg = set([
             'DW_OP_const1u', 'DW_OP_const1s', 'DW_OP_const2u', 'DW_OP_const2s',
-            'DW_OP_const4u', 'DW_OP_const4s', 'DW_OP_constu', 'DW_OP_consts',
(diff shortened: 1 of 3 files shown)


woodruffw:
Ran this on an internal corpus and confirmed that it worked as expected.

woodruffw:
> Any way to add a test that would exercise it?

Yep, I can try to create a minimal DWARF sample that exercises it.

woodruffw:
Alright, I went ahead and added some DWARF expression tests that exercise these changes. While writing the tests I noticed that I hadn't updated the DWARF description handling code to reflect the changes I made in `DW_OP_const8*`, so I updated those as well (and added a confirming test).

woodruffw:
Thanks for the merge!

Issue #317, opened 2020-06-05
aarch64 and ppc64 support?


nickdesaulniers:
Hi @eliben !
I wrote [a utility to help debug](https://github.com/ClangBuiltLinux/frame-larger-than) `-Wframe-larger-than=` warnings (which are pretty unhelpful in telling you what particular variables may be causing this, especially when many child calls have been inlined.  Apparently all of that info about local variables in captured in the debug info).

We used it for debugging a warning for a ppc64 Linux kernel build, but it seemed the object file failed to parse.  We [observed](https://groups.google.com/g/clang-built-linux/c/GjVJqQMvjvg/m/45gaOrW0AAAJ) the thrown ElfError: `Unsupported relocation type: 1`.

I haven't dug into reproducing yet, but I wanted to check first if pyelftools supports ppc64? I'm not sure if ppc64 uses a different target triple for endianess or not.  I also haven't dumped the relocation kinds, which may be helpful for this report.

nickdesaulniers:
Looks like ppc is not supported: https://github.com/eliben/pyelftools/blob/430e3a550f9c7b64b4b3630d319442aa79e3db19/elftools/elf/relocation.py#L155-L178

eliben:
Yep, no one added arm64 and ppc support yet. Should be fairly straightforward. Patches welcome!

trevorwhitaker:
Hey @nickdesaulniers , I'm getting the same(ish) error on ppc.

Does the PR address `elftools.common.exceptions.ELFRelocationError: Unsupported relocation type: 38`?

If not, how would I go about patching in support for this?

Thanks!

nickdesaulniers:
@trevorwhitaker sorry, no, I've only had time to implement some aarch64 support.  I haven't done a ton of testing, so I'll bet there's more relocations or endianness issues.  Take a look at #318; though it was helpful IIRC that many relocations were already implemented (there generally are a lot of relocation types for a given ISA target).

trevorwhitaker:
Ok, thanks! I'll take a crack at ppc though I am quite new to this. Could you expand on your last point a bit, I'm not quite sure I understand?

"though it was helpful IIRC that many relocations were already implemented (there generally are a lot of relocation types for a given ISA target)."

nickdesaulniers:
[...]

PR #253, opened 2019-11-27
Basic relocation support added for AArch64 architecture

awerv:
The implementation is mostly based on how the script handles the ARM architecture, however AArch64 may use REL and RELA as well. For the relocation codes and their specification [this](https://static.docs.arm.com/ihi0056/b/IHI0056B_aaelf64.pdf) documentation was used.

PR #318, opened 2020-06-06
initial readelf support for aarch64 little endian

nickdesaulniers:
See also:
https://static.docs.arm.com/ihi0056/b/IHI0056B_aaelf64.pdf
for relocation types and
https://developer.arm.com/docs/ihi0057/c/dwarf-for-the-arm-64-bit-architecture-aarch64-abi-2018q4
for DWARF register names.

TODO:
1. add more register names; there are 128 in total
2. add more relocation types; works for the test case added but I haven't checked a full kernel image yet.
3. big endian support.

I figured this is a good start though.

Issue #317
Link: https://github.com/ClangBuiltLinux/frame-larger-than/issues/4
Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>


nickdesaulniers:
damn, I just saw #253 now...

eliben:
> damn, I just saw #253 now...

Yeah, sorry, I forgot about that one myself. Given that it's old and there's no response I'm fine with you redoing it here - feel free to pick up whatever #253 has that seems interesting and we can close that one after this one lands.

nickdesaulniers:
I think this PR is good to go, there's nothing else from #253 that's useful here.  I can add the SIMD registers or further relocations in follow up PRs.

PR #321, opened 2020-06-18
Fix typo when referencing DW_FORM_ref_addr

gpollo:
Found a quick typo when I was trying to resolve DIEs coming from a `DW_AT_abstract_origin` reference.


gpollo:
I'll check if I can have a simple test case. I'll update this PR later.

eliben:
I'll merge this for now, since the change seems correct. If you find a test case, please open a separate PR

PR #308, opened 2020-04-16
dwarf.CallFrameInfo: Support parsing LSDA pointers from FDEs.

ltfish:
Add support in dwarf.CallFrameInfo to retrieve LSDA pointers for FDEs.

PR #323, opened 2020-07-07
Fix Python 2 support of FDE.

ltfish:
My previous PR (#308) broke Python 2 support by omitting arguments to `super()`. This commit fixes it.

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -636,7 +636,7 @@ class CIE(CFIEntry):
 
 class FDE(CFIEntry):
     def __init__(self, header, structs, instructions, offset, augmentation_bytes=None, cie=None, lsda_pointer=None):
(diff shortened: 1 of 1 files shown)


eliben:
Please fix the Travis CI for this PR

eliben:
> Thanks for the fast response!
> 
> I saw the failing test cases, and it seems to me that the failure has nothing to do with this PR. In fact, the test case fails because it is _non-deterministic_: Both `DW_OP_lo_user` and `DW_OP_GNU_push_tls_address` are 0xe0, so `dump_expr()` on 0xe0 may return one of the two. It is merely an unfortunately constant value collision.
> 
> I think the best way is to fix the failed test cases and make them accept both `DW_OP_lo_user` and `DW_OP_GNU_push_tls_address`. I will go ahead and submit a new PR to fix the two failing test cases if you agree with the route going forward.

Sounds good, thank you.

PR #326, opened 2020-07-13
Return the correct number of program headers when e_phnum is 0xffff

pagabuc:
While using pyelftools to parse a QEMU memory dump (ELF corefile), I noticed that pyelftools returned only the first 65535 program headers, while readelf returned 65843. 
Per specification:
"If the number of program headers is greater than or equal to PN_XNUM (0xffff), this member has the value PN_XNUM (0xffff). The actual number of program header table entries is contained in the sh_info field of the section header at index 0."

This PR should implement this behaviour and fix the problem.

In case you want to test the code, [here](https://github.com/eliben/pyelftools/files/4915397/sample0.min.gz) you can find the corefile!

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -120,7 +120,10 @@ class ELFFile(object):
     def num_segments(self):
         """ Number of segments in the file
         """
-        return self['e_phnum']
+        if self['e_phnum'] < 0xffff:
+            return self['e_phnum']
+        else:
+            return self.get_section(0)['sh_info']
 
     def get_segment(self, n):
         """ Get the segment at index #n from the file (Segment object)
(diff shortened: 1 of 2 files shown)


pagabuc:
Ok! The smallest reproducer file I have been able to create is around 3.5 MB. Should I add it as is, or you prefer a gzipped version (152KB)?

elftools/elf/elffile.py
```diff
@@ -120,7 +120,10 @@ def iter_sections(self):
     def num_segments(self):
         """ Number of segments in the file
         """
-        return self['e_phnum']
+        if self['e_phnum'] < 0xffff:
```
eliben, line 123: Could you please add a comment here with the relevant text of the spec and a link to the spec section?

eliben:
> Ok! The smallest reproducer file I have been able to create is around 3.5 MB. Should I add it as is, or you prefer a gzipped version (152KB)?

I see... Let's leave it out for now, then. I realize it's difficult to test with a small file - after all, this is a special case for large header tables. I'll see if I can cook up something easier later on. Please address my latest PR review and then I'll merge this.

PR #327, opened 2020-07-22
Update code to work with pickling

valdegtyarev:
When pickling objects using the "dill" package, the pickler does not know how to handle singletons and will throw an error. The __reduce__ method I added on 1300 tells the pickler how to handle the object's serialization. After I was able to pickle the object, I noticed that I was having a few issues with unpickling. It seemed as if there might have been a small issue with how the parameters were being passed in on line 348. The error message said it required two arguments, not one. By adding the self argument, it managed to fix the problems.


eliben:
Thanks!

PR #328, opened 2020-07-24
Add support for ARM exception handler ABI

LeadroyaL:
Dwarf and ehabi are two different unwind model. ARM ELF file use `.ARM.exidx` and `.ARM.extab` secions to unwind stack trace. The description of data can be shown with command binutils `readelf -u filename`. Pyelftools doesn't parse these two sections, which is reason for this pull request.

Like `DWARFInfo`, `EHABIInfo` is also independent with `ELFFile`. It just need initialized by a `SHT_ARM_EXIDX` section.

Also I add argument `-au`, `--arm_unwind` in `readelf.py`. I cannot use `-u` because this pull request only compatible for ARM.

Referense:
https://github.com/llvm/llvm-project/blob/master/llvm/tools/llvm-readobj/ARMEHABIPrinter.h
https://developer.arm.com/documentation/ihi0038/b/
binutils: readelf.c


eliben:
Thank you for the contribution. I'm not much familiar with ARM and its EHABI. CCing some folks who contributed ARM stuff before who could help review this

@frederiksdun @nickdesaulniers

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+            if GPRMask & (1 << i) != 0:
+                hits.append(self.GPRRegisterNames[i])
+        return '{%s}' % ', '.join(hits)
+
+    def _PrintRegisters(self, VFPMask, Prefix):
+        hits = []
+        for i in range(32):
+            if VFPMask & (1 << i) != 0:
+                hits.append(Prefix + str(i))
+        return '{%s}' % ', '.join(hits)
```
nickdesaulniers, line 84: Does this fit 80 lines as an array comprehension?
```python
x = [Prefix + str(i) for i in range(32) if VFPMask & (1 << i) != 0]
return '{%s} % ', '.join(x)
```

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        #   SW.startLine()
+        #     << format("0x%02X 0x%02X ; %s",
+        #               Opcode0, Opcode1, GPRMask ? "pop " : "refuse to unwind");
+        #   if (GPRMask)
+        #     PrintGPR(GPRMask);
+        GPRMask = (op1 << 4) | ((op0 & 0x0f) << 12)
+        if GPRMask == 0:
+            return 'refuse to unwind'
+        else:
+            return 'pop %s' % self._PrintGPR(GPRMask)
```
nickdesaulniers, line 101: Is implicit conversion to true preferred?
```python
if GPRMask:
    return 'pop %s' % self._PrintGPR(GPRMask)
else:
    return 'refuse to unwind'

File: elftools/ehabi/decoder.py [hunk 3]
```diff
@@ -0,0 +1,271 @@
...
+
+    def _Decode_11001yyy(self):
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'spare'
+
+    def _Decode_11000nnn(self):
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._PrintRegisters((((1 << ((opcode & 0x07) + 1)) - 1) << 10), "wR")
```
nickdesaulniers, line 224: Want to comment the line from the LLVM decoder, as the other methods have?

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def _Decode_11010nnn(self):
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._PrintRegisters((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
```
nickdesaulniers, line 229: Decode comment?

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,217 @@
...
+def arm_expand_prel31(address, place):
+    """
+       address: uint32
+       place: uint32
+       return: uint64
+    """
+    Location = address & 0x7fffffff
+    if Location & 0x04000000:
+        Location |= 0xffffffff80000000
+    return Location + place & 0xffffffffffffffff
```
nickdesaulniers, line 217: I suspect there's already code in tree for decoding this type of relocation.

I think you should add an entry for `R_ARM_PREL31` to `_RELOCATION_RECIPES_ARM ` in elf/relocation.py, then use...hmm...that might not be as straightforward as I thought.

Anyways, I don't think the local variable `Location` should be capitalized. Also, `offset` would be more appropriate than `place` for an identifier.

elftools/ehabi/decoder.py
nickdesaulniers: Same with `_PrintGPR`.

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        for b in reversed(uleb_buffer):
+            value = (value << 7) + (b & 0x7F)
+        return 'vsp = vsp + %u' % (0x204 + (value << 2))
+
+    def _Decode_10110011_sssscccc(self):
+        # these two decoders are equal
+        return self._Decode_11001001_sssscccc()
+
+    def _Decode_101101nn(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 166: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        op0 = self._bytecode_array[self._index]
+        self._index += 1
+        op1 = self._bytecode_array[self._index]
+        self._index += 1
+        Start = ((op1 & 0xf0) >> 4)
+        Count = ((op1 & 0x0f) >> 0)
+        return 'pop %s' % self._PrintRegisters((((1 << (Count + 1)) - 1) << Start), "d")
+
+    def _Decode_11001yyy(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 217: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        self._index += 1
+        return 'pop %s' % self._PrintRegisters((((1 << ((opcode & 0x07) + 1)) - 1) << 10), "wR")
+
+    def _Decode_11010nnn(self):
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._PrintRegisters((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
+
+    def _Decode_11xxxyyy(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 232: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def _Decode_10101nnn(self):
+        # SW.startLine() << format("0x%02X      ; pop ", Opcode);
+        # PrintGPR((((1 << ((Opcode & 0x7) + 1)) - 1) << 4) | (1 << 14));
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._PrintGPR((((1 << ((opcode & 0x7) + 1)) - 1) << 4) | (1 << 14))
+
+    def _Decode_10110000(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 134: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        else:
+            return 'pop %s' % self._PrintGPR(GPRMask)
+
+    def _Decode_10011101(self):
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'reserved (ARM MOVrr)'
+
+    def _Decode_10011111(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 109: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+        #   if (GPRMask)
+        #     PrintGPR(GPRMask);
+        GPRMask = (op1 << 4) | ((op0 & 0x0f) << 12)
+        if GPRMask == 0:
+            return 'refuse to unwind'
+        else:
+            return 'pop %s' % self._PrintGPR(GPRMask)
+
+    def _Decode_10011101(self):
+        opcode = self._bytecode_array[self._index]
```
nickdesaulniers, line 104: dead store

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,271 @@
...
+                bool, whether bytecode should be decoded during construction.
+                (default : true)
+    """
+
+    def __init__(self, bytecode_array, auto_decode=True):
+        self._bytecode_array = bytecode_array
+        self._index = None
+        self.mnemonic_array = None
+        if auto_decode:
+            self.decode()
```
nickdesaulniers, line 39: I feel like `decode` should just be merged into `__init__`.  You're probably not going to redecode an `EHABIBytecodeDecoder` once constructed.

test/test_ehabi_elf.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    def test_parse_object_file(self):
+        # FIXME: `.ARM.exidx.text.XXX` need relocation, it's too complex for current unittest.
+        fname = os.path.join('test', 'testfiles_for_unittests', 'arm_exidx_test.o')
+        with open(fname, 'rb') as f:
+            elf = ELFFile(f)
+            ehabi_infos = elf.get_ehabi_infos()
+            self.assertEqual(0, len(ehabi_infos))
+            # TODO
```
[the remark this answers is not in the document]
LeadroyaL: Same with "# TODO: support relocatable file".

elftools/elf/elffile.py
```diff
@@ -227,6 +228,26 @@ def get_dwarf_info(self, relocate_dwarf_sections=True):
...
+
+    def get_ehabi_infos(self):
+        """ Generally, shared library and executable contain 1 .ARM.exidx section.
+            Object file contains many .ARM.exidx sections.
+            So we must traverse every section and filter sections which type is SHT_ARM_EXIDX.
+        """
+        _ret = []
+        if self.header['e_type'] == 'ET_REL':
+            # TODO: support relocatable file
+            print ("Current version of pyelftools doesn't support relocatable file.")
```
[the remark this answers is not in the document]
LeadroyaL: In shared library file, there is no relocation in `.ARM.exidx` and `.ARM.extab`. 
In object file, there is one `.ARM.exidx.FUNC_NAME`, one `.rel.ARM.exidx.FUNC_NAME`, one `.ARM.extab.FUNC_NAME` and one `.rel.ARM.extab.FUNC_NAME` for every function.

I'm not familiar with relocation of object file. So it isn't general limitation, it's my limitation.

elftools/ehabi/decoder.py
LeadroyaL: I think you misunderstand `C` and `Python`. 
Type of `GPRMask` is int. ` if GPRMask` always return TRUE, it's different from `if GPRMask == 0`.

elftools/ehabi/ehabiinfo.py
LeadroyaL: First, `elf/relocation.py` is too heavy for me. It provides relocation for whole elf file, rather than single entry.

Second, I think `R_ARM_PREL31` and `arm_expand_prel31` are two different concepts. `R_ARM_PREL31` is relocation type in `.rel.ARM.exidx` section.  `arm_expand_prel31` is like an encoding format.

`Location` is my fault, I just copy the code from binutils/readelf.c.

elftools/ehabi/decoder.py
LeadroyaL: I use `auto_decode` for developer who doesn't care the disasm result of bytecode. It's ok to merge, but I prefer to give the choice to developer.

LeadroyaL:
> Should the tests have the source file that produced them, as well as comments on how to rebuild the test object files?

You're right. I'll try to add the testcase from LLVM-readelf and produce them later.

> Looking at some of the bitwise operations, I'm curious about the perennial question for bi-endian architectures; does this work for big endian (compile with -mbig-endian). If yes; please add little and big endian binaries. If not, well I guess little endian is more standard, and something is better than nothing.

Sorry I forget to handle endian,  I'll fix it later also.

nickdesaulniers:
The implementation LGTM; just style nits that `elftools/ehabi/decoder.py` has way too many UpperCamelCased local variable identifiers.  I think those should only be used for classes. (Github highlights them in orange).  Some methods also use _UpperCamelCase, and I think those should be _snake_case.  PEP8 says:

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,212 @@
...
+        return self._num_entry
+
+    def get_entry(self, n):
+        """ Get the exception handler entry at index #n. (EHABIEntry object or a subclass)
+        """
+        if n >= self.num_entry():
+            raise IndexError('Invalid entry %d/%d' % (n, self._num_entry))
+        eh_index_entry_offset = self._arm_idx_section['sh_offset'] + n * EHABI_INDEX_ENTRY_SIZE
+        eh_index_data = struct_parse(self._struct.EH_index_struct, self._arm_idx_section.stream, eh_index_entry_offset)
+        Word0, Word1 = eh_index_data['Word0'], eh_index_data['Word1']
```
nickdesaulniers, line 54: `word0`, `word1`

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+        return 'vsp = vsp + %u' % (((opcode & 0x3f) << 2) + 4)
+
+    def _decode_01xxxxxx(self):
+        # SW.startLine() << format("0x%02X      ; vsp = vsp - %u\n", Opcode,
+        #                          ((Opcode & 0x3f) << 2) + 4);
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'vsp = vsp - %u' % (((opcode & 0x3f) << 2) + 4)
+
+    GPR_register_names = ("r0", "r1", "r2", "r3", "r4", "r5", "r6", "r7",
```
nickdesaulniers, line 69: `gpr_register_names`

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+                        "r8", "r9", "r10", "fp", "ip", "sp", "lr", "pc")
+
+    def _printGPR(self, GPR_mask):
+        hits = [self.GPR_register_names[i] for i in range(32) if GPR_mask & (1 << i) != 0]
```
nickdesaulniers, line 73: `gpr_mask`

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+        return '{%s}' % ', '.join(hits)
+
+    def _print_registers(self, VFP_mask, prefix):
+        hits = [prefix + str(i) for i in range(32) if VFP_mask & (1 << i) != 0]
```
nickdesaulniers, line 77: `vfp_mask`

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+        #   SW.startLine()
+        #     << format("0x%02X 0x%02X ; %s",
+        #               Opcode0, Opcode1, GPRMask ? "pop " : "refuse to unwind");
+        #   if (GPRMask)
+        #     PrintGPR(GPRMask);
+        GPR_mask = (op1 << 4) | ((op0 & 0x0f) << 12)
+        if GPR_mask == 0:
+            return 'refuse to unwind'
+        else:
+            return 'pop %s' % self._printGPR(GPR_mask)
```
nickdesaulniers, line 95: `gpr_mask`

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if Word0 & 0x80000000 != 0:
+            print ('Corrupt ARM exception handler table entry: %x' % n)
+            return CorruptEHABIEntry()
+
+        function_offset = arm_expand_prel31(Word0, self._arm_idx_section['sh_offset'] + n * EHABI_INDEX_ENTRY_SIZE)
```
nickdesaulniers, line 60: can we use `self.section_offset()` in this expression?

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,212 @@
...
+
+        if Word1 == 1:
+            # 0x1 means cannot unwind
+            return CannotUnwindEHABIEntry(function_offset)
+        elif Word1 & 0x80000000 == 0:
+            # highest bit is zero, point to .ARM.extab data
+            eh_table_offset = arm_expand_prel31(Word1, self._arm_idx_section['sh_offset'] + n * EHABI_INDEX_ENTRY_SIZE + 4)
```
nickdesaulniers, line 67: can we use `self.section_offset()` in this expression?

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,212 @@
...
+                    # arm compact model 1/2
+                    more_word = (Word0 >> 16) & 0xff
+                    opcode = [(Word0 >> 8) & 0xff, (Word0 >> 0) & 0xff]
+                    self._arm_idx_section.stream.seek(eh_table_offset + 4)
+                    for i in range(more_word):
+                        r = struct_parse(self._struct.EH_table_struct, self._arm_idx_section.stream)['Word0']
+                        opcode.append((r >> 24) & 0xFF)
+                        opcode.append((r >> 16) & 0xFF)
+                        opcode.append((r >> 8) & 0xFF)
+                        opcode.append((r >> 0) & 0xFF)
```
nickdesaulniers, line 94: these patterns are generally problematic for bi-endianness.

elftools/ehabi/decoder.py
nickdesaulniers: 👍

elftools/ehabi/ehabiinfo.py
LeadroyaL: Your concern is unnecessary, I've handled endianness compatibility with code:

```
struct_parse(self._struct.EH_table_struct, self._arm_idx_section.stream)
```

The `opcode` variable is uint32 in correct endianness.

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+
+    def __init__(self,
+                 function_offset,
+                 personality,
+                 bytecode_array,
+                 eh_table_offset=None,
+                 unwindable=True,
+                 corrupt=False
+                 ):
```
nickdesaulniers, line 146: I find this line wrapping a little odd, compared to methods like `CorruptEHABIEntry` that seem to wrap somewhere between 80-100 chars.  Maybe I'm just on a smaller screen now,  but `__init__` vs other places has what seems like inconsistent line wrapping.

elftools/ehabi/ehabiinfo.py
nickdesaulniers: Would you please:
1. add the C source file that produced the test binaries.
2. add binaries compiled with `-mbig-endian` to prove the code handles bi-endianness.

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+    def _decode_11010nnn(self):
+        #   SW.startLine() << format("0x%02X      ; pop ", Opcode);
+        #   PrintRegisters((((1 << ((Opcode & 0x07) + 1)) - 1) << 8), "d");
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._print_registers((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
+
+    def _decode_11xxxyyy(self):
+        self._index += 1
+        return 'spare'
```
nickdesaulniers, line 247: same implementation as `_decode_11001yyy`. Combine them?

nickdesaulniers: and `_decode_101101nn`.

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,284 @@
...
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._print_registers((((1 << ((opcode & 0x07) + 1)) - 1) << 10), "wR")
+
+    def _decode_11010nnn(self):
+        #   SW.startLine() << format("0x%02X      ; pop ", Opcode);
+        #   PrintRegisters((((1 << ((Opcode & 0x07) + 1)) - 1) << 8), "d");
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._print_registers((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
```
nickdesaulniers, line 243: I see this subexpression a lot:
`self._print_registers((((1 << ((X) + 1)) - 1) << Y)`
Do you think creating a helper method that accepts parameters `X`, and `Y` would help [DRY up](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this code?

elftools/ehabi/ehabiinfo.py
LeadroyaL: I designed in this format just for readable, rather than for code style. Do you think it really matters?

elftools/ehabi/ehabiinfo.py
LeadroyaL: > add the C source file that produced the test binaries.

Done, the C source is very simple. `arm_exidx_test.cpp`.

> add binaries compiled with -mbig-endian to prove the code handles bi-endianness.

I've tried to compile and produce a big-endian elf, but I failed. 

Try1:  I'm an ARM developer with android-ndk, when I compile from C to shared library or executable, link error occurs. I notice that android-ndk doesn't provide big-endian environment.
Try2: I use `apt install gcc-arm-linux-gnueabi` and cross compile. Link error also occurs.

However, I convince that my code is compatible and do that test: I compare the `bytecode_array` between `self.little_endian=True` and `self.little_endian=False`, they are invert mutually.

elftools/ehabi/decoder.py
LeadroyaL: I reviewed every cross reference of `_print_registers` and `_print_GPR`. Most of the input parameter are different. Only 3 of them are the same. I think it's unnecessary to create sub_method for them.

BTW, I would like to follow the computing from `llvm/tools/llvm-readobj`.

elftools/ehabi/ehabiinfo.py
nickdesaulniers: Did you forget to `git add` the test `arm_exidx_test.cpp`?

nickdesaulniers: Indeed, the NDK doesn't distribute the C runtime objects necessary for linking a cross compiled executable targeting arm big endian.  Ok, that

elftools/ehabi/ehabiinfo.py
nickdesaulniers: maybe @eliben has thoughts?

elftools/ehabi/ehabiinfo.py
LeadroyaL: > Did you forget to git add the test arm_exidx_test.cpp?

Sorry. Attached now.

> Though you can skip linking by just compiling object files with -c.

You're right. But, this is the limitation we discussed before, current version doesn't support relocatable object file.

elftools/ehabi/ehabiinfo.py
eliben: pyelftools tries to roughly follow https://www.python.org/dev/peps/pep-0008/ and be consistent. I think this wrapping is OK overall, but do place the closing `):` higher up (`corrupt=False):`), since the declaration is already visually distinct from the code below it (different indentation level, following the "wrapping" advice in PEP 8)

elftools/ehabi/constants.py
```diff
@@ -0,0 +1 @@
+EHABI_INDEX_ENTRY_SIZE = 8
```
eliben, line 1: Nit: please end the file/line with a newline

elftools/ehabi/decoder.py
eliben: Can you elaborate on the scenario in which a developer would want to create an `EHABIBytecodeDecoder` without decoding the bytecode?

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        self._index = 0
+        self.mnemonic_array = []
+        while self._index < len(self._bytecode_array):
+            for mask, value, _decode_handler in self.ring:
+                if (self._bytecode_array[self._index] & mask) == value:
+                    start_idx = self._index
+                    mnemonic = _decode_handler(self)
+                    end_idx = self._index
+                    self.mnemonic_array.append(MnemonicItem(self._bytecode_array[start_idx: end_idx], mnemonic))
```
eliben, line 52: wrap to 80 chars (or very close to it)

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,285 @@
...
+        opcode = self._bytecode_array[self._index]
+        self._index += 1
+        return 'pop %s' % self._print_registers((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
+
+    def _decode_11000110_sssscccc(self):
+        #  SW.startLine() << format("0x%02X 0x%02X ; pop ", Opcode0, Opcode1);
+        #  uint8_t Start = ((Opcode1 & 0xf0) >> 4);
+        #  uint8_t Count = ((Opcode1 & 0x0f) >> 0);
+        #  PrintRegisters((((1 << (Count + 1)) - 1) << Start), "wR");
+        self._index += 1
```
eliben, line 181: In some `_decode`s you first grab a byte from `self._index` and then increment it, and in some you first increment it. Can you add a comment explaining the two scenarios?

elftools/ehabi/decoder.py
```diff
@@ -0,0 +1,285 @@
...
+        return 'pop %s' % self._print_registers((((1 << ((opcode & 0x07) + 1)) - 1) << 8), "d")
+
+    def _decode_11xxxyyy(self):
+        return self._spare()
+
+    def _spare(self):
+        self._index += 1
+        return 'spare'
+
+    ring = (
```
eliben, line 250: Consider using a `namedtuple` for these - then the field names are self-documention. `namedtuple` is used in many places throughout pyelftools

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,211 @@
...
+            eh_index_data = struct_parse(self._struct.EH_table_struct, self._arm_idx_section.stream, eh_table_offset)
+            word0 = eh_index_data['word0']
+            if word0 & 0x80000000 == 0:
+                # highest bit is one, generic model
+                return GenericEHABIEntry(function_offset, arm_expand_prel31(word0, eh_table_offset))
+            else:
+                # highest bit is one, arm compact model
+                # highest half must be 0b1000 for compact model
+                if word0 & 0x70000000 != 0:
+                    print ('Corrupt ARM compact model table entry: %x' % n)
```
eliben, line 77: There should not be `print`s this deep in the library

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,211 @@
...
+                        opcode.append((r >> 8) & 0xFF)
+                        opcode.append((r >> 0) & 0xFF)
+                    return EHABIEntry(function_offset, per_index, opcode, eh_table_offset=eh_table_offset)
+                else:
+                    print ('Unknown ARM compact model %d at table entry: %x' % (per_index, n))
+                    return CorruptEHABIEntry()
+        else:
+            # highest bit is one, compact model must be 0
+            if word1 & 0x7f000000 != 0:
+                print('Corrupt ARM compact model table entry: %x' % n)
```
eliben, line 102: Ditto re 'print' here and elsewhere

elftools/ehabi/ehabiinfo.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class EHABIEntry(object):
+    """ Exception handler abi entry.
+
+        Accessible attributes:
+
+            function_offset:
+                Integer.
+                Return None when corrupt.
```
eliben, line 115: I'm not sure what this "return None when corrupt" comment refers to. This attribute is set in the constructor, no? Maybe it would be clearer as "None if corrupt"? Same below for personality and others

elftools/elf/elffile.py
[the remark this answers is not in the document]
eliben: Here too, don't "print"

test/testfiles_for_unittests/arm_exidx_test.cpp
```diff
@@ -0,0 +1,23 @@
...
+void func2(int i) {
+    if (i == 0)
+        return;
+    func1(i - 1);
+}
+
+int main(int argc, char **argv) {
+    std::string hello = "Hello from C++";
+    std::cout << hello << std::endl;
+}
```
eliben, line 23: Nit: end file with newline

eliben:
Just a note that I'll be offline for a week or so, so this gives you plenty of time to address the (mostly minor) comments. I'll re-review and merge when I'm back

elftools/ehabi/decoder.py
LeadroyaL: I consider deeply and cannot imagine that scenario, I'll remove `auto_decode`. Thanks.

elftools/ehabi/ehabiinfo.py
LeadroyaL: There are several sub-class of `EHABIEntry`, such as `CorruptEHABIEntry`, `CannotUnwindEHABIEntry`. They will set `function_offset` and other attribute to `None`.

Maybe you think my comment is strange, but it just comes from ARM unwind documentation. I'll make more comment about this.

> Maybe it would be clearer as "None if corrupt"

It's better, thanks.

PR #331, opened 2020-09-22
added a method for returning the index of a section by name

JonBruchim:
Moved section to map to another function
Created an method to return section index by section name
Signed-off-by: Jonathan <yonbruchim@gmail.com>

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -106,14 +106,9 @@ class ELFFile(object):
         # mapping
         #
         if self._section_name_map is None:
-            self._make_section_to_name_map()
+            self._make_section_name_map()
         secnum = self._section_name_map.get(name, None)
         return None if secnum is None else self.get_section(secnum)
-
-    def _make_section_to_name_map(self):
-        self._section_name_map = {}
-        for i, sec in enumerate(self.iter_sections()):
-            self._section_name_map[sec.name] = i
 
     def get_section_index(self, section_name):
         """ Gets the index of the section by name. Return None if no such
@@ -123,7 +118,7 @@ class ELFFile(object):
         # mapping
         #
         if self._section_name_map is None:
-            self._make_section_to_name_map()
+            self._make_section_name_map()
         if not self._section_name_map.has_key(section_name):
             return None
         return self._section_name_map[section_name]
@@ -560,6 +555,11 @@ class ELFFile(object):
             return self._make_gnu_hash_section(section_header, name)
         else:
             return Section(section_header, name, self)
+
+    def _make_section_name_map(self):
+        self._section_name_map = {}
+        for i, sec in enumerate(self.iter_sections()):
+            self._section_name_map[sec.name] = i
 
     def _make_symbol_table_section(self, section_header, name):
         """ Create a SymbolTableSection
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -106,11 +106,27 @@ class ELFFile(object):
         # mapping
         #
         if self._section_name_map is None:
-            self._section_name_map = {}
-            for i, sec in enumerate(self.iter_sections()):
-                self._section_name_map[sec.name] = i
+            self._make_section_to_name_map()
         secnum = self._section_name_map.get(name, None)
         return None if secnum is None else self.get_section(secnum)
+
+    def _make_section_to_name_map(self):
+        self._section_name_map = {}
+        for i, sec in enumerate(self.iter_sections()):
+            self._section_name_map[sec.name] = i
+
+    def get_section_index(self, section_name):
+        """ Gets the index of the section by name. Return None if no such
+            section name exists.
+        """
+        # The first time this method is called, construct a name to number
+        # mapping
+        #
+        if self._section_name_map is None:
+            self._make_section_to_name_map()
+        if not self._section_name_map.has_key(section_name):
+            return None
+        return self._section_name_map[section_name]
 
     def iter_sections(self):
         """ Yield all the sections in the file
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -119,9 +119,7 @@ class ELFFile(object):
         #
         if self._section_name_map is None:
             self._make_section_name_map()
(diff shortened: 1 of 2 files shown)


elftools/elf/elffile.py
```diff
@@ -106,12 +106,23 @@ def get_section_by_name(self, name):
...
+            section name exists.
+        """
+        # The first time this method is called, construct a name to number
+        # mapping
+        #
+        if self._section_name_map is None:
+            self._make_section_name_map()
+        if not self._section_name_map.has_key(section_name):
+            return None
+        return self._section_name_map[section_name]
```
rupran, line 124: I think you could put this into a single statement (like in the function above, line 110):

```python
return self._section_name_map.get(section_name, None)
```

rupran (commented):
I noticed that LLVM/clang can emit multiple sections with the same name in order to save space in the resulting binary when -ffunction-sections is used. The current code also doesn't handle this case, so using get_section_by_name() on such an object file will always return the last section with the non-unique name. See the following example:

```
$ cat two_f.c
void a(){}
void b(){}

$ clang-9 -ffunction-sections -fno-unique-section-names -c -o two_f.o two_f.c

$ readelf -S two_f.o
There are 11 section headers, starting at offset 0x208:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .strtab           STRTAB           0000000000000000  000001b0
       0000000000000057  0000000000000000           0     0     1
  [ 2] .text             PROGBITS         0000000000000000  00000040
       0000000000000000  0000000000000000  AX       0     0     4
  [ 3] .text             PROGBITS         0000000000000000  00000040
       0000000000000006  0000000000000000  AX       0     0     16
  [ 4] .text             PROGBITS         0000000000000000  00000050
       0000000000000006  0000000000000000  AX       0     0     16
[...]

$ python3
Python 3.6.9 (default, Jul 17 2020, 12:50:27)
[GCC 8.4.0] on linux
>>> from elftools.elf.elffile import ELFFile
>>> elf = ELFFile(open('two_f.o', 'rb'))
>>> print(elf.get_section_by_name('.text').header)
Container({'sh_name': 1, 'sh_type': 'SHT_PROGBITS', 'sh_flags': 6, 'sh_addr': 0, 'sh_offset': 80, 'sh_size': 6, 'sh_link': 0, 'sh_info': 0, 'sh_addralign': 16, 'sh_entsize': 0})
```

In the same manner, your code will always return the last index (in our example above 4) for '.text'.

I'm not sure if that's something that needs general fixing but I would like to keep that in mind.

Additionally, it would be nice to have some test files for your code, be it with non-unique section names or not.

JonBruchim:
Thank you for the reply,
Actually not that familiar with LLVM, good to know!

I've added some tests.

JonBruchim:
@rupran anything else I need to do?

rupran:
From my side this looks good.

@eliben, you want to take a look as well?

eliben:
I wonder what is the use case here?

The need hasn't come up before, and if someone does need it, it seems straightforward to build the mapping outside pyelftools. It's just a simple convenience around `iter_sections`, it seems. What am I missing?

Theoretically we could add similar convenience methods for many other things, but why?

The multiple-sections-with-same-name issue @rupran mentioned casts another light on this; it's not clear what the right policy is in that case, since the `(section name, section index)` function is not injective.

JonBruchim:
@eliben I've initially built a mapping outside of pyelftools, but as I've looked into the source code, I've noticed that its basically implemented.
Just need to create a function for it.
As you could see, I'm using an existing map.

I'd be glad to help implementing other convenient methods as this one =)

eliben:
> @eliben I've initially built a mapping outside of pyelftools, 

Can you share more about your use case? Why did you need to map sections from names to indices in the file, and how did you use it?

> but as I've looked into the source code, I've noticed that its basically implemented.
> Just need to create a function for it.
> As you could see, I'm using an existing map.

The fact that the internal implementation includes a data structure doesn't mean that this data structure should be exposed in a public API. Such exposure makes it more difficult to change the internal representation in the future.

JonBruchim:
@eliben I've wanted to map an section to a symbol.
wanted to map data/text/bss (etc.) to its symbols

eliben:
> @eliben I've wanted to map an section to a symbol.
> wanted to map data/text/bss (etc.) to its symbols

Could you elaborate? It's not clear to me how you're using a section name -> index mapping for this. Could you provide a usage code sample?

JonBruchim:
> > @eliben I've wanted to map an section to a symbol.
> > wanted to map data/text/bss (etc.) to its symbols
> 
> Could you elaborate? It's not clear to me how you're using a section name -> index mapping for this. Could you provide a usage code sample?

Sorry for the long wait.
Unfortunately the original code is in a private network but I've written a piece of code to show usage.


`from elftools.elf.elffile import ELFFile
from elftools.elf.sections import SymbolTableSection

def map_index_to_section_name(elf, dic):
	for i, section in enumerate(elf.iter_sections()):
		dic[i] = section.name

def main():
	sections = {".data": ([], "STT_OBJECT"),
				".bss": ([], "STT_OBJECT"),
				".text": ([], "STT_FUNC"),
				".rodata": ([], "STT_OBJECT")}
	index_to_name = {}

	with open("a.out") as f:
		elf = ELFFile(f)
		map_index_to_section_name(elf, index_to_name)
		for sec in elf.iter_sections():
			if not isinstance(sec, SymbolTableSection):
				continue
			for sym in sec.iter_symbols():
				print(sym.name)
				if sym.entry.st_size != 0:
					section_name = index_to_name[sym.entry.st_shndx]
					cur_section = sections[section_name]
					print("{} {}".format(sym.name, sym.entry.st_info.type))
					if cur_section[1] == sym.entry.st_info.type:
						print(cur_section[0].append(sym.name))
	print(sections)` 


sections will contain a list of all symbols of each (named)section

eliben (approved):
Thanks

PR #335, opened 2020-10-08
DWARFv1 constants in enums, DW_FORM_ref parsing

sevaa:
[A customer asked for support for DWARF v1.](https://github.com/sevaa/dwex/issues/11)

Building support for DWARFv1 in `pyelftools` is probably not a fruitful endeavor, since many assumptions of DWARFv2+ break down in v1 (there is no abbrev table, no CU headers, expression and lineprogram format is different). [I've already slapped together a DWARFv1 parser](https://github.com/sevaa/dwex/blob/master/dwex/dwarfone.py) that mimics `pyelftools` and reuses some of its pieces.

That said, there are some `DW_TAG_`, `DW_AT_`, and `DW_FORM_` constants that were deprecated in DWARFv2 but can be seen in the v1 spec, and also in my reference DWARFv1 binary. The enums are otherwise identical (modulo the shift-by-4 thing for attribute codes). I'd rather have those constants right there in `pyelftools` than patch the dicts on the fly.

If there *is* interest in having pyelftools support DWARFv1, let me know.


elftools/dwarf/locationlists.py
```diff
@@ -106,7 +106,7 @@ def parse_from_attribute(self, attr, dwarf_version):
 
     @staticmethod
     def _attribute_has_loc_expr(attr, dwarf_version):
-        return ((dwarf_version < 4 and attr.form == 'DW_FORM_block1' and
+        return ((dwarf_version < 4 and attr.form.startswith('DW_FORM_block') and
```
eliben, line 109: Is this correct?

The existing code seems to be specifically targeting `DW_FORM_block1`, not the other kinds; why is this changing? What new forms are needed for DWARFv1?

eliben:
This PR seems mostly non-controversial, but general support for DWARF v1 is not planned.

elftools/dwarf/locationlists.py
sevaa: Saw a location expression of form`DW_FORM_block2` in the reference binary. There was no `block1` in DWARF1, and no locations sections either.

If this is a dealbreaker, we can leave this one out and I'll monkeypatch.

eliben (approved):

Issue #336, opened 2020-10-19
DWARFError while parsing mipsel libgcc.so


plafosse:
```
Python 3.7.7 (default, Mar 10 2020, 15:43:33)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from elftools.elf.elffile import ELFFile
   ...: elf_file = ELFFile(open('linux/mipsel/libgcc.so', 'rb'))
   ...: dwarf_info = elf_file.get_dwarf_info()
   ...: for cu in dwarf_info.iter_CUs():
   ...:     for die in cu.iter_DIEs():
   ...:         die.offset
   ...:
---------------------------------------------------------------------------
DWARFError                                Traceback (most recent call last)
<ipython-input-1-a8cda90efc13> in <module>
      3 dwarf_info = elf_file.get_dwarf_info()
      4 for cu in dwarf_info.iter_CUs():
----> 5     for die in cu.iter_DIEs():
      6         die.offset
      7

~/Library/Python/3.7/lib/python/site-packages/elftools/dwarf/compileunit.py in iter_DIEs(self)
    103             Note that null DIEs will also be returned.
    104         """
--> 105         return self._iter_DIE_subtree(self.get_top_DIE())
    106
    107     def iter_DIE_children(self, die):

~/Library/Python/3.7/lib/python/site-packages/elftools/dwarf/compileunit.py in get_top_DIE(self)
     92                 cu=self,
     93                 stream=self.dwarfinfo.debug_info_sec.stream,
---> 94                 offset=self.cu_die_offset)
     95
     96         self._dielist.insert(0, top)

~/Library/Python/3.7/lib/python/site-packages/elftools/dwarf/die.py in __init__(self, cu, stream, offset)
     92         self._parent = None
     93
---> 94         self._parse_DIE()
     95
     96     def is_null(self):

~/Library/Python/3.7/lib/python/site-packages/elftools/dwarf/die.py in _parse_DIE(self)
    171             return
    172
--> 173         abbrev_decl = self.cu.get_abbrev_table().get_abbrev(self.abbrev_code)
    174         self.tag = abbrev_decl['tag']
    175         self.has_children = abbrev_decl.has_children()

~/Library/Python/3.7/lib/python/site-packages/elftools/dwarf/compileunit.py in get_abbrev_table(self)
     76         if self._abbrev_table is None:
[...]

PR #339, opened 2020-10-26
DebugSectionDescriptor.size initialized with decompressed section size

sevaa:
Since `DebugSectionDescriptor.size` is being used to validate offsets (e. g. in the abbrev table), it should be initialized to decompressed size as opposed to raw size. Addresses #336.


eliben:
Thanks!

Issue #344, opened 2021-01-05
Introduce sections.iter_sections(self, type=None) method


evelikov:
Please consider add an optimised menthod iter_sections() which optionally takes the section type, similar to the existing iter_tags().

I have noticed that, when working with raw archives the overhead or looping over the elf header becomes quite obvious. In one particular [use-case](https://lists.archlinux.org/pipermail/arch-projects/2021-January/005411.html) switching from iter_tags() to iter_tags(type) has resulted in massive speed increase.

While I don't expect the results to be quite as significant, the consistent API in itself is worth it.

Thanks in advance

eliben:
While I haven't looked at the use-case and existing API in detail, feel free to send a PR. I'll review it in detail then

evelikov:
Thanks for the prompt reply eliben. I've already had a look through the code-base and it's a little over my python skills at the moment. So in the short-term it's unlikely that I'll submit PR - more than happy to test though :grinning:

evelikov:
@rupran you're a star - patch works just fine.

As anticipated there is no performance boost, even with the pathological example - but the shorter code is a win in itself.

rupran:
No worries, always happy to help!

PR #345, opened 2021-01-11
ELFFile: allow filtering of sections by type in iter_sections

rupran:
As stated in the corresponding issue, we can already filter the output of Dynamic.iter_tags() by the type of the tag we're looking for.

Let's adapt the iteration over the sections of the ELF file so that it only yields sections of a certain type if the optional type parameter is passed to iter_sections().

By doing this we can also simplify two call sites inside the ELFFile class.

Fixes: #344

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -135,11 +135,17 @@ class ELFFile(object):
             self._make_section_name_map()
         return self._section_name_map.get(section_name, None)
 
-    def iter_sections(self):
-        """ Yield all the sections in the file
+    def iter_sections(self, type=None):
+        """ Yield all the sections in the file. If the optional |type|
+            parameter is passed, this method will only yield sections of the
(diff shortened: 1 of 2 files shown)


elftools/elf/elffile.py
```diff
@@ -135,11 +135,13 @@ def get_section_index(self, section_name):
             self._make_section_name_map()
         return self._section_name_map.get(section_name, None)
 
-    def iter_sections(self):
+    def iter_sections(self, type=None):
```
eliben, line 138: Please expand the method comment to explain what the `type` param means

eliben (changes requested):
Thanks for the PR! Just one comment

elftools/elf/elffile.py
rupran: Yep, I missed that, thanks!

eliben (approved):

PR #347, opened 2021-01-13
fix wrong prpsinfo in 32bit coredump

Kyle-Kyle:
According to https://elixir.bootlin.com/linux/v5.10.7/source/include/linux/elfcore.h
ELF PRPSINFO is defined as 
~~~
struct elf_prpsinfo
{
	char	pr_state;	/* numeric process state */
	char	pr_sname;	/* char for pr_state */
	char	pr_zomb;	/* zombie */
	char	pr_nice;	/* nice val */
	unsigned long pr_flag;	/* flags */
	__kernel_uid_t	pr_uid;
	__kernel_gid_t	pr_gid;
	pid_t	pr_pid, pr_ppid, pr_pgrp, pr_sid;
	/* Lots missing */
	char	pr_fname[16];	/* filename of executable */
	char	pr_psargs[ELF_PRARGSZ];	/* initial part of arg list */
};
~~~
and `pid_t` is defined at
~~~
typedef __kernel_pid_t		pid_t;
~~~
`__kernel_pid_t` is defined at 
~~~
typedef int		__kernel_pid_t;
~~~

so `pr_pid, pr_ppid, pr_pgrp, pr_sid` should be all 4bytes

PR #348, opened 2021-01-13
Change pid_t elements in Elf_Prpsinfo to be 32 bit sized

rhelmot:
`pid_t` ends up typedefing down to `int` in both 32-bit and 64-bit environments. This fix makes 32-bit core files parse correctly.


rhelmot:
I am a fool. duplicate of #347

PR #351, opened 2021-01-25
Fix/extend aarch64 register names table

plafosse:
It seems as the the register number to name mapping was incomplete. This pull request adds the additional missing registers. The included file demonstrates the issue.
[libc.so.zip](https://github.com/eliben/pyelftools/files/5868031/libc.so.zip)


eliben (approved):
LGTM, thanks

plafosse:
No problem, thanks for the great project :)

PR #354, opened 2021-02-27
fixed parsing for structures containing uids or gids in core dumps for most architectures

Lukas-Dresel:
This fixes an issue where coredumps loaded from mips would be parsed incorrectly. On `mips`, the `gid` and `uid` field are 32bit long despite the architecture being 32bits. The list of architectures for 16-bit `uid/gid` fields is based on `libbfd` from binutils, which is what readelf, objdump, etc. use.

### DIFF
--- a/elftools/elf/structs.py
+++ b/elftools/elf/structs.py
@@ -373,6 +373,20 @@ class ELFStructs(object):
 
     def _create_note(self, e_type=None):
         # Structure of "PT_NOTE" section
+
+        self.Elf_ugid = self.Elf_half if self.elfclass == 32 and self.e_machine in {
+            'EM_MN10300',
+            'EM_ARM',
+            'EM_CRIS',
+            'EM_CYGNUS_FRV',
+            'EM_386',
+            'EM_M32R',
+            'EM_68K',
+            'EM_S390',
+            'EM_SH',
+            'EM_SPARC',
+        } else self.Elf_word
+
         self.Elf_Nhdr = Struct('Elf_Nhdr',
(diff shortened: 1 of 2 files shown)


eliben:
Thanks. Can you add a simple test?

Lukas-Dresel:
Yes, I'll add one sometime this week.

eliben:
Any luck with the test?

Lukas-Dresel:
Sorry, forgot about this, creating one now.

Lukas-Dresel:
I ended up finding a bug in QEMU's coredumping while trying to create this lol. Cleaning up the branch now, as some other stuff got mixed in.

Lukas-Dresel:
@eliben I added a testcase that illustrated the failure in the old behavior using a testcase from a simple crashing program: ```int main(int argc, char* argv[])
{
	int* p = (void*)0x1234;
	*p = 0x5678;
}
```

Lukas-Dresel:
Beware, current QEMU will drop an invalid core file, as their uid/gid dumping is also flawed. It writes out a 32-bit value but only byte-swaps the lower 16 bytes. I will be submitting a patch there soon as well. For now, the core file I provided is created with a patched version of QEMU.

eliben:
Is it ready for final review now?

Lukas-Dresel:
Yeah.

Lukas-Dresel:
Everything should be there, testcase is a correct corefile (at least as far as I can tell).

eliben (approved):
Thanks!

PR #361, opened 2021-05-13
[example] bug fixes in dwarf_decode_address example

jeffy1009:
I was building call trace analyzer based on the examples in pyelftools, and found out that line program entry with end_sequence=True is not handled, so it made my analyzer fail to get source code information for some cases.

I think it should be handled as in this PR.

Also, highpc should not be included in `decode_funcname`.
Because of it, multiple DIEs were matched for specific addresses.
I think it is right also according to DWARF v4 spec:
> DW_AT_high_pc  ...  is the relocated address of the first location past the last instruction associated with the entity

But I'm not 100% sure since I'm not the expert. I just spent a few days with pyelftools..

### DIFF
--- a/examples/dwarf_decode_address.py
+++ b/examples/dwarf_decode_address.py
@@ -67,7 +67,7 @@ def decode_funcname(dwarfinfo, address):
                               highpc_attr_class)
                         continue
 
-                    if lowpc <= address <= highpc:
+                    if lowpc <= address < highpc:
                         return DIE.attributes['DW_AT_name'].value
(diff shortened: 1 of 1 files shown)


eliben (commented):
Thanks for the contribution. Can you add a comment in the new location explaining the logic and why it has to be tested after the `if prevstate and prevstate.address...` condition?

jeffy1009:
@eliben Done :)

eliben (approved):

Issue #325, opened 2020-07-13
Add support for DWARF version 5


gpollo:
pyelftools still doesn't support DWARF version 5 as seen [here](https://github.com/eliben/pyelftools/blob/master/elftools/dwarf/dwarfinfo.py#L434). Is there anything considerable to be done except changing those lines and adding the new `DW_TAG` and `DW_AT`? If no work as been done, I might give it a try as I need some DWARF version 5 information.

sevaa:
Locations section parsing need to be rewritten.

pogo59:
There's actually a significant amount of change in DWARF v5.
New line table header format, which can move pathnames to a new section.
Location lists and range lists have a radically different encoding and use new section names.
References to .debug_str from .debug_info can be indirect through a .debug_str_offsets section.
Addresses can be in a separate .debug_addr section and referenced by index.
If you want to get super fancy there are several new unit types which can refer to debug info in a separate ".dwo" (dwarf object) file.

Those are the big items, off the top of my head.  There's a bunch of new FORMs to support the indirect references, and of course new DW_TAG and DW_AT items.

I know the DWARF side of things pretty well but I'm not such a great Python coder, and it's not clear my employer would give me time to work on this.  But I thought it was worth spelling out the scope a little more fully.

woodruffw:
I'm taking an initial stab at this, starting with the changes to CU headers.

pombredanne:
Since this has been merged, a new release would be awesome!

woodruffw:
One thing to keep in mind: the support I added is *very* preliminary. I’m positive additional work will be needed to fully support DWARF5!

pombredanne:
> One thing to keep in mind: the support I added is *very* preliminary. I’m positive additional work will be needed to fully support DWARF5! 

That's OK IMHO... at least we can start playing with these.

sevaa:
For the paranoid among us, one could make DWARF 5 support explicitly opt-in. Like, a flag that is False by default.

That said, I, as a consumer of pyelftools, would rather take my chances with the experimental code as it is.

noppej:
Hi, I would like to support @sevaa 's request for a release with experimental code. As a user of this, it is better to have a solution that **sometimes** works (experimental DWARF 5 support), than one that **never** works (no DWARF 5 support)

eliben:
Version 0.28 was released with all the latest stuff inside

sevaa:
How come this is still open? :)

PR #363, opened 2021-05-24
dwarf: initial DWARFv5 support

woodruffw:
~~This is _very_ WIP.~~ My plan is to get the initial stream and CU parsing in place, stub out support for the new line table format, and add some tests.

See #325.


woodruffw:
It looks like the latest stable version of `readelf` (2.36.1) in GNU binutils still doesn't have great support for DWARFv5, so direct comparisons against their output probably aren't going to be very fruitful.

However, as of 778fd8a on this branch, I can run `readelf.py --debug-dump=info WHATEVER` on a few trivial ELFs that use DWARFv5 on my system. I plan to work more on line program handling and other major parts that have changed, but I figured I'd make an initial stop here to see if it's best to merge this initial work first.

woodruffw:
Misc. notes:

* `DW_FORM_addrx3` and `DW_FORM_strx3` parsing require support for loading 24-bit unsigned integers. Newer versions of Construct support `UBInt24`, but the vendored one doesn't. Adding a custom `Construct` subclass for it probably wouldn't be too hard, but I have no idea how many DWARF emitters actually use these 24-bit forms. Maybe not worth it.
* I've done some plumbing to connect some of the new DWARFv5 sections up, so that the values of the various new forms (e.g. `DW_FORM_strx`) can be properly resolved. But those new sections aren't actually being parsed yet.

woodruffw:
I think this is ready for an initial review. I've added a basic DWARFv5 sample as a smoke test, just to confirm that we can populate a `DWARFInfo` at minimum and succeed at parsing v5-style CU headers.

eliben (approved):

eliben:
Thank you, this looks like a good start!

It's a shame `readelf` doesn't support it yet; I wonder if we can find some alternative for more exhaustive testing. The readelf-comparison tests give pyelftools very strong test coverage.

woodruffw:
> It's a shame `readelf` doesn't support it yet; I wonder if we can find some alternative for more exhaustive testing. The readelf-comparison tests give pyelftools very strong test coverage.

I've been doing my local testing against `llvm-dwarfdump`, which seems to perform admirably against DWARFv5 (i.e., it both parses correctly **and** resolves most of the new indirect references/indices).

Unfortunately, its output is pretty different (in terms of format) from `readelf`, so adding support for it (via another CLI?) would probably be nontrivial. I'll see if there's some way to get it to dump a JSON or YAML representation...

PR #372, opened 2021-09-17
Keep raw note descriptors in ELF note sections as raw bytes

mebeim:
Currently all note descriptors in notes sections (`SHT_NOTE`) are being treated as `latin-1` encoded strings by the library when iterated over through `iter_notes()`:

https://github.com/eliben/pyelftools/blob/9db67b19b237f0d75f119633b3f23f4b67a60b3d/elftools/elf/notes.py#L32-L33

This is not necessarily true, and the decoding should only be done if needed. I've removed the default decoding and moved it in the special case for `NT_GNU_BUILD_ID`.

---

Here's an example on my system on an ELF binary with a note type that is not fully parsed by this library, leaving it raw.

Before:

```none
Container({'n_namesz': 4, 'n_descsz': 16, 'n_type': 5, 'n_offset': 232, 'n_name': 'GNU', 'n_descdata': '\x02\x00\x00À\x04\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00', 'n_desc': '\x02\x00\x00À\x04\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00', 'n_size': 32})
```

After:

```none
Container({'n_namesz': 4, 'n_descsz': 16, 'n_type': 5, 'n_offset': 232, 'n_name': 'GNU', 'n_descdata': b'\x02\x00\x00\xc0\x04\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00', 'n_desc': b'\x02\x00\x00\xc0\x04\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00', 'n_size': 32})
```

---

I've noticed that my initial change (first commit) breaks the tests since the test suite is relying on the returned note descriptor always being a string. My suggestion would be to match known descriptor types (e.g. `NT_GNU_GOLD_VERSION` etc) and only decode the string if needed, which is what I did in the subsequent 3 commits.

### DIFF
--- a/elftools/common/py3compat.py
+++ b/elftools/common/py3compat.py
@@ -21,6 +21,7 @@ if PY3:
     # and strings are different types and bytes hold numeric values when
     # iterated over.
 
+    def bytes2hex(b): return b.hex()
     def bytes2str(b): return b.decode('latin-1')
     def str2bytes(s): return s.encode('latin-1')
     def int2byte(i): return bytes((i,))
@@ -41,6 +42,7 @@ else:
     import cStringIO
     StringIO = BytesIO = cStringIO.StringIO
 
+    def bytes2hex(b): return b.encode('hex')
     def bytes2str(b): return b
     def str2bytes(s): return s
     int2byte = chr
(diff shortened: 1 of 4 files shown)


eliben (approved):

eliben:
Thanks for the contribution!

PR #374, opened 2021-10-15
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 . `assertEqual` is present in both Python 2 and 3


tirkarthi:
Thanks @eliben .

PR #378, opened 2021-11-05
fix setup.py test

jlevon:
Attempting to run tests using setup.py fails with:

$ python3 setup.py test
======================================================================
ERROR: all_tests (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: all_tests
  File "/usr/lib/python3.9/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/home/jlevon/src/pyelftools/test/all_tests.py", line 12, in <module>
    from utils import is_in_rootdir
ModuleNotFoundError: No module named 'utils'

Fix by using a relative import where necessary.

Signed-off-by: John Levon <levon@movementarian.org>

### DIFF
--- a/test/all_tests.py
+++ b/test/all_tests.py
@@ -9,7 +9,7 @@
 #-------------------------------------------------------------------------------
 from __future__ import print_function
 import subprocess, sys
-from utils import is_in_rootdir
+from .utils import is_in_rootdir
 
 def run_test_script(path):
     cmd = [sys.executable, path]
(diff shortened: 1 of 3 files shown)


eliben (changes requested):
Thanks for the contribution.

Note that the tests are failing in CI now -- can you fix it?

jlevon:
Yep, thanks, apparently there's no simple way to import from the same dir in Python? Closing, as this isn't worth further effort.

eliben:
One could add the current directory to PYTHONPATH, I guess

jlevon:
Right, but seems too ugly to bother trying to fix.

PR #386, opened 2021-11-29
Add support for .note.gnu.property notes section

mebeim:
This adds support for the `.note.gnu.property` (`NT_GNU_PROPERTY_TYPE_0`) note section (documented [here](https://github.com/hjl-tools/linux-abi/wiki/linux-abi-draft.pdf)). It's already been implemented and supported for at least a couple of years by glibc, binutils (`gcc`, `readelf`, `ld`, `gold`, ...), LLVM, the Linux kernel ELF loader and others.

I plan to add support for the property types `GNU_PROPERTY_X86_ISA_1_{USED,NEEDED}` and `GNU_PROPERTY_X86_FEATURE_1_AND` when/if this gets merged. An example output below includes `GNU_PROPERTY_X86_FEATURE_1_AND` (`0xc0000002`) which is currently reported as `<unknown>`.

Everything seems to work fine, but there is a little bit of a problem that I came across while trying to add a simple test for this. I wanted to add an ELF binary with a `.note.gnu.property` section in `test/testfiles_for_readelf` (you can find it in my fork, [here](https://github.com/mebeim/pyelftools/blob/test-gnu-property/test/testfiles_for_readelf/note_gnu_property.elf), if you want to test), but:

1. Your `readelf` (v2.29) prints `LOOS+0x474e553` instead of `GNU_PROPERTY` for a program header of type `PT_GNU_PROPERTY` (mine is v2.35.2 and prints `GNU_PROPERTY`) even though it knows about gnu property notes (strange). Your `scripts/readelf.py` was actually broken for `PT_GNU_PROPERTY` program headers because it was missing from `_DESCR_P_TYPE` in `descriptions.py`. I added it in this PR, and now `scripts/readelf.py` prints `GNU_PROPERTY`.

   ```
   $ scripts/readelf.py -l my_note_gnu_property_linked_above.elf
   ...
     GNU_PROPERTY   0x0000000000000158 0x0000000000400158 0x0000000000400158
                    0x0000000000000030 0x0000000000000030  R      8
   ...

   $ test/external_tools/readelf -l my_note_gnu_property_linked_above.elf
   ...
     LOOS+0x474e553 0x0000000000000158 0x0000000000400158 0x0000000000400158
                    0x0000000000000030 0x0000000000000030  R      0x8
   ...
   ```

2. `readelf` (both yours v2.29 and mine v2.35.2) does not print any human-readable information along with `NT_GNU_PROPERTY_TYPE_0`, while my current implementation in this PR added `NT_GNU_PROPERTY_TYPE_0='program properties'` in the `_DESCR_NOTE_N_TYPE` dictionary and `scripts/readelf.py` prints that (otherwise it would have printed `(<unknown>)`), which would still make testing fail.

   ```
   $ scripts/readelf.py -n my_note_gnu_property_linked_above.elf
   
   Displaying notes found in: .note.gnu.property
   Owner                 Data size Description
   GNU                  0x00000020	NT_GNU_PROPERTY_TYPE_0 (program properties)
       Properties: stack size: 0x123000
         <unknown>: type 0xc0000002 (processor-specific) data: 03000000

   $ test/external_tools/readelf -n my_note_gnu_property_linked_above.elf

   Displaying notes found in: .note.gnu.property
   Owner                 Data size	Description
   GNU                  0x00000020	NT_GNU_PROPERTY_TYPE_0
       Properties: stack size: 0x123000
         x86 feature: IBT, SHSTK
   ```

I'm unsure about how to proceed for the above two points, any suggestions? Maybe just skip adding the test for now and update `test/external_tools/readelf` to a new version first? 

---

Implementation notes:

- `NT_GNU_PROPERTY_TYPE_0` is kind of a special notes section, in this case `n_desc` is an array with an unknown number of dynamically sized structures. The number of structures can only be known by parsing them one by one (i.e. until `n_descsz` is reached). There does not seem to be a macro (`macros.py`) capable of doing this automatically, as the `ctx` that is passed around does not know about the `Elf_Nhdr` and its relative `.n_descsz`. Therefore I opted for directly using [a loop in `iter_notes()`](https://github.com/eliben/pyelftools/pull/386/files#diff-2a58d8e87a15a91ede0a4bc4a1da30297ae7a571576552383b8995196d749fe7R50-R57).
- `GNU_PROPERTY_STACK_SIZE` should always have `pr_datasz` of `4` or `8` depending on ELF class (32 vs 64). I am matching the sizes [with a `Switch` macro](https://github.com/eliben/pyelftools/pull/386/files#diff-d07a98ebf046d52c49b17c34f64ed14758cd6245e4a0d61a65a03cb7c03bc3d1R378-R399) and I [emit a warning](https://github.com/eliben/pyelftools/pull/386/files#diff-a258af450c181130c62612e77d447d4929f9dbc2f2d369c42c23e1207e703d80R258-R265) in `scripts/readelf.py` for unexpected sizes, same as the actual `readelf` does.
- Same goes for `GNU_PROPERTY_NO_COPY_ON_PROTECTED` which should always have `pr_datasz == 0`.

References:

- Doc: https://github.com/hjl-tools/linux-abi/wiki/linux-abi-draft.pdf
- Linux: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=00e19ceec80b03a43f626f891fcc53e57919f1b3
- Glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86/dl-prop.h;h=385548fad3e4ad71dbdcdbfada58585c2f24ea5e;hb=HEAD
- Binutils: https://sourceware.org/git/?p=binutils-gdb.git&a=search&h=HEAD&st=commit&s=NT_GNU_PROPERTY_TYPE_0


eliben:
Thanks for the contribution.

I'm OK with a separate PR to update the readelf version to a more recent one (this may require some fixes and tweaks in the tests).

Is this sufficient to resolve the issue? I'd like your test to run, if possible. In some cases where readelf has bugs there are workarounds in the test runner and we can add more if needed.

mebeim:
Ok @eliben I think we could do it like this: 

1. Implement the test in this PR tweaking the python test script to convert `LOOS+X` in the `GNU_PROPERTY` when checking, and ignoring the missing human-readable description in `readelf`.
2. After this one is merged, try updating `readelf` to a newer version and update the tests accordingly (this might take sone work, I don't know how many things would change.
3. Remove the tweaks introduced in point 1 (though the tweak to ignore the missing human readable description would still be needed since I don't think latest `readelf` prints it.

What do you think?

eliben:
Sounds like a good plan, thanks!

mebeim:
Done! See the test ELFs here: https://github.com/eliben/pyelftools/pull/386/commits/577517fcf84cafedf28126e1e7842405b0c5c9cc. I've also added the source assembly file, it can be used to easily add new GNU property types and values to test in the future. Let me know what you think.

eliben (approved):
Thanks, LGTM!

I'm going to squash this because it's much easier to attribute changes to single commits later on, and much easier to roll back if need be. One PR = one commit

PR #388, opened 2021-12-12
Add support for note GNU_PROPERTY_X86_FEATURE_1_AND

mebeim:
Implement support for GNU property note type `GNU_PROPERTY_X86_FEATURE_1_AND` (which is a feature bitmask) and its relative flags. Also fix off-by-one in "Data size" column alignment for `readelf.py` note sections dump.

Reference DOC: https://gitlab.com/x86-psABIs/x86-64-ABI

Implementation notes:

1. The function name for describing `GNU_PROPERTY_X86_FEATURE_1_AND` is `describe_note_gnu_property_x86_feature_1()` and not `...feature_1_and()` because it can be used regardless if the features should be ORed or ANDed by the linker. `GNU_PROPERTY_X86_FEATURE_1_OR` could be defined in the future and this function would also be able to describe that without modification.
2. The `GNU_PROPERTY_X86_*` property types are always 4 bytes so I've special-cased them all to `('GNU_PROPERTY_X86_*', 4, 0)` in `elftools/elf/structs.py` to avoid filling up the `Switch` with unnecessary and redundant alternatives.
3. The enum `ENUM_GNU_PROPERTY_X86_FEATURE_1_FLAGS` is currently unused by other pyelftools code, it's only added for convenience for library users.
4. Using `_reverse_dict(ENUM_GNU_PROPERTY_X86_FEATURE_1_FLAGS)` for `_DESCR_NOTE_GNU_PROPERTY_X86_FEATURE_1_FLAGS` in `elftools/elf/descriptions.py` would have been just painful to manage in a decent way, I opted for a simple list of bitmasks and descriptions.
5. `readelf` silently ignores unknown bits set in `GNU_PROPERTY_X86_FEATURE_1_AND`, and so do I in `readelf.py`.


eliben (approved):
Thanks!

PR #389, opened 2022-01-02
Add PS3/CellOS OSABI identifier

iMoD1998:
Hello, this adds the OSABI ELFOSABI_CELLOS_LV2 for PS3.


eliben (commented):
I cannot find references to this enum value anywhere online - can you point to some reference documentation that describes it?

iMoD1998:
> I cannot find references to this enum value anywhere online - can you point to some reference documentation that describes it?

My apologies, its very undocumented but here is something.

https://www.psdevwiki.com/ps3/SELF_-_SPRX#OS_ABI_identification

I have binaries I can provide that also use this value if needed.

eliben:
Thanks. Any particular reason you chose a slightly different constant name - with `OS` in it; without it, the value is definitely more googleable :)

iMoD1998:
I was using a readelf from an SDK and thats what it output as the OSABI.

iMoD1998:
image

eliben:
If that's the reason, I suggest leaving the `OS` out of the constant name. You already have the readelf-matching name set in descriptions.py

iMoD1998:
I agree, just changed it.

iMoD1998:
Whoops missed a comma.

eliben (approved):

PR #390, opened 2022-01-05
Add support for bytes-like object as input stream

dungsaga:
when input is of type bytes or bytearray, convert it to a binary stream

### DIFF
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -70,7 +70,7 @@ class ELFFile(object):
     """
     def __init__(self, stream):
         if isinstance(stream, bytes) or isinstance(stream, bytearray):
-            stream = io.BytesIO(stream)
+            stream = BytesIO(stream)
         self.stream = stream
         self._identify_file()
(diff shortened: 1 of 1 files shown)


eliben (commented):
It's not clear to me why this is needed in pyelftools -- can you clarify?

Can't the caller wrap whatever they need in `BytesIO` if needed?

eliben:
If you can find additional evidence that this feature was requested in the past, I'll consider it

dungsaga:
I think it's easier to wrap it inside pyelftools than repeatedly wrap it every time I need to use pyelftools. I work with byte buffers more often than working with files.

dungsaga:
@eliben do you think that my reason is enough for this PR?

eliben:
Sorry, I don't think it adds sufficient value and I haven't seen any additional support in the ~month this issue has been open.

In the fullness of time it's possible to bloat any library with additions like these that make things a tiny bit more convenient, but at the end of the day there are harmful effects like increased complexity, additional axes on the testing matrix, and so on.

PR #392, opened 2022-01-13
Remove unused imports

brendanhaines:
Since these import from elftools.constuct rather than ..construct this import statement interferes with using pyelftools as a local (not pip installed) package


eliben:
@mebeim FYI

eliben (approved):
Thanks!

PR #395, opened 2022-02-11
Add support for DT_RELR/SHT_RELR compressed relocation sections

rupran:
As more and more tools now support DT_RELR compressed relocations (most notably, the just released GNU binutils 2.38 (0)), let's add support for reading these relocations as well.

The original discussion about advantages of packed RELATIVE relocations can be found at (1). In a nutshell, the format exploits the fact that RELATIVE relocations are often placed next to each other and (for x86_64) stores up to 64 relocations in two 8-byte words. In a regular .rela.dyn table, these would take up 24 * 64 = 1536 bytes.

The compressed relocations work as follows:

The first word in the section describes a base address and contains an offset for a relocation. This offset must always
lie at an even address. Following this entry can be one or more bitmap(s) which have their least significant bit set to 1.
All other bits describe (in increasing order of significance) if the following continuous offsets also contain a relocation. The
addends for existing relocations are stored at the corresponding offsets in the file (that is, they work like REL relocations).
A good description of the history of this feature and its current adoption is the following blog post (2).

(0): https://lists.gnu.org/archive/html/info-gnu/2022-02/msg00009.html
(1): https://groups.google.com/g/generic-abi/c/bX460iggiKg?pli=1
(2): https://maskray.me/blog/2021-10-31-relative-relocations-and-relr


elftools/elf/relocation.py
```diff
@@ -106,6 +107,76 @@ def __init__(self, header, name, elffile):
...
+        while relr < limit:
+            entry = struct_parse(self._relr_struct,
+                                 self.elffile.stream,
+                                 stream_pos=relr)
+            entry_offset = entry['r_offset']
+            if (entry_offset & 1) == 0:
+                # We found an anchor, take the current value as the base address
+                # for the following bitmaps and move the 'where' pointer to the
+                # beginning of the first bitmap.
+                base = entry_offset
```
eliben, line 141: It's not easily clear where this `base` is coming from, should it be "declared" above the loop with a comment?

eliben (commented):
Thanks, looks good. Just one minor question

elftools/elf/relocation.py
rupran: Yep, sounds good, I'll push an updated commit in a second.

eliben (approved):

Issue #398, opened 2022-02-19
dwarf v5 missing code in abbreviations table


wideglide:
I'm attempting to use pyelftools on a binary compiled with GCC 11 which uses dwarf version 5.  I get a key error in the abbreviation table lookup.  I'm not sure how to debug further.  It seems it is reading the abbrev stream just fine?  I'm using v0.28.

Here's an interactive session:
```python
Python 3.8.10 (default, Nov 26 2021, 20:14:08)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from elftools.elf.elffile import ELFFile
>>> fp = open("busybox-1_21_stable_gcc-11.2.0_mips_64_O1_busybox_unstripped.elf", "rb")
>>> elffile = ELFFile(fp)
>>> dwarf = elffile.get_dwarf_info()
>>> from ipdb import launch_ipdb_on_exception
>>> with launch_ipdb_on_exception():
...   lines = [dwarf.line_program_for_CU(CU) for CU in dwarf.iter_CUs()]
...
KeyError(68)
> /data/venv/lib/python3.8/site-packages/elftools/dwarf/abbrevtable.py(36)get_abbrev()
     35         """
---> 36         return self._abbrev_map[code]
     37

ipdb> self._abbrev_map.keys()
dict_keys([1, 2, 3, 4, 5, 7, 8, 9, 10, 11, 12, 13, 15, 16, 17, 18, 19, 20, 21, 23, 24, 28, 31, 32, 33, 34, 35, 36, 38, 42, 43, 45, 46, 47, 49, 51, 53, 54, 56, 58, 59, 61, 62, 64, 67, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 91, 92, 93, 94, 95, 96, 97])
ipdb>
```

sevaa:
68 is 0x44 is `DW_AT_namelist_item`.

sevaa:
Does `readelf` and/or `dwarfdump` have a problem with that file? Can you share the offending file?

wideglide:
`readelf` version 2.34 and `llvm-dwarfdump-12` process the file fine.

`llvm-dwarfdump-9` has issues:

```
error: decoding address ranges: invalid range list offset 0x1c556
error: dumping support for LLE of kind 1 not implemented
```

[...]

PR #400, opened 2022-04-22
Improve DWARF 5 compatibility.

rdunklau:
Hello,

I was running into the same kind of issues as #398 and figured where it came from.
A new special form has been added in DWARF5, DW_FORM_implicit_const, which change the format of the attribute specification: when it is present, a constant value is stored in the attribute spec itself.
If we fail to treat it specially, the constant value will be parsed as the beginning of the next attribute, which can lead the parser to merge several abbreviation declaration together, resulting in missing keys in the final table.

The second commit adds support for DW_FORM_line_strp, which uses a new section. 

I don't know the project's policy with regards to documentation and testing, so please let me know if I need to add something.


rupran:
Thanks for looking into this!

For fixing an issue like this (and I think, in general) it would be great to have a test binary added which exhibits the previously broken behaviour and write a short unit test to trigger the changed code.

I tend to add the minimal source code and compiler version/flags to the top of the Python test file in order to make this as replicable as possible (for example, see the [test_relr.py](https://github.com/eliben/pyelftools/blob/c2cfef0557a7297dc945fc29ff4f5d44ca6b1f51/test/test_relr.py) file).

You could probably add a new file into `test/testfiles_for_unittests` and extend the [test_dwarf_v5.py](https://github.com/eliben/pyelftools/blob/c2cfef0557a7297dc945fc29ff4f5d44ca6b1f51/test/test_dwarf_v5.py) file with the respective code.

rdunklau:
I added a new test file test_dwarf_v5_forms.py, let me know what you think.

samuelsadok:
I ran into the same issue and can confirm that this PR fixes the issue for the file I have at hand.

woodruffw:
Thanks for the ping. I'll take a look now.

elftools/dwarf/die.py
```diff
@@ -237,11 +237,17 @@ def _parse_DIE(self):
...
         # values from the stream.
-        for name, form in abbrev_decl.iter_attr_specs():
+        for spec in abbrev_decl['attr_spec']:
+            form = spec.form
+            name = spec.name
             attr_offset = self.stream.tell()
-            raw_value = struct_parse(structs.Dwarf_dw_form[form], self.stream)
-
-            value = self._translate_attr_value(form, raw_value)
+            # Special case here:
```
woodruffw, line 244: Comment here could use a little elaboration on why it's a special case.

sevaa: E. g.: 
```
# Special case here: the attribute value is stored in the attribute definition
# in the abbreviation spec, not in the DIE
```

sevaa:
Checked both features, `line_strp` and `implicit_const`, they both are implemented by the spec and in line with the general design philosophy of pyelftools. The unit test covers both.

elftools/dwarf/die.py
rdunklau: Thank you, I added this to the commit introducting support for implicit consts.

elftools/dwarf/die.py
```diff
@@ -237,11 +237,18 @@ def _parse_DIE(self):
 
         # Guided by the attributes listed in the abbreviation declaration, parse
         # values from the stream.
-        for name, form in abbrev_decl.iter_attr_specs():
+        for spec in abbrev_decl['attr_spec']:
```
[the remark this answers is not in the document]
rdunklau: Because we may need the third attribute (value) from the attr spec, when the form is implicit const.
I didn't like changing the method to return an almost useless value in most cases as a third member in the returned tuple, or changing it's behaviour completely to not return a tuple anymore.

But this is probably not ideal. If you think of a better way of handling this without changing the whole interface, let me know !

sevaa:
Release please, I have crash reports that trace back to `DW_FORM_implicit_const`.

eliben:
@sevaa honestly, I'm not too keen on releasing for every couple of landed commits. Releasing is a pain. Is there no way to tell your users to use the master-branch version?

sevaa:
I used to ship patched pyelftools, thought I won't have to anymore... :(

PR #401, opened 2022-04-23
Add DW_FORM_line_strp to structs.py

AlanRosenthal:
From https://dwarfstd.org/doc/DWARF5.pdf
`Chapter 7. Data Representation`: Page 218
```
In the 32-bit DWARF format, the
representation of a DW_FORM_strp,  or
DW_FORM_strp_sup value is a 4-byte unsigned offset; in the 64-bit
DWARF format, it is an 8-byte unsigned offset (see Section 7.4 on
page 196).
```

Testing Steps:
`test.c`
```
#include <stdint.h>

uint32_t fn1(void) {
    return 95;
}
```
`gcc -g3 -c test.c -o test.o`
`python examples/examine_dwarf_info.py --test test.o`
```
Processing file: elftools/common/test.o
  Found a compile unit at offset 0, length 164
    Top DIE with tag=DW_TAG_compile_unit
    name=7/0
```

### DIFF
--- a/elftools/common/py3compat.py
+++ b/elftools/common/py3compat.py
@@ -26,7 +26,10 @@ if PY3:
             return b.hex()
         return sep.join(map('{:02x}'.format, b))
 
-    def bytes2str(b): return b.decode('latin-1')
+    def bytes2str(b):
+        if isinstance(b, int):
+            return str(b)
+        return b.decode('latin-1')
     def str2bytes(s): return s.encode('latin-1')
     def int2byte(i): return bytes((i,))
     def byte2int(b): return b
--- a/elftools/dwarf/structs.py
+++ b/elftools/dwarf/structs.py
@@ -223,6 +223,7 @@ class DWARFStructs(object):
             DW_FORM_ref_addr=self.Dwarf_target_addr('') if self.dwarf_version == 2 else self.Dwarf_offset(''),
 
             DW_FORM_indirect=self.Dwarf_uleb128(''),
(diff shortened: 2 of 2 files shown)


rdunklau:
I implemented it here: https://github.com/eliben/pyelftools/pull/400/commits/b7a16b1060fc6365f087bf1974e3b44d0bbc7fc0
I'm not sure why you are parsing it like you do: in my understanding, it's exactly the same thing as a DW_FORM_strp, except that the offset refers to the .debug_line_str section instead of .debug_str.

vsoch:
heyo! I'm hitting this issue too - can we pick up discussion again to review / add this attribute?

sevaa:
Second that. The sample binary on #411 has them.

vsoch:
So I was just doing some testing - I think a fix might be added to the main branch here but not released yet?  E.g., the commit referenced above: https://github.com/eliben/pyelftools/commit/b7a16b1060fc6365f087bf1974e3b44d0bbc7fc0.

At least installing 0.28 from pip produced an error about this attribute, but cloning and installing from github resolved it. So perhaps we just need a release soon?

sevaa:
Wait, I have two PRs in the pipeline, and material for another :)

eliben:
Seems like this was already subsumed by @sevaa's PRs, is this correct?

@AlanRosenthal `DW_FORM_line_strp` exists in structs now -- does it have what you need?

eliben:
I'm going to close this PR for now; please reopen if the issue has not been resolved yet.

Issue #406, opened 2022-05-20
'missing terminator', FieldError('expected 1, found 0'


ytxloveyou:
when i analyze my elf file to call cu.iter_DIEs,  it reports 'missing terminator', FieldError('expected 1, found 0'
Any idea how to solve it?  the elf was compiled from windriver compiler (Diab).  I try several Diab elf files, then all of them will have such kind of issues. also sometimes , it detects an error abbrev code(only 0~34 supported), but 106 or 90 abbrevcode was extracted.

If the issues mentioned above is not easy to be fixed, could you add a feature like " an option to allow skip die if it is not valid"

Thanks a lot O(∩_∩)O

eliben:
Please upload a *minimal* ELF file that demonstrates this behavior, as well as the exact code / example / invocation you used to observe the error.

Do readelf and other tooling process your files successfully?

sevaa:
Well then, upload *that* ELF file. Is it big? Is it sensitive? Debugging by proxy is quite a chores.

sevaa:
Got the file. At least the first exception reproduces on the master branch.

sevaa:
The fix is in #407. At least pyelftools' `readelf.py --debug-dump=info` doesn't error out on the sample file.

Now, the fix is unlikely to be published to pypi right away. Until it is, you may integrate the master branch of pyelftools in your setup by hand, or monkey patch the function `iter_DIE_children` in class `CompileUnit` to the contents currently in the master branch.

sevaa:
There is another, subtler issue in that file. Sometimes, a DIE (a `subprogram`) would be marked as having children in the abbreviation table, but the only child it would contain would be a null DIE. In this case, the DIE object returns `has_children=True`, but iter_children would return an empty collection. However, checking for real children is next to impossible without parsing the children DIEs; even a one byte DIE can be a legit one. @eliben - is it something to address, or let the consumer beware?

ytxloveyou:
> The fix is in #408. At least pyelftools' `readelf.py --debug-dump=info` doesn't error out on the sample file.
> 
> Now, the fix is unlikely to be published to pypi right away. Until it is, you may integrate the master branch of pyelftools in your setup by hand, or monkey patch the function `iter_DIE_children` in class `CompileUnit` in `elftools/dwarf/compileunit.py` to the contents currently in the master branch.

Thanks a lot for your quick feedback.  I just try it in my iteration of DIE. It seems that all exceptions was solved.
[...]

PR #408, opened 2022-05-20
Support for sibling in form DW_FORM_ref_addr

sevaa:
Fix for #406


eliben (commented):
Thanks for the fix!

Any chance we could have a test for this somehow?

sevaa:
Maybe @ytxloveyou could build a sample binary. This seems specific to that one compiler, Wind River.

eliben (approved):

Issue #410, opened 2022-05-23
Support DWARF5 standardization of GNU extension expression opcodes


rdunklau:
In dwarfv5, a number of GNU extensions have been standardized.
I don't know how to test for them to be honest, but it would be nice to add synonyms to the GNU extension opcodes we already parse.

It seems to me that the following should be added:

DW_OP_GNU_entry_value => DW_OP_entry_value
DW_OP_GNU_const_type => DW_OP_const_type
DW_OP_GNU_regval_type => DW_OP_regval_type
DW_OP_GNU_deref_type => DW_OP_deref_type
DW_OP_GNU_implicit_pointer => DW_OP_implicit_pointer
DW_OP_GNU_convert => DW_OP_convert

rdunklau:
Thanks ! I did not provide a PR since it's trivial to add, and the problem is actually having a test case for it as you noted. I guess worst case scenario would be to manually replace opcodes in the binary with the GNU_ versions and run through it ?

sevaa:
EDIT: rebuilt the same code for Linux/x64 with DWARF5, got a test binary. At least 4 out of 6 opcodes are there. Stay tuned for a PR.

sevaa:
I'd like to. Barring more surprises from GNU, this shouldn't take long.

rdunklau:
I have a rudimentary implementation outside of pyelftools for this, let me make a proper patch for it.

sevaa:
I'm already halfway :)

rdunklau:
Great :-) Please ping me for the review if you want.

vegard:
Hi, I ran into issues trying to use the `dwarf_lineprogram_filenames.py` example on files from these archives:
* http://ftp.de.debian.org/debian/pool/main/l/linux/linux-image-5.17.0-3-amd64-dbg_5.17.11-1_amd64.deb
* http://ddebs.ubuntu.com/pool/main/l/linux/linux-image-unsigned-5.15.0-37-generic-dbgsym_5.15.0-37.39_amd64.ddeb
I'm pretty sure the errors are related to DWARF5 as I'm getting various errors (including what looks like opcode-related ones):

On `/usr/lib/debug/lib/modules/5.17.0-3-amd64/kernel/drivers/edac/edac_mce_amd.ko` from the 5.17.11 deb above:
```
  File "/home/ubuntu/.local/lib/python3.8/site-packages/elftools/dwarf/dwarfinfo.py", line 241, in line_program_for_CU
[...]

Issue #409, opened 2022-05-23
Error reading line programs from binary compiled with gcc 11.2


wheirman:
I'm getting the following error when reading line programs (`readelf.py --debug-dump=decodedline`) from a binary compiled with gcc 11.2.0 on Ubuntu 22.04:

`ELF error: ('missing terminator', ArrayError('missing terminator', FieldError('expected 1, found 0',)))`

The binary is attached here: [fft.zip](https://github.com/eliben/pyelftools/files/8752301/fft.zip)

I tried the code from #408 as well but that didn't help. `readelf.py --debug-dump=info` works fine on the same binary.

sevaa:
Confirmed, the issue exists, and it's unrelated to #408. The line program format has been revised in DWARF5, we knew this was coming, now might as well be the time.

sevaa:
@wheirman: mind if I make a test case out of your binary?

wim-heirman-at-intel:
I'm not sure what the licensing is on that fft example. But a simple hello world gives the same error, you can use this one:
[test-dwarf5-lineprogram.zip](https://github.com/eliben/pyelftools/files/8767982/test-dwarf5-lineprogram.zip)

This was compiled with just `-g` as an option.

```
$ gcc --version
gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

eliben:
Can you please verify that #411 fixes it (it landed on the master branch)?

Cirn09:
Work fine! 👍

eliben:
Closing this for now. Please reopen if you see it still doesn't work on the main branch

PR #411, opened 2022-05-25
Version 5 lineprogram header, take 1

sevaa:


sevaa:
No positive test yet. Waiting if the OP of #409 would let me use their binary as a test case.

sevaa:
@eliben, who might know about the call frame unwind data format? That's the last readelf discrepancy that I've got here.

eliben:
Do you mean `eh_frame`, or something else that's not specified in the DWARF standard itself?

sevaa:
`.eh_frame` indeed.

eliben:
Unfortunately that's the least documented part of toolchains, in my experience. Back in the day it was almost impossible to find documentation (beyond reading the source of binutils), but maybe things have improved in the last decade.

If it helps, #155 added it originally.

sevaa:
Now I have a strong hunch that the GNU `readelf` is the one in the wrong here. Let me see if I can get on *their* maillist.

sevaa:
@pmderodat - any chance you could please take a look at [my write-up of the potential GNU readelf bug](https://lists.gnu.org/archive/html/bug-binutils/2022-06/msg00009.html), and tell me if my analysis is off? Sorry for the formatting, it looked better in my outbox.

sevaa:
No feedback either way. :( Unless I hear something by June 7, 2022, 21:00 UTC, I'll go ahead and exclude the offending bit from the test.

test/run_readelf_tests.py
```diff
@@ -80,6 +80,15 @@ def run_test_on_file(filename, verbose=False, opt=None):
                 testlog.info('.......................SKIPPED')
             continue
 
+        # TODO(sevaa): I've checked the readelf sources, the DWARF docs,
```
eliben, line 83: Can we do it slightly differently, like excising the the `eh_frame` section from this sample file?

elftools/dwarf/enums.py
```diff
@@ -394,3 +394,24 @@
...
+ENUM_DW_UT = dict(
+    DW_UT_compile       = 0x01,
+    DW_UT_type          = 0x02,
+    DW_UT_partial       = 0x03,
+    DW_UT_skeleton      = 0x04,
+    DW_UT_split_compile = 0x05,
+    DW_UT_split_type    = 0x06,
+    DW_UT_lo_user       = 0x80,
+    DW_UT_hi_user       = 0xff
+)
```
eliben, line 417: Nit: add a newline here

elftools/dwarf/descriptions.py
```diff
@@ -312,12 +317,29 @@ def _describe_attr_block(attr, die, section_offset):
     DW_LANG_UPC: '(Unified Parallel C)',
     DW_LANG_D: '(D)',
     DW_LANG_Python: '(Python)',
+    DW_LANG_OpenCL: '(OpenCL)', # Not tested
```
eliben, line 320: What does "not tested" mean here?
AFAICT most of the entries above were not tested

elftools/dwarf/dwarfinfo.py
```diff
@@ -456,6 +456,34 @@ def _parse_line_program_at_offset(self, debug_line_offset, structs):
...
+                        replace_value(data, field.content_type, self.get_string_from_linetable)
+                    elif field.form == 'DW_FORM_strp':
+                        replace_value(data, field.content_type, self.get_string_from_table)
+                    elif field.form in ('DW_FORM_strp_sup', 'DW_FORM_strx', 'DW_FORM_strx1', 'DW_FORM_strx2', 'DW_FORM_strx3', 'DW_FORM_strx4'):
+                        raise NotImplementedError()
+
+        resolve_strings(self, lineprog_header, 'directory_entry_format', 'directories')
+        resolve_strings(self, lineprog_header, 'file_name_entry_format', 'file_names')
+
+        # DWARF5: provide compatible file/directory name arrays for legacy lineprogram consumers
```
eliben, line 478: Here and elsewhere, could you please clarify what "legacy" means?

elftools/dwarf/structs.py
```diff
@@ -282,27 +284,80 @@ def _create_lineprog_header(self):
...
-                lambda obj, ctx: len(obj.name) == 0,
-                self.Dwarf_lineprog_file_entry),
-            )
+            If(ver5,
+                PrefixedArray(
+                    Struct('directory_entry_format',
+                        Enum(self.Dwarf_uleb128('content_type'), **ENUM_DW_LNCT),
+                        Enum(self.Dwarf_uleb128('form'), **ENUM_DW_FORM)),
+                    self.Dwarf_uint8("directory_entry_format_count"))),
+            If(ver5, # Name deliberately doesn't match the legacy object, since the format can't be made compatible
```
eliben, line 337: Can the comment elaborate a bit more about what the legacy object is and the incompatibility?

test/run_readelf_tests.py
sevaa: I could try and patch the instruction area of the offending FDE with NOPs. Is there a control sum in ELF?

elftools/dwarf/descriptions.py
sevaa: I've added all the language codes that are listed in DWARF5. However, I didn't match the externally visible language designations (as seen in readelf data dumps) against the sources of GNU readelf. Since by now I've got the binutils' sources anyway, I will check that.

elftools/dwarf/dwarfinfo.py
sevaa: On this one, I could use some advice. In lineprogram header v4 and below, there are two arrays - directories, which is an array of strings, and files, which is an array of (name, directory_index, last_mod_time, size) tuples. These are stored in the lineprogram header object as two collections - `include_directory` and `file_entry`, respectively.

In v5, both arrays are present but they have a dynamic structure - they are each prefixed with an array of fields. So I could not reuse the old variable names, `include_directory` and `file_entry`, without either losing data or breaking item datatype. I've introduced two new data items in the header - `directories` and `file_names`. The names came straight from the DWARF5 spec. However, since the v5 data structures can be (but don't have to be) supersets of the v4 ones, I've provided code that would generate `include_directory` from `directories` and `file_entry` from `file_names` if the latter are present. This way, if an existing library consumer encounters a binary with lineprogram header v5, they will still see the familiar data structures.

This changes the externally visible API, but not in a breaking way. Let me know if I should have handled this differently. The fact that file names and directory names are data items (as opposed to, say, method calls) is a given.

elftools/dwarf/structs.py
sevaa: See above.

elftools/dwarf/descriptions.py
eliben: Thanks!

elftools/dwarf/dwarfinfo.py
eliben: I understand, thanks for clarifying. I think this all sounds fair and you should document it in a comment in the code.

test/run_readelf_tests.py
eliben: I was thinking just deleting the whole section like with `objcopy` (https://stackoverflow.com/questions/31453859/how-to-remove-a-specific-elf-section-without-stripping-other-symbols) or maybe there was another tool for this - I don't recall exactly.

This isn't a huge priority - fine to leave as is if this is hard. Just looking for a way to avoid more hacks in this test file if possible

eliben:
Pinging some folks in case they're interested in reviewing - this is a fairly major step in DWARF v5 support.

@woodruffw @rupran 

Just a friendly ping, no worries if this doesn't work out

elftools/dwarf/dwarfinfo.py
sevaa: Commented in the `DWARFInfo.line_program_for_CU()` heredoc.

elftools/dwarf/descriptions.py
sevaa: Matched against binutils, a couple of minor fixes.

test/run_readelf_tests.py
sevaa: Patched. One instruction that threw binutils off replaced with two zero bytes (NOPs). Made a note in the test script.

sevaa:
I wanted to push one more PR in the near future, one that the examples autotest runs on Windows (it currently fails, over `os.path.join()`), but that's all internal stuff, can be done before or after the release.

examples/dwarf_die_tree.py
```diff
@@ -9,12 +9,22 @@
...
+import posixpath
 
 # If pyelftools is not installed, the example can also run from the root or
 # examples/ dir of the source distribution.
 sys.path[0:0] = ['.', '..']
 
 from elftools.elf.elffile import ELFFile
+from elftools.common.py3compat import bytes2str
+
+# Replacement for DIE.get_full_path, with POSIX paths instead of host patsh
```
eliben, line 21: Nit: s/patsh/paths/

examples/dwarf_die_tree.py
```diff
@@ -9,12 +9,22 @@
...
+def get_full_path_posix(die):
+    comp_dir_attr = die.attributes.get('DW_AT_comp_dir', None)
+    comp_dir = bytes2str(comp_dir_attr.value) if comp_dir_attr else ''
+    fname_attr = die.attributes.get('DW_AT_name', None)
+    fname = bytes2str(fname_attr.value) if fname_attr else ''
+    return posixpath.join(comp_dir, fname)
```
eliben, line 27: It's puzzling that you have to mostly duplicate `get_full_path` here. Also, I don't see `posixpath` in Python 3

What problem is this trying to solve, and can we solve it in some other way?

eliben (commented):
Thanks for fixing the test file - much nicer not to have an ugly skip in the tester to worry about

examples/dwarf_die_tree.py
sevaa: The test of example scripts against the reference outputs would fail, if run on Windows, on every line where a directory name and a file name needed to be joined. The reference output was generated on a POSIX system and uses / as a separator; `get_full_path` uses `os.path.join`, which uses the separator from the host, which is \ on Windows.

One alternative I've considered was passing an optional, defaulted parameter to `get_full_path`, telling it which flavor of `path` to use (choices: host=default, auto, NT, POSIX) but this is just for the sake of the test, not because anyone complained, so I've decided to leave the API alone.

eliben: Fair point about the CI.

I didn't consider Windows compatibility for these, hmm... Still I feel like we'd need to find a different solution here. TBH I'm surprised the tool still works on Windows at all! I certainly haven't tested it on Windows in many years and the CI doesn't run on Windows either.

I feel like we'll need a different solution here anyway - I don't want the logic of get_full_path duplicated in many places. The `pathlib` module has an `as_posix` utility - could this be useful to convert the returned result of `get_full_path`?

sevaa: The returned result, on Windows, is more often than not a Frankenmonster - something like `/home/john/myproject\mysource.cpp`, technically not a proper path either way. `Pathlib.Path()` on Windows seems to take those, though.

I will change the two snippets where I'd rewrite `get_full_path()` to use this instead, and leave the lineprogram bit alone. How's that?

As for general Windows compatibility, the unit test works as expected, the example test works once the path issue is fixed. Readelf would a trickier endeavor, since running the built-in `readelf` executable is off the table (except maybe with WSL, which I didn't investigate). There are several builds of readelf that run under Windows (Cygwin, Android NDK), but I've tried testing against the former and ran into output discrepancies right away. Didn't bother to debug those.

sevaa:
OBTW, I have a pretty good idea what to do about #410. Should I:
 - Stuff it under this PR?
 - Wait until this is integrated, then start a new one?
 - Start a parallel branch?

eliben:
> OBTW, I have a pretty good idea what to do about #410. Should I:
> 
>     * Stuff it under this PR?
> 
>     * Wait until this is integrated, then start a new one?
> 
>     * Start a parallel branch?

Separate PR for sure.

GitHub is not great at having stacked PRs, so when to do it really depends on whether the two PRs are dependent on each other. If not, you can create a different branch a new PR from it. If yes, maybe better to wait until this one lands.

sevaa:
They might be dependent, since I'm already getting readelf ..info discrepancies on the test binary. I'll wait for this one.

examples/dwarf_die_tree.py
sevaa: Changed. `Pathlib` is not in Python2.

eliben (approved):

eliben:
Thanks, I'm merging this.

Is there anything else related to this that you'd like to get in before I release a new version to PyPI?

Issue #417, opened 2022-06-14
examples/dwarf_lineprogram_filenames.py: TypeError: tuple indices must be integers or slices, not str


vegard:
I cloned a fresh copy of pyelftools to check DWARF5 line program support and I get this exception:

```
pyelftools$ python3 examples/dwarf_lineprogram_filenames.py --test /tmp/tmp31cx7ko3/usr/lib/debug/lib/modules/5.17.0-3-amd64/kernel/drivers/atm/ambassador.ko
Processing file: /tmp/tmp31cx7ko3/usr/lib/debug/lib/modules/5.17.0-3-amd64/kernel/drivers/atm/ambassador.ko
  Found a compile unit at offset 0, length 147144
Traceback (most recent call last):
  File "examples/dwarf_lineprogram_filenames.py", line 96, in <module>
    process_file(filename)
  File "examples/dwarf_lineprogram_filenames.py", line 46, in process_file
    line_entry_mapping(line_program)
  File "examples/dwarf_lineprogram_filenames.py", line 62, in line_entry_mapping
    filename = lpe_filename(line_program, lpe.state.file)
  File "examples/dwarf_lineprogram_filenames.py", line 82, in lpe_filename
    dir_index = file_entry["dir_index"]
TypeError: tuple indices must be integers or slices, not str
```

This is the unmodified example program that comes with pyelftools.

sevaa:
The first one I sort of know what to do (I broke it in the first place :)). The second one will need a more careful look. I'm on another piece at the moment, but I'll get to this. Thanks for bringing it up.

vegard:
```
$ python3 --version
Python 3.8.10
```

vegard:
Nice, if I just change it like this, it works:
```
--- a/examples/dwarf_lineprogram_filenames.py
+++ b/examples/dwarf_lineprogram_filenames.py
@@ -79,7 +79,7 @@ def lpe_filename(line_program, file_index):
 
     # File and directory indices are 1-indexed.
     file_entry = file_entries[file_index - 1]
-    dir_index = file_entry["dir_index"]
+    dir_index = file_entry.dir_index
 
     # A dir_index of 0 indicates that no absolute directory was recorded during
     # compilation; return just the basename.
```
output:
```
$ python3 examples/dwarf_lineprogram_filenames.py --test /tmp/tmp31cx7ko3/usr/lib/debug/lib/modules/5.17.0-3-amd64/kernel/drivers/atm/ambassador.ko
[...]

PR #414, opened 2022-06-07
Tuple indices must be integers or slices, not str

yuzeming:
```
Traceback (most recent call last):
  File "/home/yzm/pyelftools/examples/dwarf_lineprogram_filenames.py", line 95, in <module>
    process_file(filename)
  File "/home/yzm/pyelftools/examples/dwarf_lineprogram_filenames.py", line 45, in process_file
    line_entry_mapping(line_program)
  File "/home/yzm/pyelftools/examples/dwarf_lineprogram_filenames.py", line 61, in line_entry_mapping
    filename = lpe_filename(line_program, lpe.state.file)
  File "/home/yzm/pyelftools/examples/dwarf_lineprogram_filenames.py", line 81, in lpe_filename
    dir_index = file_entry["dir_index"]
TypeError: tuple indices must be integers or slices, not str
```

PR #418, opened 2022-06-15
DWARF 5 operations and DWARF5 location lists

sevaa:
The big change here is support for the `.debug_loclists` section, which in DWARF 5 replaces `.debug_loc`.

A pre-v5 `debug_loc` section contained records of three types:
 - Location (contains a PC address range, relative to some base, and an expression)
 - Base address (resets the base PC address)
 - List terminator (not returned to callers)
 
Different entry types were distinguished by magic values in the leading fields.

In the API, we had two ways of looking at locations: `LocationLists.get_location_list_at_offset()` (the offset would come from a DIE attribute value), and `LocationLists.iter_location_lists()`. **Most use cases would use the former**; the latter is only relevant for readelf dump purposes. Both methods returned location lists as lists of `LocationEntry` and `BaseAddressEntry` named tuples.

---

In DWARFv5, they have introduced the `debug_loclists` section, which has a different structure. It's now an array of CU blocks. Each block contains a header with an optional array of offsets; that can be used for an alternative way to reference location lists, but my test binary doesn't contain the offsets. Inside the CU, there are location lists, but also something that is called location view pairs (internally a pair of ULEB numbers). Location view pairs work with the user attribute `DW_AT_GNU_locviews`; if there is one next to `DW_AT_location`, the location list is expected to be preceded by a set of location view pairs, one per each location-type entry. **I don't understand how do view pairs contribute to the debugging process**, but the data structure is clear enough to parse and dump.

GNU readelf dumps the view pairs, so we have to.

The entry encoding is different, too. Entries now have the type code as the first byte; the codes are in the `DW_LLE` enum. There are 4 categories of entries:
 - Location (PC address range and an expression - but there are several encoding options)
 - Base address (also several encoding options)
 - Default location (no PC range, but has an expression)
 - List terminator

In some location entry encodings, the PC ranges are given as relative to the base, in some they are absolute. There are also encodings where the PC range is given as an index into the `.debug_addr` section; those are not present in the test binary and are not supported; they throw a `NotImplementedError`.

**So how does all that new stuff map to the API surface?** Let me count the ways.

`LocationLists.get_location_list_at_offset()` still returns a list of `LocationEntry` and `BaseAddressEntry` objects. It got an optional `die` argument, which is not used now, but will be once the unsupported encodings are supported. The `LocationEntry` entry object gained an extra member `is_absolute` (boolean).

Default location type entries are returned as `LocationEntry` objects with magic value -1 both in start_offset and end_offset. The test binary doesn't contain those.

`LocationLists.iter_location_lists()` now returns lists of `LocationEntry`, `BaseAddressEntry`, and newly introduced `LocationViewPair` objects (where present). This corresponds closely to the structure of the readelf..loc data dump.

There is no API for getting the view pairs by offset.

---

Additional changes:
 - The DWARFv5 expression operations DW_OP_entry_value, DW_OP_const_type, DW_OP_deref_type, DW_OP_regval_type, DW_OP_implicit_pointer, DW_OP_convert are properly parsed and described
 - The generated `file_entry` objects are indexable by name, #414 and #417 
 - The ignoring of the View column in lineprogram dump result comparison is done more carefully
 - Readelf test runs on Windows... somewhat
 
---

On a side note, I've built the same codebase with clang++-11, hoping to get some of the new encodings, and the copy of readelf that ships with pyelftools would error out on that a lot. Maybe an upgrade of readelf would be in order at some point, were we to embrace DWARF5. That might throw off tests on existing files, though.

I can use the clang-11 binary to debug the logic on my end, but doing the usual test-against-readelf routine won't be possible for now.


eliben:
Upgrading the vendored `readelf` binary is something that can definitely be attempted separately. It's already reasonably up-to-date though (added in Dec 2021) and the changes weren't huge - see commit 662e4ab30d314d5d00e901bebc3187636d5adc59

elftools/construct/macros.py
```diff
@@ -632,3 +632,24 @@ def OnDemandPointer(offsetfunc, subcon, force_build = True):
 
 def Magic(data):
     return ConstAdapter(Field(None, len(data)), data)
+
+class StreamOffset(Construct):
```
eliben, line 636: `construct` is vendored, so we should not have changes in the `elftools/construct` dir. If we need this adapter, it should be defined outside (in pyelftools's own code)

elftools/dwarf/dwarfinfo.py
```diff
@@ -502,3 +515,17 @@ def replace_value(data, content_type, replacer):
             structs=structs,
             program_start_offset=self.debug_line_sec.stream.tell(),
             program_end_offset=end_offset)
+
+    @staticmethod
+    def _save_section(section, filename):
```
eliben, line 520: Does this actually need to be a staticmethod of this class or could it go into utils?

elftools/dwarf/locationlists.py
```diff
@@ -8,38 +8,127 @@
...
         BaseAddressEntry objects.
+
+        Starting with DWARF5, it may also contain LocationViewPair, but only
+        if scanning the section, never when requested for a DIE attribute.
+
+        The default location entries are returned as LocationEntry with 
+        begin_offset == end_offset == -1
     """
-    def __init__(self, stream, structs):
+    def __init__(self, stream, structs, version = 4, dwarfinfo = None):
```
eliben, line 29: Nit: for default param values no spaces around `=`

Can you document these params in the docstring of the class?

elftools/dwarf/locationlists.py
```diff
@@ -8,38 +8,127 @@
...
         self.stream = stream
         self.structs = structs
+        self.dwarfinfo = dwarfinfo
+        self.version = version
         self._max_addr = 2 ** (self.structs.address_size * 8) - 1
 
-    def get_location_list_at_offset(self, offset):
+    def get_location_list_at_offset(self, offset, die = None):
```
eliben, line 36: Same comment w.r.t. documentation and formatting

test/run_readelf_tests.py
```diff
@@ -142,13 +145,27 @@ def prepare_lines(s):
         return False, 'Number of lines different: %s vs %s' % (
                 len(lines1), len(lines2))
 
+    view_col_position = None
```
eliben, line 148: Can you add a comment saying what the type here is?

test/run_readelf_tests.py
```diff
@@ -142,13 +145,27 @@ def prepare_lines(s):
...
+            view_col_position = None    
+    
+        # Check if readelf..decodedline output contains the view column
+        if flag_in_debug_line_section and lines1[i].startswith('file name') and view_col_position is None:
+            view_col_position = lines1[i].find("view")
+            stmt_col_position = lines1[i].find("stmt")
+
+        # Excise the View column, if any
+        if view_col_position and not lines1[i].endswith(':'):
+            lines1[i] = lines1[i][:view_col_position] + lines1[i][stmt_col_position:]
```
eliben, line 168: What will happen if stmt_col_position wasn't defined? (the `if` before this one didn't fire)

eliben (changes requested):
Thanks!

Initial review. Could you also add the source for the new binary somewhere in testfiles_for_readelf with instructions on how to rebuild it? For example like test/testfiles_for_readelf/gcc48-simple.src.c

elftools/construct/macros.py
sevaa: I'll probably reuse this for ranges, so common/utils?

elftools/dwarf/dwarfinfo.py
sevaa: Could be utils.

sevaa:
> Could you also add the source for the new binary somewhere in testfiles_for_readelf with instructions on how to rebuild it?

It's 16 sources and a makefile. They are a part of a nonfree project, but not particularly sensitive. I could try and prune it to isolate the behaviors we are testing, but that'd be a bit of a chores.

test/run_readelf_tests.py
sevaa: Then the part of the line that corresponds to the View won't be cut from it, because that means there is no view column, of the file is not a decodedlines file to begin with :)

The if that sets view_col_position fires for the header line (, the next if fires for both header and every table line below, until the next CU, where view_col_position is reset back to None until the next table header.

test/run_readelf_tests.py
sevaa: Nullable integer :) It's a position within string.

eliben:
> > Could you also add the source for the new binary somewhere in testfiles_for_readelf with instructions on how to rebuild it?
> 
> It's 16 sources and a makefile. They are a part of a nonfree project, but not particularly sensitive. I could try and prune it to isolate the behaviors we are testing, but that'd be a bit of a chores. Still interested?

Ah, that's a shame. Maybe then you could add one or two small files that demonstrate a subset of the issues? It's really hard to reproduce stuff without a source sometimes.

sevaa:
Oh well, I guess I'll chase the behavior down. DWARF Explorer is your friend. The loclists section is pretty much a given with DWARFv5, but the new opcodes need some chasing. Does it have to be in this PR?

elftools/dwarf/locationlists.py
```diff
@@ -8,38 +8,136 @@
...
+        # attribute. A set of locview pairs (which is a couple of ULEB128 values) may preceed
+        # a location list; the former is referenced by the DW_AT_GNU_locviews attribute, the 
+        # latter - by DW_AT_location (in the same DIE). Binutils' readelf dumps those.
+        # There is a view pair for each location-type entry in the list.
+        #
+        # Also, the section may contain gaps.
+        #
+        # Taking a cue from binutils, we would have to scan this section while looking at
+        # what's in DIEs.
+        stream = self.stream
```
eliben, line 70: Is there any special reason for this assignment to `stream`? The original code just used `self.stream` directly?

elftools/dwarf/locationlists.py
```diff
@@ -56,21 +154,69 @@ def _parse_location_list_from_stream(self):
                 break
             elif begin_offset == self._max_addr:
                 # Base address selection entry
-                lst.append(BaseAddressEntry(entry_offset=entry_offset, base_address=end_offset))
+                entry_length = self.stream.tell() - entry_offset
+                lst.append(BaseAddressEntry(entry_offset=entry_offset, entry_length = entry_length, base_address=end_offset))
```
eliben, line 158: nit: in all params, no spaces around `=`

elftools/dwarf/locationlists.py
```diff
@@ -56,21 +154,69 @@ def _parse_location_list_from_stream(self):
...
                                 for i in range(expr_len)]
+                entry_length = self.stream.tell() - entry_offset
                 lst.append(LocationEntry(
                     entry_offset=entry_offset,
+                    entry_length = entry_length,
                     begin_offset=begin_offset,
                     end_offset=end_offset,
-                    loc_expr=loc_expr))
+                    loc_expr=loc_expr,
+                    is_absolute = False))
```
eliben, line 173: here too in this `LocationEntry`,

Also in general not sure the `entry_length` here adds much beyond just inlining the expression into the call (in the previous chunk too)

elftools/dwarf/locationlists.py
```diff
@@ -56,21 +154,69 @@ def _parse_location_list_from_stream(self):
...
+        return lst
+
+    # Also returns an array with BaseAddressEntry and LocationEntry
+    # Can't possibly support indexed values, since parsing those requires
+    # knowing the DIE context it came from
+    def _parse_location_list_from_stream_v5(self, die = None):
+        lst = struct_parse(self.structs.Dwarf_loclists_entries, self.stream)
+        # This won't contain the terminator entry
+
+        lst = [self._translate_entry_v5(entry, die) for entry in lst]
```
eliben, line 183: I'd just inline this whole thing into a single `return` with proper indentation

test/run_readelf_tests.py
```diff
@@ -142,13 +145,33 @@ def prepare_lines(s):
...
     for i in range(len(lines1)):
         if lines1[i].endswith('debug_line section:'):
             # .debug_line or .zdebug_line
             flag_in_debug_line_section = True
 
         # readelf spelling error for GNU property notes
         lines1[i] = lines1[i].replace('procesor-specific type', 'processor-specific type')
+        
+        # The view column position may change from CU to CU:
+        if view_col_position and lines1[i].startswith('cu:'):
```
eliben, line 160: Should this be `view_col_position is not None`?

Not a huge fan of "nullable integers" :-) because in boolean contexts 0 and None behave similarly

test/run_readelf_tests.py
eliben: It would be clearer if `stmt_col_position` would be defined with some default value outside the condition. You can add an assert for good measure if needed. Python's "define stuff inside blocks to be visible outside them" is not very readable.

eliben (commented):
Also as noted in the other comment, would a simple "hello world" C file compiled with the proper compiler version help tickle these new sections? This could be a good addition to the testing. I'm OK with you leaving it to a followup PR if you prefer.

elftools/dwarf/locationlists.py
sevaa: Just for readability.

elftools/dwarf/locationlists.py
sevaa: Readelf..loc dumps the offset of the list terminator entry, which our list parser doesn't return. Previously, we'd calculate that offset from the offset and the datatype of the last entry, but with various entry encodings in the picture, that logic becomes too bulky to maintain - I'd have to re-encode the entry to its original representation to recover the length. It's easier to capture the entry length explicitly.

test/run_readelf_tests.py
sevaa: 0 is impossible - short of drastic format changes in GNU readelf, the View column is never the first one. So it's realistically either None or a positive int. But I can fix, no big deal.

test/run_readelf_tests.py
sevaa: Okay, I'll ditch the reliance on None.

elftools/dwarf/locationlists.py
sevaa: Note the bit at readelf.py:1488 - that's where `entry_length` is needed.

eliben (approved):

rdunklau:
For the record, I tested the branch against the binary I'm working with: it works as expected and I no longer have to use specific code for dwarf_v5. 
Thanks !

eliben:
@rdunklau thanks for checking! That's great to hear; can you close your issue now?

rdunklau:
@eliben Done

PR #419, opened 2022-06-17
Support for DWARFv5 debug_rnglists section

sevaa:
The changes to the ranges section format are close in spirit to those that they did to loclists (see #418) , but not as drastic. There are no CU headers.

Instead of fixed format records with magic values, range list entries are now stored as prefixed with type (`DW_RLE`), with several possible encodings for each kind, so storing both entry address and entry length became necessary for readelf dumping.

I saw range lists overlapping - one `DW_AT_ranges` pointing in the middle of a range list that was referenced in some other `DW_AT_ranges`. GNU readelf handles that gracefully, so we had to. The preexisting logic of RangeLists.iter_range_lists() didn't accommodate for that, so I changed it to scan all DIEs for `DW_AT_ranges` values and scrolling through those.

The indirect encodings are not supported, just like with loclists. **I could try and debug that on my clang-11 binary, but no autotest until we upgrade readelf. Should I?**

FYI, the tag 2.38 is present in the binutils' sources...

### DIFF
--- a/elftools/dwarf/ranges.py
+++ b/elftools/dwarf/ranges.py
@@ -33,16 +33,15 @@ class RangeLists(object):
     """ A single range list is a Python list consisting of RangeEntry or
         BaseAddressEntry objects.
 
+        Since v0.29, two new parameters - version and dwarfinfo
+
         version is used to distinguish DWARFv5 rnglists section from
         the DWARF<=4 ranges section. Only the 4/5 distinction matters.
 
         The dwarfinfo is needed for enumeration, because enumeration
         requires scanning the DIEs, because ranges may overlap, even on DWARF<=4
     """
-    # Since dwarfinfo is not a required parameter, there is fallback to the
-    # broken (binutils incompatible) enumeration logic, in case there is an old consumer
-    # that creates RangeLists directly rather than using DWARFInfo.range_lists()
-    def __init__(self, stream, structs, version=4, dwarfinfo=None):
+    def __init__(self, stream, structs, version, dwarfinfo):
         self.stream = stream
         self.structs = structs
         self._max_addr = 2 ** (self.structs.address_size * 8) - 1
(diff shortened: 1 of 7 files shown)


eliben:
> The indirect encodings are not supported, just like with loclists. **I could try and debug that on my clang-11 binary, but no autotest until we upgrade readelf. Should I?**
> 
> FYI, the tag 2.38 is present in the binutils' sources...

Do you have the ability to build it (for Linux)? 

I can build it locally and try to run the readelf comparison tests and see what changed; typically there's a bunch of silly output format changes to sort through.

elftools/dwarf/ranges.py
```diff
@@ -12,18 +12,42 @@
...
         BaseAddressEntry objects.
+
+        version is used to distinguish DWARFv5 rnglists section from
+        the DWARF<=4 ranges section. Only the 4/5 distinction matters.
+
+        The dwarfinfo is needed for enumeration, because enumeration
+        requires scanning the DIEs, because ranges may overlap, even on DWARF<=4
     """
-    def __init__(self, stream, structs):
+    # Since dwarfinfo is not a required parameter, there is fallback to the
```
eliben, line 42: I'm OK with making it mandatory and simplifying the code (no separate fallback logic). Passing an extra dwarfinfo shouldn't be an issue if someone finds this troublesome.

eliben:
> You mean the binutils? Built the latest (from a couple of weeks ago). Then tried building the tag 2.38, got some errors. So mixed results. Maybe I can sort through the errors. Would you like to put this under this PR, or sort this out and start another? Were I to tackle that binary, it would trigger all kinds of changes that are unrelated to ranges - I saw some new form codes, etc.

Fine to have it as a separate PR

elftools/dwarf/ranges.py
sevaa: The natural way to get to RangeLists is via DWARFInfo.range_lists() anyway.

eliben (approved):

sevaa:
On to readelf upgrade. There seems to be one systemic change - no longer outputting directory name in decodedlines.

eliben:
Yes, I noticed there are some changes around the code you recently added when I tried to rerun tests with the newest readelf. On first look, it seemed like superficial differences like the one you've mentioned.

eliben:
If this helps, here's a readelf I built yesterday from the 2-38 branch of binutils: https://www.dropbox.com/s/ek0e3bo6chf03js/readelf?dl=0

```
$ ./readelf --version
GNU readelf (GNU Binutils) 2.38.20220617
Copyright (C) 2022 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or (at your option) any later version.
This program has absolutely no warranty.
```

It was built on a 64-bit Ubuntu and should work with the Github actions CI in this repo

sevaa:
Works on my Debian Buster. On the clang-11 binary, it gives some warnings over offsets being too large; I don't yet know which one here is bogus. The readelf autotest *mostly* works. I'll file a PR with nothing but this readelf and the changes to match, how's that?

eliben:
Fixing up our `readelf.py` and the tests each time binutils decides to make gratuitous changes is annoying, but we derive a *huge* benefit of having these readelf-comparison tests around. I firmly believe they're the number 1 reason for the stability and correctness of pyelftools and without them making deep changes would be practically impossible w/o breaking a whole bunch of stuff every time.

PR #420, opened 2022-06-17
Readelf upgraded to 2.38, format fixes

sevaa:


eliben (approved):
Well, that was easy :-)

sevaa:
On to the clang-11 binary.

In line with the project's philosophy, when facing a choice between minor changes to the API and slow, complicated but compatible code, I'll go for the former.

PR #423, opened 2022-06-28
Recognize DW_FORM_ref_udata as a reference type.

rdunklau:
References to other DIEs can also be implemented with a form
DW_FORM_ref_udata, for using the ULEB128 encoding.


eliben:
Closing this PR since it seems like #426 covers it. Please reopen if this isn't so

Issue #424, opened 2022-06-28
debug_ranges_sec and debug_rnglists_sec are not mutually exclusive.


rdunklau:
Hello,

I've just tested the range list implementation for dwarfv5, and I'm not sure it's the right way to approach it. In the same ELFFile, we can have compile units in DWARF version 5, and other in version 4. Each of those will reference their respective .debug_rnglists and .debug_ranges sections. I guess it can also happen with location lists ?

As for how to produce such a binary, I suspect that having different compile options for different compilation units which are then all linked together will produce that. I don't know how often it happens in real life though.

sevaa:
If you manage produce a binary like that, let me know. It'd be a fun test case. I won't be surprised, however, if readelf, our reference implementation of choice, won't handle that either.

sevaa:
A binary would be good.

sevaa:
I'll roll the fix for that into the loclists/rnglists redesign I'm currently on. It's a surprisingly cheap change.

rdunklau:
In terms of API, it would probably make sense to have the location_list / range_list API at the CompileUnit level, don't you think ?
And maybe keep the existing ones with an optional dwarf_version parameter if we want to be explicit while staying backwards-compatible.
For the binary, I'll open a PR with it.

eliben:
Can this be closed after #429 landed?

rdunklau:
I think we can, thanks !

PR #425, opened 2022-06-29
Add binary for testing dwarf_versions mix.

rdunklau:
As discussed in #424, here is a binary exhibiting a mix of dwarf version 4 and 5 compile units, and both .debug_ranges and .debug_rnglists as a result.


sevaa:
The PR won't go as is, but I've got the file. Thank you!

rdunklau:
Yep it's just that I didn't know how to send it in another way :-)

sevaa:
You can attach a file to a Github comment. Zip it up and drag the archive onto the comment edit box.

PR #426, opened 2022-06-30
Supplementary object files

rdunklau:
Since dwarf5, we have the concept of "supplementary object files", which are meant to optimize the size of DWARFInfo by letting several files refer to a common one for their common DIEs.

It works by adding a .debug_sup section, containing a header and a filename which will be referred to as the supplementary object file.
Some forms are designed to reference the supplementary object file instead of the current one:
 - DW_FORM_ref_sup4
 - DW_FORM_ref_sup8
 - DW_FORM_strp_sup

The references are either directly at the attribute level, or can be used in a DW_TAG_importedunit to import a whole compilation unit from the supplementary object file.

A similar thing also exists as GNU proprietary extension, using a .gnu_debugaltlink section with a slightly different format, and different form names:
 - DW_FORM_GNU_strp_alt
 - DW_FORM_GNU_ref_alt
 
In this implementation, I worked around the fact that we need to know the base directory by making it an optional argument to ELFFile. Ideally, an ELFFile would know it's location by itself.

Simple tests have been added. I will expand to them to make sure the DW_FORM_ref_sup4 and DW_FORM_ref_sup8 are properly supported. The test_.*2 files are not necessary by themselves, they are needed to generate the common files with dwz though.

This PR embeds the commit which is proposed separately in #423: one of dwz optimizations is to switch to ULEB128 encoding when it can.


rdunklau:
It seems that at least Fedora and probably other RPM-based distros incorporate dwz optimizations as part of their standard build toolchain.

sevaa:
From my vantage point, the notion that the supplementary file is an ELF one is a tricky proposition. Up until now, the DWARF parsing portion of the library was ELF agnostic, and I'd use it happily for Mach-O and PE, by constructing a `DWARFInfo` directly on top of the sections.

Some kind of easily replaceable hook for resolving the supplementary file name to a DWARFInfo would be nice.

rdunklau:
I understand your concern, that is why the dwarfinfo does not take an ELF file but a supplementary DWARFInfo directly.

For me the most concerning thing is that it's based around relative file names, and since the ELFFile API has no concept of files, it is a bit of a tricky proposition. 

So maybe I could implement it in the following way instead:
 - leave ELFFile basically untouched, only providing the structs definition for parsing the two sections but not adding any kind of relative file names awareness
 - have a resolve_sup hook in the dwarf info, which can be installed
 - lastly, provide a utility that takes an elf_file, a base_directory, and a dwarf_info to install the hook.
 
 Another question is how do we want to manage those references ? Should they always be resolved when we have a supplementary hook, hiding the initial value ? Especially in the case of imports.
 
If you don't think it's fitting for the library, I'll be fine with maintaining the infrastructure in my project, but I feel others may need this capability since it's in the standard, and used in real-world binaries.

sevaa:
Parametrizing by a lambda is the Pythonic way :)

Normally I'd go with resolving things automagically. While there might be someone who uses pyelftools as a reference implementation to match DWARF parsing against, I suspect most users are still building debugger adjacent applications with it, and would rather know what's going on in the debuggee process, than what exactly is in the DWARF per se.

eliben:
I'll actually be happy for @sevaa to take the lead on this review because I have absolutely 0 experience using pyelftools in non-ELF settings, so I don't know what the constraints there are.

sevaa:
> what would be the benefit of having a hook in dwarfinfo instead of an additional argument of DWARFInfo type ?

So that the consumer code won't have to parse the DWARF twice - once to retrieve the supplementary file name, once again to parse the same DWARF *with* a supplementary file provided.

> Is there a use case when we don't know about the supplementary one when instantiating the DWARFInfo ?

Every use case is like that. How would they know?

rdunklau:
> So that the consumer code won't have to parse the DWARF twice - once to retrieve the supplementary file name, once again to parse the same DWARF with a supplementary file provided.

> I see you've placed the debug_sup parsing into the ELF parsing part, but it belongs to the DWARF parsing, I think. Again, thinking about the non-ELF use case.

I think I can see the misunderstanding now. I've put it in the ELF parsing part, as the filename is relative to the file. Even though ELFFile doesn't really have a concept of a file, working with streams, the DWARFInfo part even less so as it works with sections directly. We don't have to parse anything twice.

> Also, did you consider the situation where there is a reference to a supplementary file but one is not available?
As of now it would raise a FileNotFoundError, but maybe we want something different.

As a side note, I have to admit I do not really understand why every section is passed separately to DWARFInfo instead of passing it an ELFFile directly, and why some things belong to ELFFile or DWARFInfo.  As a newcomer, this makes things a bit hard to understand at times. For example, .eh_frame is defined independently of ELF, yet it resides in DWARFInfo. So if one were to extend pyelftools to support .eh_frame_hdr sections, where should that parsing go ?

Since I'm just discovering the project, there is something I don't know as I don't understand the rationale for not having access to the full ELFFile in DWARFInfo itself and reading the whole dwarf sections in memory instead. I gathered from the history that it used to be the case. So please excuse me if what I'm asking seems obvious to you :-)

sevaa:
People can and do parse DWARF using pyelftools in files that are not ELF. To do that, they construct a `DWARFInfo`, passing the memory blocks for sections that are retrieved from files in other executable formats by other, external to pyelftools, means. `ELFFile.get_dwarf_info()` is not the only way to get a `DWARFInfo`.

`.debug_sup` is a part of the DWARF spec, not of ELF proper. Ergo, its parsing belongs to `DWARFInfo` and should be equally available to non-ELF file consumers.

rdunklau:
With what I'm proposing, yes that's the case.

> I don't see a business case for supplying a SupplementaryLoader other than a DWARFInfo, maybe that's a bit overly abstract, but I'm not going to pick on that.

I can get rid of the abstraction, if we think a regular DWARFInfo is all we are ever going to be needing. You probably know better than me what kind of usecase exists for DWARFInfo completely separated from actual ELFFile. 

> The supplementary filename is always a pure filename, no leading relative path component, right?
In the binary I ran into, the supplementary filename is a relative one... And this is why the base directory information was added to ELFFile, as it doesn't really works with files but rather with streams. 

Another approach which maybe would be cleaner, would be to not add anything to ELFFile itself for setting the supplementary dwarfinfo file (just parsing it), and not make it aware about anything related to files. That would leave it to the client code to enrich the DWARFInfo from what they found in the ELFFile.
Then, provide a helper taking a proper filename (or Path, if we're going to drop support for old python versions as was hinted in the release thread) which would: 
  - construct an ELFFile
  - use the supplementary object file section parsing facility from the ELFFile to build the actual path to the file
  - obtain a DWARFInfo from the ELFFile
  - set the supplementary_loader (or supplementary_dwarfinfo, whichever we agree on) on the dwarfinfo 
  - and finally returns it

That would keep ELFFile independent from an actual file or it's location, while still providing convenience for the user interested in resolving debug symbols.

sevaa:
I see some value in keeping `ELFFile` path unaware. One way to do that while not making ELFFile consumers write several extra lines for supplementary file loading would be - provide a function parameter to the ELFFile constructor that takes a relative path and returns a stream. This way, the ELFFile.get_dwarf_info() can open and read the supplementary file while not knowing anything about its location.

Also, I think `SupplementaryLoader` as a separate entity should go away :) And if we do that, the `supplementary_loader` property of DWARFInfo should be renamed to something else, e. g. `supplementary_dwarfinfo`.

Making life super convenient for non-`ELFFile` consumers is not a major priority; constructing a DWARFInfo from a non-ELF file is already a bit of a chores. So I'd be okay with making them check for a supplementary file manually, then obtaining and supplying a `supplementary_dwarfinfo` by hand, the way it is in the current implementation.

rdunklau:
> I see some value in keeping ELFFile path unaware. One way to do that while not making ELFFile consumers write several extra lines for supplementary file loading would be - provide a function parameter to the ELFFile constructor that takes a relative path and returns a stream. This way, the ELFFile.get_dwarf_info() can open and read the supplementary file while not knowing anything about its location.

I went a bit further with that, and provided a classmethod to load an elffile from a path, which sets up the loader as you suggested. To make this more convenient to use, I also added a close method to close the underlying stream, and implemented the context manager methods. This makes it possible to use an ELFFile like this:

```
with ELFFile.load_from_path(some_path) as elffile:
   ....
```

I feel like it makes for a nicer API.

> Also, I think SupplementaryLoader as a separate entity should go away :) And if we do that, the supplementary_loader property of DWARFInfo should be renamed to something else, e. g. supplementary_dwarfinfo.

Done. 

> Making life super convenient for non-ELFFile consumers is not a major priority; constructing a DWARFInfo from a non-ELF file is already a bit of a chores. So I'd be okay with making them check for a supplementary file manually, then obtaining and supplying a supplementary_dwarfinfo by hand, the way it is in the current implementation.

Thanks for the insight. This should not change much in that case then.

sevaa:
I like the current iteration of the API. @eliben - any comments?

elftools/elf/elffile.py
```diff
@@ -69,7 +70,7 @@ class ELFFile(object):
             e_ident_raw:
                 the raw e_ident field of the header
     """
-    def __init__(self, stream):
+    def __init__(self, stream, stream_loader=None):
```
eliben, line 73: The docstring has to explain this new parameter in detail (from the POV of its usage)

elftools/elf/elffile.py
```diff
@@ -88,6 +89,19 @@ def __init__(self, stream):
         self._section_header_stringtable = \
             self._get_section_header_stringtable()
         self._section_name_map = None
+        self.stream_loader = stream_loader
+
+    @classmethod
+    def load_from_path(cls, path):
```
eliben, line 95: Docstring with documentation

elftools/elf/elffile.py
```diff
@@ -203,7 +217,7 @@ def has_dwarf_info(self):
             self.get_section_by_name('.zdebug_info') or
             self.get_section_by_name('.eh_frame'))
 
-    def get_dwarf_info(self, relocate_dwarf_sections=True):
+    def get_dwarf_info(self, relocate_dwarf_sections=True, follow_links=True):
```
eliben, line 220: Document this new param too

eliben:
> I like the current iteration of the API. @eliben - any comments?

Thanks for the review process! Overall the API SGTM - I left an initial round of comments mainly asking to document the new functionality and parameters - this will help be better understand how things fit together. I'll do another round of review when that's addressed.

@rdunklau please ping when the comments are addressed (the best way to do so is using the "Re-request review" button next to my name on the top right of the PR -- looks a bit like a recycling logo)

elftools/elf/elffile.py
```diff
@@ -46,6 +47,12 @@ class ELFFile(object):
     """ Creation: the constructor accepts a stream (file-like object) with the
         contents of an ELF file.
 
+        Optionally, a stream_loader function can be passed as the second
```
eliben, line 50: Be a bit more prescriptive in describing the function: it takes a XXX and returns a XXX 

Also nit: s/pathes/paths/

elftools/elf/elffile.py
```diff
@@ -203,12 +227,15 @@ def has_dwarf_info(self):
...
 
-    def get_dwarf_info(self, relocate_dwarf_sections=True):
+    def get_dwarf_info(self, relocate_dwarf_sections=True, follow_links=True):
         """ Return a DWARFInfo object representing the debugging information in
             this file.
 
             If relocate_dwarf_sections is True, relocations for DWARF sections
             are looked up and applied.
+
+            If follow_links is True, we will try to load the supplementary
```
eliben, line 237: Why was the name `follow_links` chosen? Why not something that conveys that this is specifically about supplementary DWARF files?

eliben:
Thanks @rdunklau 

@sevaa my remaining review comments are minor. If LGTY, I will merge this after finishing up the review.

sevaa:
No objections from me. Support for non-`ELFFile` consumers was my one concern, that's addressed.

eliben:
@rdunklau do you have the time to finish the review for this soon (by tomorrow morning PST)? If not - no problem - I'll start working on 0.29 tomorrow and this can get merged when ready and will have time to settle for 0.30

elftools/elf/elffile.py
rdunklau: follow_links is used by readelf. I think it is more neutral, as this could be used for other types of external debug files. I haven't looked much into the spec for this, but I think .dwo files could be another case of external files to load.

rdunklau:
@eliben I addressed your comments.

eliben (approved):

sevaa:
Does this mean #423 can go?

elftools/dwarf/dwarfinfo.py
```diff
@@ -507,6 +517,11 @@ def replace_value(data, content_type, replacer):
                         replace_value(data, field.content_type, self.get_string_from_linetable)
                     elif field.form == 'DW_FORM_strp':
                         replace_value(data, field.content_type, self.get_string_from_table)
+                    elif field.form in ('DW_FORM_strp_sup', 'DW_FORM_GNU_strp_alt'):
+                        if self.supplementary_dwarfinfo:
+                            replace_value(data, field.content_type, self.supplementary_dwarfinfo.get_string_fromtable)
```
pmhahn, line 522: While working on #514 I found this bug: `supplementary_dwarfinfo.get_string_from{ → _}table`
```console
$ git grep -n 'get_string_from_\?table'
elftools/dwarf/die.py:322:            return self.dwarfinfo.get_string_from_table(raw_value)
elftools/dwarf/die.py:326:            return self.dwarfinfo.supplementary_dwarfinfo.get_string_from_table(raw_value)
elftools/dwarf/die.py:338:            return self.dwarfinfo.get_string_from_table(str_offset)
elftools/dwarf/dwarfinfo.py:328:    def get_string_from_table(self, offset: int) -> bytes | None:
elftools/dwarf/dwarfinfo.py:703:                        replace_value(data, field.content_type, self.get_string_from_table)
elftools/dwarf/dwarfinfo.py:706:                            replace_value(data, field.content_type, self.supplementary_dwarfinfo.get_string_fromtable)  # FIXME
```

eliben: NIce! Feel free to send fixes for this and other bugs discovered in the process

pmhahn: Fix in #602

PR #429, opened 2022-07-07
Mixing v4 and v5 loclists and rangelists sections

sevaa:
Addresses #424. OBTW, the binary contains two rangelists sections, but only a V4 loclists one.

As usual with a new binary, the readelf was thrown off in unrelated places. Fixes of note:

The aranges section has empty blocks in it - ones with only one, null/terminating entry. The previous implementation didn't return and dump those, readelf dumps them. I've introduced a compatibility flag.

Some more GNU notes had to be introduced.


elftools/dwarf/aranges.py
```diff
@@ -56,7 +56,7 @@ def cu_offset_at_addr(self, addr):
 
 
     #------ PRIVATE ------#
-    def _get_entries(self):
+    def _get_entries(self, need_empty=False):
```
eliben, line 59: Please document in the docstring what `need_empty` means

scripts/readelf.py
```diff
@@ -1296,8 +1312,11 @@ def _dump_debug_aranges(self):
         aranges_table = self._dwarfinfo.get_aranges()
         if aranges_table == None:
             return
-        # seems redundent, but we need to get the unsorted set of entries to match system readelf
-        unordered_entries = aranges_table._get_entries()
+        # Seems redundent, but we need to get the unsorted set of entries
```
eliben, line 1315: Nit: redundant

eliben (commented):
Just a couple of minor comments, LGTM otherwise

sevaa:
Fixed.

rdunklau:
Thanks for this PR. I don't really see how the LocationListsPair can be used. Is the client code supposed to test the return value with isinstance anytime it calls location_lists on the DWARFInfo ?

IMO, a more useful API would sit at the CU level, to fetch the locationlists corresponding to the CU dwarf version.

rdunklau:
Ok that makes sense. Thanks !

sevaa:
Okay. Next, support for indirect forms and indirect loclist/rangelist encodings, with that one binary as the test case.

Issue #421, opened 2022-06-17
Release v0.29


eliben:
I'm planning to release pyelftools v0.29 by tagging it and sending a version to PyPI.

sevaa:
Can we stuff the elements of that clang-11 binary under the release? So far, I see a handful of new forms and operators. Shouldn't take long.

eliben:
> Can we stuff the elements of that clang-11 binary under the release? So far, I see a handful of new forms and operators. Shouldn't take long.

Sure, there's no rush.

> 
> I then wanted to do a housekeeping PR about the sample binaries - reduce the clutter. This will make the CI run faster, for one thing. That can be done after the release.

Sounds like a good plan. I will defer the Python 2 purge to after 0.29 as well

sevaa:
A bit of trouble in paradise. Readelf 2.38 dumps quite a few DIEs of my clang-11 binary, then bails with an error code and the following message:

> readelf: Warning: DIE at offset 0xc875 refers to abbreviation number 2915 which does not exist

I can tell that the parsing process went off the rails long before that; for one thing, the DIE nesting level in the output goes into negatives after a while. But that means the autotest is broken, too.

Can you build, instead of 2.38, the *latest* binutils? Maybe they've fixed that. If not, I have a working environment for debugging binutils, I could try and patch whatever's going on. Feels like cheating, but it won't be the first time.

eliben:
Sure, here's a readelf built from the master branch: https://www.dropbox.com/s/69lm3e7zgg7d7u1/readelf-master?dl=0

If it doesn't help, maybe binutils simply hasn't caught up with the newest DWARF v5 yet... :shrug:

sevaa:
Still a bunch of "offset is too great" warnings, but no error exit. Guess we'll use this one.

eliben:
What's special about that binary? Could it be that the compiler is generating something fishy?

If this is clang-generated, it may be interesting to dump it using `llvm-dwarfdump` to see if it also gets confused.

sevaa:
Can't tell yet if the binary is fishy or GNU readelf is off. For all we kno, could be the latter.

sevaa:
Update: readelf is definitely to blame. I'll try submitting patches this time.

eliben:
Back from vacation.

Is there anything else blocking this release? The new support does not have to be final...

I'd like to release 0.29 and start removing Python 2 support, since this will make working with the code easier for everyone.

[...]

PR #430, opened 2022-07-11
Indirect encoding support

sevaa:
Adds support for strx, addrx, loclistx, rnglistx forms, also for RLE_...x entries in rangelists, LLE_...x entries in loclists.

One kink of note: generally, resolving indirect-form attributes is possible at parse time, but with one exception: sometimes one would get an indirect-form attribute in the very top DIE of a CU where the needed base_addr attribute is, and before the respective base_addr. This requires either some kind of look-ahead logic, or a second pass. I went with the latter, and only for the top DIE.


elftools/dwarf/ranges.py
```diff
@@ -16,19 +16,22 @@
 
 RangeEntry = namedtuple('RangeEntry', 'entry_offset entry_length begin_offset end_offset is_absolute')
 BaseAddressEntry = namedtuple('BaseAddressEntry', 'entry_offset base_address')
+# If we ever see a list with a base entry at the end, there will be an error that entry_length is not a field.
 
-def not_implemented(e):
-    raise NotImplementedError("Range list entry %s is not supported yet" % (e.entry_type,))
+def _translate_startx_length(e, cu):
+    # Avoid retrieving the start offset twice - I don't think Python can do inline assignments
```
eliben, line 22: This comment can be removed, IMHO. It's fairly clear what's going on

eliben (commented):
Nice!

One question: does the double parsing always happen now for the top CU or just when these attrs are present?

sevaa:
The second round of parsing (`_translate_indirect_attributes()`) is always invoked for the top DIE, but *only* for the top DIE. It bypasses all attributes with forms that are not indirect and translates all ones that are. See compileunit.py:100 and die.py.320.

With some effort, I can make the second pass conditional upon the presence of attributes that can't be translated on the first pass, but this will incur a penalty for non-top DIEs. The challenge is that translating the indirect forms involves, in most straightforward implementation, calling `get_top_DIE()` to get the base address, and if the DIE being parsed *is* the top DIE, then you get an infinite recursion. Adding a check there will marginally slow things down for every other DIE.

elftools/dwarf/ranges.py
sevaa: Removed.

sevaa:
Fixed.

eliben (approved):

sevaa:
Okay. Now, I'll be AFK for a couple of weeks, but then, I have plans to revise my test codebase to reduce the size and build complexity while keeping the coverage. The thing with nontrivial DWARF operations is that they only come up when the code is both convoluted and optimized; you won't get them with "Hello world" type programs.

eliben:
SGTM, I'll be AFK for a bit too. Enjoy your vacation and thanks for all the contributions recently!

When we've both back, it may be time to consider releasing the next version (#421)

PR #434, opened 2022-08-07
Added out of bounds read detection for corrups section headers

halos:
Parsing a malware file I found that the creators of it modified the `e_shoff` and `e_shnum` filling them with `0xFF` and the file was still executable. While `readelf` was able to obtain information from it (obviously not all the information), `pyelftools` crashed when it tried to parse the file in the initialization process of `elftools.elf.elffile.ELFFile`.

I was going to add the corrupt file to the `testfiles_for_readelf` folder, but it'll require more modifications to be done. I'll wait until you like these changes I propose or if you prefer to do that checks in other way.

Don´t worry about the ELF file I added to the tests folder, it's not malware ;)


eliben (commented):
This looks fine, thank you. Maybe @sevaa would like to take a look as well

What happens when you add the test file to the readelf tests?

Also, what's less clear is the general strategy - say `_get_section_header`, which would previously either return the header or raise an exception. Now it can also return `None` -- what does that mean? Should it instead raise `ELFParseError`? Or would that also be a "crash"?

sevaa:
This moves the point of failure on this particular class of file corruption a bit further down. The field `ELFFile._section_header_stringtable`, which is normally used by ELFFile itself for section name retrieval, may be a None after the ELFFile construction. Any attempt to work with section names (on this particular file) will error out, and with a nondescript error, too.

Some careful parsing can be still done on the file. That said, this is a step down the rabbit hole of replicating other tools' behavior, including replicating what kind of corruption they would tolerate.

At the very least, may I ask that the all points of usage of `_section_header_stringtable` get a None check and raise an `ELFParseError` with a reasonable message (something like "The stringtable is corrupt")?

halos:
@eliben
If I execute the `run_readelf_tests.py` on this file, it fails:
```
$ test/run_readelf_tests.py test/testfiles_for_unittests/corrupt_sh.elf
Test file 'test/testfiles_for_unittests/corrupt_sh.elf'
@@ aborting - 'scripts/readelf.py -e' returned '1'

Conclusion: FAIL
```
Executing `$ scripts/readelf.py -h test/testfiles_for_unittests/corrupt_sh.elf` works the same as `readelf`, but other options will fail as they need to access some headers and the don't properly handle this corruption.

Raising an exception was the other way I thought about, but returning `None` was simpler to program and is a first approach to check what could be the best way to manage this kind of error.

What about creating a new exception? (Only if it is useful) Something like `CorruptedELFError`. Of course, `ELFParseError` is also a good option.

halos:
@sevaa The idea with this first commit is to at least allow the `__init__` to finish its execution without crashing, but this patch will require more checks around the code.
Of course, I'll add that that checks.

sevaa:
I think `ELFParseError` will do, no need for a new error class.

I'm not sure what would be the correct behavior of the autotest here. Do we make an "error matches error" rule in the test harness, i. e. if GNU readelf fails on a file, then readelf.py is expected to also fail and that's a "test passed" condition?

sevaa:
Thanks.

sevaa:
@halos: I suggest you run the autotests before pushing your commits. Saves a lot of time. If you are a Windows user, put together a Linux VM, or use WSL.

halos:
Sorry! I run that test, but only with python3 (I have no python2 in my main machine), and I didn't know `seek` doesn't return anything in python2. Next time I'll run the test in both python.

eliben (approved):

PR #435, opened 2022-08-11
Cache instantiation of DWARF structs.

rdunklau:
Hello,

I've taken an interest in profiling some of my code using pyelftools. It turns out that a lot of time is spent instantiating DWARFStructs, when they are realistically called with the same arguments over and over again. 

I decided to run a quick benchmark on the postgres debug symbols I'm working with.

```
import sys

with ELFFile.load_from_path(sys.argv[1].encode('utf8')) as f:
    dwarf_info = f.get_dwarf_info()
    cus = list(dwarf_info.iter_CUs())
    print(f"Number of compile units: {len(cus)}")
```

On unpatched master:

```
# time python /tmp/test.py /var/lib/machines/fedora/usr/lib/debug/usr/bin/postgres-14.3-2.fc36.x86_64.debug  
Number of compile units: 6229

real    0m16,524s
user    0m16,309s
sys     0m0,200s
```

With this small instantiation cache:

```
# time python /tmp/test.py /var/lib/machines/fedora/usr/lib/debug/usr/bin/postgres-14.3-2.fc36.x86_64.debug  
Number of compile units: 6229

real    0m0,142s
user    0m0,122s
sys     0m0,020s
```


sevaa:
Which version of Python was `__new__` introduced in?

rdunklau:
I'm not sure, but I think it was something like 2.2. Ages ago, basically :-)

rdunklau:
I changed the cache to be a class member of DWARFStructs, and making everything private. 

I'm not sure changing the class instantiation is the best way to go about implementing that, but it was the least invasive to demonstrate what I was trying to do. I'm not really seeing any use case where you wouldn't want the cache so it's probably fine.

sevaa:
Why the need for `__inited`? The docs say `__init__` is called on the new instance right after `__new__` returns one; how can a double init possibly happen?

Also, I believe the convention is for a single underscore before private member names, isn't it?

rdunklau:
> Why the need for __inited? The docs say __init__ is called on the new instance right after __new__ returns one; how can a double init possibly happen?
It's because we don't want __init__ to be called a second time when we return the instance from the cache.
Admittedly, we could perform the whole initialization in the __new__ method itself, and get rid of the __init__ implementation entirely.

> Also, I believe the convention is for a single underscore before private member names, isn't it?
Sorry, I don't know the conventions for this project. I assumed that hiding that implementation detail by using double underscores and the name jumbling that it triggers would be a good idea, but I can revert to a single underscore if you prefer.

sevaa:
> we don't want init to be called a second time when we return the instance from the cache.

Understood. Missed that.

>  the conventions for this project

I'm just looking at the bulk of the existing code.

eliben:
A higher-level question first: why do structs get loaded multiple times at all for a single `DWARFInfo`?

Is this because of the new supplementary dwarfinfo functionality?

sevaa:
Bitness and DWARF version can legitimately vary between CUs (not sure about endianness). Also, in that one place you construct a temp one just to parse the first few fields of the header, then construct the real thing based on said few fields.

elftools/dwarf/structs.py
```diff
@@ -76,6 +76,18 @@ class DWARFStructs(object):
...
         See also the documentation of public methods.
     """
+
+    __StructsCache = {}
+
+    def __new__(cls, little_endian, dwarf_format, address_size, dwarf_version=2):
+        key = (little_endian, dwarf_format, address_size, dwarf_version)
+        if key in cls.__StructsCache:
+            return cls.__StructsCache[key]
+        newinst = super(DWARFStructs, cls).__new__(cls)
```
eliben, line 86: Nit: I believe you should be able to do just `super()...` here

eliben (changes requested):
As was mentioned in the discussion with @sevaa, IMHO this would be simpler without `__init__` at all. Then `__inited` would not be needed.

In a way, `DWARFStructs` is just a data container with no behavior and a rather complicated construction sequence

elftools/dwarf/structs.py
rdunklau: I couldn't because Python 2 support was not dropped at that time :-) 
Done.

rdunklau:
@eliben I did implement the requested changes. Thanks for looking at it !

eliben (approved):

eliben:
Thanks for the contribution!

PR #440, opened 2022-08-21
Corrupt ELF e_shstrndx handling

halos:
Added test file and test for EFL files with corrupt `e_shstrndx`.
Added code to allow `ELFFile` to execute its `__init__` without failing when loading an ELF file with corrupt `e_shstrndx`.


elftools/elf/elffile.py
```diff
@@ -549,7 +549,11 @@ def get_shstrndx(self):
         if self['e_shstrndx'] != SHN_INDICES.SHN_XINDEX:
             return self['e_shstrndx']
         else:
-            return self._get_section_header(0)['sh_link']
+            header = self._get_section_header(0)
```
eliben, line 552: I'm not sure why this protection exists. Here we can't even get the section header for the first section in the file -- what kind of corruption is this and what do you expect to happen?

halos: This protection comes from a malware sample I tried to open with `ELFFile` but the `__init__` crashed trying to access the tampered section header.

[This](https://www.virustotal.com/gui/file/8b9bfe8d5d32d7218059fcd24660a15a177a4ee75670cc1af86b357d59903cc7/detection) is one of the multiple samples that crash when I try to open it. This sample is completely executable although their headers are a little bit tampered as you can see in VirusTotal in the ["Details" tab](https://www.virustotal.com/gui/file/8b9bfe8d5d32d7218059fcd24660a15a177a4ee75670cc1af86b357d59903cc7/details), "ELF info" section.

I expect `ELFFile` doesn't crash during the initialization as I need to get info from the program headers. I'm using `pyelftools` [here](https://github.com/NozomiNetworks/upx-recovery-tool/blob/main/upx-recovery-tool.py#L187) to know the end of the program headers. The file ([test/testfiles_for_unittests/corrupt_e_shstrndx.elf](https://github.com/eliben/pyelftools/pull/440/files#diff-22865a7128b02800febcf128350860b88afb51c44b823d6bd28bc0223c6bf0a9)) has similar headers to that malware samples.

halos: This protection exists to fool or crash some security tools [like mine](https://github.com/NozomiNetworks/upx-recovery-tool/blob/main/upx-recovery-tool.py#L187).

eliben (commented):
So in case of a corrupt file, you're just using ELFFile to read the header because you can't do much else, right?

I'm not sure I like these changes, TBH. When the first one was sent, this was my concern too - there will be a propagation of `if None...` checks throughout the codebases, making it really hard to debug actual problems.

You can still easily use parts of `elftools` to parse just the program header if you want, without loading the whole `ELFFile`

halos:
It's true that now I only use `elftools` to get a couple of values from the header but I'll use it for a deeper parsing in the future. But my project is not the important point here. I can do some workarounds into the files before, but I thought it could be interesting sharing it with you if you think this could be interesting for the library.

On the one hand I know these modifications I'm proposing are intended for a couple of weird EFL files that are only a very very small percentage out of the executables this library is parsing every day. That's why I only suggested changes related to the functions used in the `__init__`. In this case, if I'd need to access `self._section_header_stringtable`, that I know it could potentially be `None`, I'd check that in my own code if that header is available in the `ELFFile` instance. I was only interested on the initial ELF parsing and I didn't want to modify more code beyond that.

On the other hand, the VERY low number of ELF files with this kind of weird fields that is making the code more complex, makes me completely understand your point of view. I can always create a fork or use only the portions of your code I need. :)

halos:
I've read [a short article](https://dustri.org/b/screwing-elf-header-for-fun-and-profit.html) where a researcher checks which ELF header fields can be safely tampered (for a specific executable) to fool analysis tools and keep the file executable. Only 3 of them could be modified without affecting that executable: `e_shoff`, `e_shnum` and `e_shstrndx`. It seems there will be no need to add additional checks for the ELF header in the case you accept the PR.

eliben:
Thanks for the extra details.

Having thought about this a bit more, I'm going to reject the PR, keeping pyelftools simple. Corrupted ELF files are not the goal of pyelftools, and these `None`\s will make error handling sub-optimal for 99.9% of the library's users

Issue #405, opened 2022-05-19
suggest add -W para for readelf.py


sfncat:
readelf.py -s have the fix length 25
 self._emitline('%6d: %s %s %-7s %-6s %-7s %4s %.25s%s' % (
suggest like readelf add -W para, so the ndx string can display all.
       -W
       --wide
           Don't break output lines to fit into 80 columns. By default
           readelf breaks section header and segment listing lines for
           64-bit ELF files, so that they fit into 80 columns. This
           option causes readelf to print each section header resp. each
           segment one a single line, which is far more readable on
           terminals wider than 80 columns.

sevaa:
As far as I understand, readelf.py only exists so that one may match the parsing results of `pyelftools` with those of regular `readelf`. It was never meant as a first class ELF file visualizer.

WillDenissen:
I have a branch ready that add -W -w support to readelf.py which passes all tests. I'm new however to github, so how do I create a pull request addressing this issue?

sevaa:
@sfncat I though we agreed we are not adding -W support to readelf. Let's close this.

sevaa:
@sfncat
@WillDenissen 

Can we please close this?

eliben:
Closing.

Feel free to reopen if new, relevant information comes up

PR #441, opened 2022-09-14
Wide support

WillDenissen:
Implements issue https://github.com/eliben/pyelftools/issues/405 for -W --wide support for readelf.py

### DIFF
--- a/scripts/readelf.py
+++ b/scripts/readelf.py
@@ -104,6 +104,8 @@ class ReadElf(object):
 
             output:
                 output stream to write to
+            args:
+                command line arguments to apply
         """
         self.elffile = ELFFile(file)
         self.output = output
@@ -252,8 +254,7 @@ class ReadElf(object):
             (Elf file type is...)
         """
         self._emitline()
-        nseg = self.elffile.num_segments()
-        if nseg == 0:
+        if self.elffile.num_segments() == 0:
             self._emitline('There are no program headers in this file.')
             return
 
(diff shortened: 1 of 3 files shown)


scripts/readelf.py
```diff
@@ -98,7 +98,7 @@ def _get_cu_base(cu):
 class ReadElf(object):
     """ display_* methods are used to emit output into the output stream
     """
-    def __init__(self, file, output):
+    def __init__(self, file, output, args):
```
eliben, line 101: Please add docstring section for the new parameter `args`

scripts/readelf.py
```diff
@@ -251,20 +252,24 @@ def display_program_headers(self, show_heading=True):
             (Elf file type is...)
         """
         self._emitline()
-        if self.elffile.num_segments() == 0:
+        nseg = self.elffile.num_segments()
```
eliben, line 255: I'd rather you just inline the call everywhere, this isn't much of a performance optimization anyhow

eliben (commented):
Quick round. Would be happy if @sevaa took a look at the PR as well

sevaa:
I can only reiterate that I don't see the point of this feature, unless someone wants to repurpose readelf.py as a more or less drop-in replacement for GNU readelf. Which is only needful if one is parsing its output for further programmatic processing, which I can't possibly endorse, since my preferred approach would be writing your CLI based on pyelftools.

From what I can see, autotest against `readelf -W` with GNU is not in the books. That's good, since attempting that would incur a ton of maintenance down the road, as it would cause twice the amount of output format discrepancies.

That said, the core library is intact. Merge away, no hard veto from me.

eliben:
@sevaa makes a good point, and I'm not sure you've addressed that in the original issue, @WillDenissen. Indeed, full feature parity between `readelf.py` and the GNU tool is not a goal of the project - it's a huge amount of work overall and the benefit is unclear. Obviusly, `readelf.py` demonstrates that such a tool _can_ be built using `elftools`, and I encourage others to develop it separately, if desired.

Can you please comment whether this PR improves the coverage of the elftools library itself, or whether it's primarily an extension to the `readelf.py` tool?

WillDenissen:
This PR does not improve the coverage of the elftools library itself. It only aligns the readelf.py output with the readelf reference implementation. It was for me an exercise to implement a "harmless feature" and go through the PR workflow for the first time.

sevaa:
The reference implementation itself is a moving target, too. The version that we have here has spotty DWARFv5 support, and the latest master has gratuitous changes in output format here and there.

Issue #311, opened 2020-05-10
[BUG] [DWARF] Missing Compile Units


SjRNMzU:
I'm using the latest git version of pyelftools with python 3.8.

`pyelftools` misses CUs that `objdump` can display when dumping DWARF info from MUSL libc.

The ELF64 binary I'm analyzing can be found [here](https://transfer.sh/Ai5XH/libc.so).

If I iterate through all the CUs using `dwarf_info.iter_CUs()` `pyelftools` misses all CUs from offset `0x66c71` and above.

Trying to specify the CU explicitly returns: 
```
In [3]: dwarf_info.get_CU_at(0x66c71)                      
---------------------------------------------------------------------------
DWARFError                                Traceback (most recent call last)
~/abc/src/classes/dwarf.py in <module>
----> 1 dwarf_info.get_CU_at(0x66c71)

/usr/local/lib/python3.8/dist-packages/pyelftools-0.26-py3.8.egg/elftools/dwarf/dwarfinfo.py in get_CU_at(self, offset)
    191             self.has_debug_info,
    192             'CU lookup but no debug info section')
--> 193         dwarf_assert(
    194             0 <= offset < self.debug_info_sec.size,
    195             "offset %s beyond .debug_info size" % offset)

/usr/local/lib/python3.8/dist-packages/pyelftools-0.26-py3.8.egg/elftools/common/utils.py in dwarf_assert(cond, msg)
     81     """ Assert that cond is True, otherwise raise DWARFError(msg)
     82     """
---> 83     _assert_with_exception(cond, msg, DWARFError)
     84 
     85 

/usr/local/lib/python3.8/dist-packages/pyelftools-0.26-py3.8.egg/elftools/common/utils.py in _assert_with_exception(cond, msg, exception_type)
    112 def _assert_with_exception(cond, msg, exception_type):
    113     if not cond:
--> 114         raise exception_type(msg)

DWARFError: offset 420977 beyond .debug_info size
```

Dumping DWARF info with `objdump --dwarf` produces the following excert:

```
...
<1><66c70>: Abbrev Number: 0
  Compilation Unit @ offset 0x66c71:
   Length:        0x33d (32-bit)
   Version:       4
   Abbrev Offset: 0x272a1
   Pointer Size:  8
 <0><66c7c>: Abbrev Number: 1 (DW_TAG_compile_unit)
[...]

Issue #193, opened 2018-08-08
Support for .debug_types to support DWARFv4


dinkar1208:
Compilers with DWARFv4 support (eg:- IAR embedded workbench) tend to put enums and typedefs present in header files into the .debug_types section. I would be good to have support to parse this section as well.

dinkar1208:
Without support for this section, readelf fails to decode debug_info section as it attempts to read hash signatures (DW_FORM_ref_sig8) as abbrev_table offsets.

eliben:
Please see my reply in #201

PR #442, opened 2022-09-15
Added Dwarfv5 CU headers

WillDenissen:
Added support for parsing DWARFv5 7.5.1.{1,2,3} CU headers, and dumping in readelf.py

Aligned readelf.py output with reference readelf
Added test file compiled with -gdebug_types-section to readelf tests

Partially addresses Issues: 
- issue: #325 #311 

Note it does not address issue #193


eliben:
@sevaa please help review

sevaa:
Okay, the problem is real - the CU header format in v5 varies with unit type, which is one of the leading fields. There are two possible flavors there. 

The proposed patch slightly breaks the API - the field header.unit_type used to be numeric, the patch makes it a string from the `DW_UT_` enum.

Caching/reuse of lineprograms is a good idea regardless.

eliben:
> Okay, the problem is real - the CU header format in v5 varies with unit type, which is one of the leading fields. There are two possible flavors there.
> 
> The proposed patch slightly breaks the API - the field header.unit_type in `CompileUnit` used to be numeric, the patch makes it a string from the `DW_UT_` enum.
> 
> Caching/reuse of lineprograms is a good idea regardless.

This sounds good. Does the PR LGTY?

sevaa:
Do we take an issue with the change of that one header field's format that I've mentioned?

eliben:
> Do we take an issue with the change of that one header field's format that I've mentioned?

I think it's fine, seems like a reasonable change to clarify an untyped int to an actual enum. As mentioned before, I'm not hugely worried about API changes at this level of specificity.

sevaa:
Okay, then no objections.

eliben (approved):
Nice work!
Thanks for the contribution

PR #443, opened 2022-09-20
Add pyproject.toml

twizmwazin:
This adds a pyproject.toml for compatibility with PEP 517/518-based build systems. With this, modern pip and other tools can build a wheel without relying on fallback setuptools compatibility.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 46.4.0", "wheel"]
+build-backend = "setuptools.build_meta"


eliben:
Can you clarify how/if I'm expected to use this when building releases of pyelftools? Or is someone else going to be using this?

twizmwazin:
Strictly speaking, nothing needs to change in the release process. Including a pyproject.toml in the source distribution will indicate to pip and related tools like python-build that the project is built using setuptools rather than relying on those tools falling back to legacy setup.py-based behavior. pip's documentation is [here](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/), and setuptools' docs are [here](https://setuptools.pypa.io/en/latest/build_meta.html) if you would like to read further.

eliben:
Can you please say how you've tested this flow?

twizmwazin:
I'm not entirely sure what you are asking for, but as an example, when using `pip wheel .` to build a wheel, the output will either say `Building wheel for pyelftools (pyproject.toml)` or `Building wheel for pyelftools (setup.py)`. This indicates if pip is using the PEP 517 interface to build the wheel, or if it is falling back to invoking setuptools via setup.py. python-build does not seem to mention in the output which method it uses. Building a wheel using `python setup.py bdist_wheel` invokes the build system directly rather than through a generic frontend and is therefore not affected by this change. In all cases the same wheel is produced.

eliben (approved):

PR #445, opened 2022-10-13
add support for aarch64 to dwarfdump

nickdesaulniers:
llvm-dwarfdump prints
elf64-littleaarch64

We probably need more logic to differentiate between big endian (less common) vs little endian (more common), but at least this allows us to use ./scripts/dwarfdump on aarch64 targets now.

I probably should have added this to #318 / #317.

### DIFF
--- a/scripts/dwarfdump.py
+++ b/scripts/dwarfdump.py
@@ -342,7 +342,7 @@ class ReadElf(object):
         self.elffile = ELFFile(file)
         self.output = output
         self._dwarfinfo = self.elffile.get_dwarf_info()
-        arches = {"EM_386": "i386", "EM_X86_64": "x86-64"}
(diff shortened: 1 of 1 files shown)


eliben:
@sevaa for sanity checking

> I probably should have added this to #318 / #317.

The dwarfdump tool is new, @sevaa added it recently since `readelf` was falling behind with the newest DWARFv5 featuers

nickdesaulniers:
I found the implementation of LocationExpr parsing in scripts/dwarfdump very very helpful!

(I'm playing with something that can draw me a picture of what precisely my stack looks like, w/ crayons!)

sevaa:
LGTM.

Display names of various architectures/CPUs in tools are pretty much arbitrary, I should have gone to the sources of llvm-dwarfdump and copied theirs, but I didn't, The dwarfdump autotest was a (forced) step along the way, not the end goal, so I didn't go all out on that. What test binaries I had I've built support for.

eliben (approved):

PR #466, opened 2023-04-09
add reloc type and recipe for eBPF

flier:


eliben (approved):

eliben:
> Makes sense to me. The only comment I have - and that's a preexistent design, not the current PR's fault - is that we match ELF architecture codes against user visible display names, e. g. `self.elffile.get_machine_arch() == 'Linux BPF - in-kernel virtual machine'`. That's fragile.

I agree, while in the previous clauses it looked reasonable (`x64` is unlikely to change at this point), this PR highlights that these name checks could be fragile.

I'll be happy to review a PR with a thorough cleanup/fix of the underlying issue.

Issue #422, opened 2022-06-21
DW_FORM_indirect not parsed properly


sevaa:
I don't have a binary for this, but I have a crash report that DW_FORM_indirect exists out in the wild.

The encoding for the form is: ULEB128, interpreted as a DWARF form code, followed by the value in that form. It's effectively dynamic typing in DIEs. The current implementation only reads the ULEB128, doesn't read the actual value. As a result, the current position is off, subsequent attributes and DIEs (if enumerating) are misread.

EDIT: now that I look at it, there is a branch in _translate_attr_value, written under the assumption that _translate_attr_value is called immediately after reading the raw value. That parser, however, loses the real form of the attribute - not good. The form carries a lot of information, especially in later DWARF versions.

0xC0DED00D:
dwarfdump.py v0.29 raises on a DW_FORM_indirect in the .o file here...
[newlib.zip](https://github.com/eliben/pyelftools/files/11523228/newlib.zip)

sevaa:
Can't reproduce on pyelftools 0.29, as packaged with DWARF Explorer. Parses and shows all DIEs. What is the parsing code on your end?

Also, this is definitely not the same issue as originally reported.

0xC0DED00D:
I did this:
1. Download and extract zip of v0.29 (https://github.com/eliben/pyelftools/archive/refs/tags/v0.29.zip)
2. Change `attr` at die.py line 125 to `repr(attr)`, and change the dict lookup at dwarfdump.py line 346 to `arches.get(self.elffile['e_machine'], self.elffile['e_machine'])` to suppress some (I think) unrelated errors:
```diff
diff -r pyelftools-0.29-1/elftools/dwarf/die.py pyelftools-0.29-2/elftools/dwarf/die.py
125c125
<             raise DWARFError('%s is not a reference class form attribute' % attr)
---
>             raise DWARFError('%s is not a reference class form attribute' % repr(attr))
diff -r pyelftools-0.29-1/scripts/dwarfdump.py pyelftools-0.29-2/scripts/dwarfdump.py
346c346
<         arch = arches[self.elffile['e_machine']]
---
>         arch = arches.get(self.elffile['e_machine'], self.elffile['e_machine'])
```
3. Run dwarfdump.py, supplying pyelftools-0.29 in PYTHONPATH:
```
[...]

PR #475, opened 2023-06-27
Reporting the real attribute form for DW_FORM_indirect

sevaa:
I'd like a binary to write a test case.

The link from 455 is bogus, this is a response to #422,


eliben:
Yep, let's wait for a binary

sevaa:
Used one from #422.

eliben (approved):
Thanks!

Issue #180, opened 2018-02-14
from Construct developer, a note


arekbulski:
I am the developer of Construct (a parsing library) since I took over the project 2 years ago. Please refer any other projects that also use this library to this post, or copy it over to them. 

Construct is undergoing heavy changes at the moment, similar to those between 2.5->2.8, including addition of compiler feature (to make it much faster than Kaitai), but also several classes are going to be removed (for a good reason), and also few were added. API will be unstable for about a month.

What you can do:
- All scheduled work is posted as tickets, in [Issues](https://github.com/construct/construct/issues) and also on [Kanban](https://github.com/construct/construct/projects/3) page. If you subscribe (watch repository) then you will have advance notice of any incoming changes. Unfortunately those tickets also include minor work items, so you would get some clutter notifications too. I appologise for that.
- I encourage everyone to browse those tickets, vote for feature requests, but also post objections for planned removal of features, or changes in semantics. Subscribing therefore is recommended.
- You can find what was already changed (at any moment) on [Transition page](https://construct.readthedocs.io/en/latest/transition29.html), which is like a growing changelog. It gets updated whenever something gets added or removed, but since the order is not chronological, new items do not end up on top or bottom of the list.
- You can find currently available classes and semantics on revamped [Core API pages](https://construct.readthedocs.io/en/latest/index.html#api-reference). Those pages are very extensive and always uptodate with the implementation.
- **Feel free, at any point in time, to open a new Issue and ask questions or make complaints** about breakage. At minimum, I can offer advice and possible alternatives. People very often use the classes in ways that were not designed for, like use String classes to process Bytes.
- **Feel free, at any point in time, to open a new Issue and request changes.** If you need a new class, and make a good clear case, I would be more than willing to add it to core library to support your project.
- API should be stable within a month, so either do version pinning and deal with it then, or keep fixing your code at every release (which is ~3 day cycle at the moment).

Close this topic at your discretion.

arekbulski:
Since this project used to depend on 2.5 which is 2 YEARS OUT OF DATE, I imagine it would require some work to bring it up to date. I am willing to help you out with it, but not do the entire work.

eliben:
@arekbulski thanks for the notice. Would it be correct to say that the API has changed significantly from 2.5 to 2.8, meaning that I'll have to rewrite most of my construct-using code?

arekbulski:
[...]

PR #477, opened 2023-07-10
Migrate to Construct 2.10

SupremeMortal:
Move away from an internal copy of Construct to the latest version in order to support new DWARFv5 data types.


eliben:
Thanks, this is a very large change, PTAL at the CI test failures... Can you clarify how this is required to support DWARFv5 stuff?

SupremeMortal:
Some of the form types mentioned [here](https://github.com/eliben/pyelftools/pull/363#issuecomment-847269798) require `UBInt24`. Although it would be possible to add these to the existing library, I noticed that it was on the TODO list to use the updated construct library so I thought I'd give it a shot.

Will investigate the CI failures.

SupremeMortal:
CI and unit test issues have been resolved.

elftools/elf/sections.py
```diff
@@ -480,38 +480,36 @@ def __init__(self, structs, stream):
...
 
             if self.tag != 'TAG_FILE':
                 self.extra = []
-                s_number = struct_parse(structs.Elf_uleb128('s_number'), stream)
+                s_number = struct_parse(structs.Elf_uleb128, stream)
 
                 while s_number != 0:
                     self.extra.append(s_number)
-                    s_number = struct_parse(structs.Elf_uleb128('s_number'),
+                    s_number = struct_parse(structs.Elf_uleb128,
```
eliben, line 491: Nit: fold into one line here too

elftools/elf/sections.py
```diff
@@ -480,38 +480,36 @@ def __init__(self, structs, stream):
...
                                             stream)
 
         elif self.tag in ('TAG_CPU_RAW_NAME', 'TAG_CPU_NAME', 'TAG_CONFORMANCE'):
-            self.value = struct_parse(structs.Elf_ntbs('value',
-                                                       encoding='utf-8'),
+            self.value = struct_parse(structs.Elf_ntbs,
```
eliben, line 495: ditto here and below

elftools/elf/segments.py
```diff
@@ -6,7 +6,9 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
-from ..construct import CString
+from construct import CString
+
```
eliben, line 10: Nit: remove blank line

elftools/common/construct_utils.py
```diff
@@ -101,12 +105,65 @@ class StreamOffset(Construct):
...
         return stream.tell()
-    def _build(self, obj, stream, context):
+    def _build(self, obj, stream, context, path):
         context[self.name] = stream.tell()
-    def _sizeof(self, context):
+    def _sizeof(self, context, path):
         return 0
+
+
+class EmbeddableStruct(Struct):
```
eliben, line 118: So the newer construct versions removed these? I wonder if they offer any alternatives -- writing these custom structs always felt like a kludge

elftools/dwarf/callframe.py
```diff
@@ -277,17 +277,19 @@ def _parse_cie_augmentation(self, header, entry_structs):
...
+            b'z': 'length' / entry_structs.Dwarf_uleb128,
+            b'L': 'LSDA_encoding' / entry_structs.Dwarf_uint8,
+            b'R': 'FDE_encoding' / entry_structs.Dwarf_uint8,
             b'S': True,
-            b'P': Struct(
-                'personality',
-                entry_structs.Dwarf_uint8('encoding'),
-                Switch('function', lambda ctx: ctx.encoding & 0x0f, {
-                    enc: fld_cons('function')
+            b'P': 'personality' / Struct(
```
eliben, line 284: Can you fix the formatting here a bit? It's hard to see that the following fields are inside the Struct
Maybe indent these parts more, or extract them into a separate var before `available_fields`?

eliben (changes requested):
Overall very nice! Still thinking about this, left some initial comments for now

eliben:
We're long overdue on this :-) One thing that is slightly annoying is that we're adding *more* special construct utils/parsers instead of removing existing ones!

@sevaa - please take a look as well, when you have some time.

@SupremeMortal did you have a chance to measure performance? I want to make sure we're not regressing here

elftools/common/construct_utils.py
SupremeMortal: Unfortunately, there are no replacements as mentioned here https://github.com/construct/construct/issues/1027#issuecomment-1498657219

I did however manage to remove `RepeatUntilExcluding`

SupremeMortal:
I have not measured performance properly yet, but on my system, the unit tests now take on average around 9.1 seconds compared to before when it took 8.3 seconds. There are some optimisations that can be made to improve it which I will investigate.

eliben:
> I have not measured performance properly yet, but on my system, the unit tests now take on average around 9.1 seconds compared to before when it took 8.3 seconds. There are some optimisations that can be made to improve it which I will investigate.

Running `python3 test/all_tests.py` takes around half a minute (wallclock time) at least on my machine, so this can provide a bit more runway for comparison.

sevaa:
On general grounds, I welcome not forking vendor packages gratutously :) That said, I'll take a closer look.

SupremeMortal:
I'm unsure if the performance will be good enough. I tested with a 1.5GB ELF file with DWARF v5 debug, and it took 20 seconds with this PR compared to 10 seconds without which is quite a large regression in performance. 

I also tried optimising structs with the new lazy-loaded structures in construct 2.10, but it ended up being unusable for most of the structs in this codebase.

eliben:
@SupremeMortal thanks for the update! This is indeed concerning. pyelftools doesn't excel at performance in general, so making it 2x slower for large inputs is a hard pill to swallow.

@arekbulski cc-ing you, given the discussion in #180

sevaa:
So a new convention to learn. Doesn't look too terrible. It did seem unnatural to me all along that parsers needed to have a name attached to them just in case the parser is embedded in a struct.

As for compiling, we could start by storing the basic building blocks (trivial DWARF forms, nondynamic headers, etc) in compiled form and see where it takes us.

arekbulski:
Compilation is a feature in Construct.

sevaa:
I have some bandwidth currently, wanted to revisit this. There's been several releases of Construct since, maybe they've done something about performance - or maybe we could apply some tweaks on our side. I don't see @SupremeMortal publishing any changes since the initial commit - if they looked at the performance issues, it was not mature enough for publishing, and anyway, the fork needs to be merged with the recent changes to the master. If @SupremeMortal doesn't respond in a few days, I could fork the fork and sort of take over.

sevaa:
FYI, for me, with this PR's fork and the latest construct, the performance difference with the master is not that great - the master is ~2% faster for a firehose parsing of DWARF CU/DIEs in a subset of our corpus. That's two percent, not times two,

eliben:
> FYI, for me, with this PR's fork and the latest construct, the performance difference with the master is not that great - the master is ~2% faster for a firehose parsing of DWARF CU/DIEs in a subset of our corpus. That's two percent, not times two,

Interesting. I wonder what benchmark @SupremeMortal is alluding to in https://github.com/eliben/pyelftools/pull/477#issuecomment-1638443003

2% is definitely tolerable!

eliben:
Superseded by #548

PR #458, opened 2023-01-29
add support for loongarch64 to dwarfdump

haomi123:
add support for loongarch64

Issue #479, opened 2023-07-12
Insane memory consumption because of caching mechanism


ShellCode33:
Hey there, thanks a lot for your awesome work. I'd like to report an issue I have in regard to memory consumption.

I'm working with a huge ELF file (914MiB) and I'm trying to iterate over DIEs. Thing is I am unable to do so because the OOM killer kicks in at some point (FYI I have 32GiB of RAM). I'm only interested in a few DIEs from the whole lot so there's no need for me to cache everything.

Though I managed to hack my way through by reimplementing methods of `CompileUnit` and `DIE` to remove cache usage, I think it would be great to be able to disable the cache completely if there's no need for it.

I don't have time to implement this right now, but maybe someone could add an optional use_cache parameter in `CompileUnit`'s constructor that would enable users of this library to choose what they want (CPU time vs RAM usage).

sevaa:
Did I hear that right - 32 MB of RAM on the machine where you parse? Forget about Python and rework in C. The Python runtime alone will happily eat most of that.

ShellCode33:
Oops, I meant 32GiB of RAM obviously

eliben:
In general, this library is probably not a great fit for very large inputs. It has several performance issues and wasn't designed to ingest enormous ELF binaries.

ShellCode33:
Yes that's what what I figured, I went with `gimli` in Rust and wrote Python bindings

sevaa:
Given its intensely interlinked nature (even more so in v5), DWARF doesn't easily lend itself to firehose, cache-nothing parsing. Take, for example, abbrevs. There is a reference to the abbrev table in every DIE. Would be crazy to return to the abbrev table and parse one again on every DIE.

Also, to begin with, pyelftools starts by pulling every debug related section into a memory buffer, and does the rest of work against those. That alone will add some gigs to your working set.

I can envision a version of pyelftools that is optimized for memory, but the one we got isn't it.

sevaa:
@ShellCode33: what exactly was your desired use case? I can see top-to-bottom iterating through stuff sans caching, but random access sans caching will incur a crazy amount of I/O.

PR #483, opened 2023-08-02
Fix the LoongArch support code and some more

xen0n:
There were #458 and #470, but unfortunately both were partly incorrect. I've fixed them and added some more support for LoongArch relocations while at it.


sevaa:
We don't have any LoongArch binaries to autotest against, do we?

xen0n:
I tried to add some tests but IIRC the readelf test failed because the prebuilt binary is too old to recognize some of the new LoongArch relocs (or the architecture itself), and at that time I don't have enough time to look into this. I may be able to follow up a few days later though.

eliben:
It may be time to bump the pre-built readelf to recognize LoongArch; otherwise I feel like we're stretching it without real files to test against.

sevaa:
The latest readelf will break a whole bunch of autotests, at least on the DWARF side. There is a reason we carry around our own. That said, that bullet needs to be eventually bitten.

sevaa:
Try grabbing an updated copy of `readelf` from  #479.

xen0n:
I'm rebasing this PR, hoping to post an updated revision this week.

eliben:
@xen0n if you want this PR to make it into the next release of pyelftools, please address the comments soon

xen0n:
> @xen0n if you want this PR to make it into the next release of pyelftools, please address the comments soon

Thanks for the reminder; I'll try to do this tonight.

xen0n:
Update: I'm blocked by the readelf DWARF test cases, still working my way through missing generic codepaths that get exercised by the LoongArch test binary I just made.

sevaa:
The basic ELF/LoongArch binary is a part of the autotest, the relocations is the only architecture specific part that needed any nontrivial treatment and that's done. Friendly register names on RISC platforms, as we've concluded, are nowhere on GNU binutils' radar and therefore shouldn't be on pyelftools', either. :) :(

I think this is good to go. 謝謝

elftools/elf/relocation.py
```diff
@@ -322,6 +331,12 @@ def _reloc_calc_sym_plus_addend(value, sym_value, offset, addend=0):
     def _reloc_calc_sym_plus_addend_pcrel(value, sym_value, offset, addend=0):
         return sym_value + addend - offset
 
+    def _reloc_calc_value_plus_sym_addend(value, sym_value, offset, addend=0):
```
eliben, line 334: If you rebase this PR, one of the recent commits made `_reloc_calc_sym_plus_value` do the same thing -- can that be used instead?

scripts/readelf.py
```diff
@@ -95,6 +96,11 @@ def _get_cu_base(cu):
     else:
         raise ValueError("Can't find the base IP (low_pc) for a CU")
 
+_CONTROL_CHAR_RE = re.compile(r'[\x01-\x1f]')
```
eliben, line 99: Please add a comment explaining what this is for

eliben (changes requested):

elftools/elf/relocation.py
xen0n: Done, thanks!

scripts/readelf.py
xen0n: I've added an explanation for it (meant for use in `_format_symbol_name`).

Issue #484, opened 2023-08-14
Cortex M33 elf binary error: no decoding mapping for 46 [tag] 


KShingala:
I am trying to use angr to generate control flow graph for CM33 binary but get this error from elftool from backend_z3.py.

I am using the following in my script to 

<code>
arch = archinfo.ArchARM(archinfo.Endness.LE)
p = angr.Project('my_binary.elf', arch=arch)
</code>

Can someone help me what I am doing wrong?

sevaa:
Tag 0x46 is `DW_TAG_dynamic_type`. It's documented in DWARFv5, but for some reason not listed in the tag enum pyelftools. It's a fairly simple fix on our side, but who knows when will the current state of the master branch get released. I'll do a PR.

KShingala:
Here are the file I used to experience the error. [](url)
[example.zip](https://github.com/eliben/pyelftools/files/12409594/example.zip)

sevaa:
Reproduced. This is not about a DWARF tag 0x46, like I thought, the error is taking place while parsing the `.ARM.attributes` section, and the enum that is lacking 46 (decimal) is `ENUM_ATTR_TAG_ARM` under elf/enums.py.

Now I need to find an authoritative reference on those, figure out what does 46 stand for, and how does one parse the rest of the attribute record...

KShingala:
I dont mind using the binary as an autotest, please feel free to use it as you like.

sevaa:
Either way, the fix seems to be a new enum value - but in a different enum. See the changeset in #491, and see if you can apply that.

PR #487, opened 2023-08-14
Missing tags

sevaa:
Addresses #484


KShingala:
Dont think just adding enums is enough, I get the below error when I try you change,

elftools.construct.adapters.MappingError: no decoding mapping for 46 [tag]

sevaa:
From here it's impossible to tell what are you doing and what's the exact nature of the error. Who wrote backend_z3.py - is that your script, or from some third party package? If the latter, why don't you report the issue to its authors and see if they can debug it?

eliben (approved):
This looks straightforward, thanks

PR #489, opened 2023-08-19
readelf 2.41 with Ranges test excluded on 2 files

sevaa:
Current state of affairs - range list dump test shorted out

### DIFF
--- a/elftools/dwarf/ranges.py
+++ b/elftools/dwarf/ranges.py
@@ -114,14 +114,21 @@ class RangeLists(object):
         return struct_parse(self.structs.Dwarf_rnglists_entries, self.stream, offset)
 
     def iter_range_lists(self):
-        """ Yield all range lists found in the section according to readelf rules.
+        """ Yields all range lists found in the section according to readelf rules.
         Scans the DIEs for rangelist offsets, then pulls those.
+        Returned rangelists are always translated into lists of BaseAddressEntry/RangeEntry objects.
         """
-        # Calling parse until the stream ends is wrong, because ranges can overlap.
-        # Need to scan the DIEs to know all range locations
+        # Rangelists can overlap. That is, one DIE points at the rangelist beginning, and another
+        # points at the middle of the same. Therefore, enumerating them is not a well defined
+        # operation - do you count those as two different (but overlapping) ones, or as a single one?
+        # For debugging utility, you want two. That's what readelf does. For faithfully
+        # representing the section contents, you want one.
+        # That was the behaviour of pyelftools 0.28 and below - calling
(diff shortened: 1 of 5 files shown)


eliben:
For reference, the issue @sevaa opened with upstream binutils is https://sourceware.org/bugzilla/show_bug.cgi?id=30781

@sevaa how many test binaries fail due to output discrepancies if the ranges test is re-enabled?

sevaa:
Two:
dwarf_test_versions_mix.elf
dwarf_v5ops.so.elf

eliben:
> Two: dwarf_test_versions_mix.elf dwarf_v5ops.so.elf

2 / 43 isn't too bad; sounds like we can maybe selectively disable it, not running range tests just on those two files, leaving a TODO pointing at the binutils issue?

elftools/dwarf/ranges.py
```diff
@@ -114,14 +114,21 @@ def get_range_list_at_offset_ex(self, offset):
...
+        Returned rangelists are always translated into lists of BaseAddressEntry/RangeEntry objects.
         """
-        # Calling parse until the stream ends is wrong, because ranges can overlap.
-        # Need to scan the DIEs to know all range locations
+        # Rangelists can overlap. That is, one DIE points at the rangelist beginning, and another
+        # points at the middle of the same. Therefore, enumerating them is not a well defined
+        # operation - do you count those as two different (but overlapping) ones, or as a single one?
+        # For debugging utility, you want two. That's what readelf does. For faithfully
+        # representing the section contents, you want one.
+        # That was the behaviour of pyelftools 0.28 and below - calling
```
eliben, line 126: Not sure I get this part -- the behavior doesn't appear to change in this PR, so what is the meaning of "this was the behavior of ... 0.28 and below"? Is something changing?

The docstring of this method already says "according to readelf rules", so it's clear which way the choice is made

sevaa: The long comment is me ~venting~ explaining why "list all rangelists" is a not well defined operation. I changed the pyelftools' behavior on that long time ago, back when implementing DWARFv5.

eliben (approved):

sevaa:
OBTW, there is some progress on the binutils readelf issue. Assuming the fix takes, would you prefer that we build `readelf` from their latest and integrate that, or wait for a numbered release?

eliben:
> OBTW, there is some progress on the binutils readelf issue. Assuming the fix takes, would you prefer that we build `readelf` from their latest and integrate that, or wait for a numbered release?

I think it's OK to build from the latest - we should just change the comment to explain exactly which commit hash it's built from.

PR #495, opened 2023-08-29
Support MIPS64 .o files - don't remove has_addend

noamraph:
A slightly modified https://github.com/eliben/pyelftools/pull/493


eliben (commented):
Thanks, this looks better to me. @sevaa WDYT?

eliben:
I like that this decouples the `has_addend` refactoring from the specific MIPS64 change. We can consider a separate `has_addend` refactoring PR once this one lands.

@sevaa specifically I'd like you to review the `run_dwarfdump_tests` changes here

sevaa:
@noamraph: may I hear more about the `(0x0000000000000000 ".text")` discrepancy? Where is it, exactly?

noamraph:
@sevaa here's the diff:
image

I tried modifying `dwarfdump.py` for just adding `".text"` for `DW_FORM_addr` with value=0, but it turns out that in other ELF files llvm-dwarfdump didn't add it. I'm not sure why llvm-dwarfdump added the `".text"` here and not in other files. It may be related to this being a `.o`.

I can modify `dwarfdump.py` to add `".text"` to  `DW_FORM_addr` with value=0 on "relocatable" ELF files. Would that be better than just ignoring it in the output?

sevaa:
Have you checked the dwarfdump sources?

noamraph:
@sevaa I tried to look in llvm-dwarfdump sources, and didn't find the place where it prints the address.
Is there a chance we could continue without going to the root of this? After all, the DWARF parsing is obviously correct. The dwarfdump output is also correct, it just doesn't match exactly the llvm-dwarfdump output. The question is, could we add support for MIPS64 .o DWARF parsing without understanding when exactly llvm-dwarfdump prints the section name.

Thanks,
Noam

sevaa:
Every time we introduce a special case treatment in the test runner, it leaves a bad taste in my mouth :) That said, no hard veto.

noamraph:
@sevaa I feel the bad taste as well, and I would really like to understand llvm-dwarfdump further and update `dwarfdump.py` accordingly. But there's a good chance I won't have the time. So I would understand it if you will reject adding support for MIPS64 .o in order to keep the test clean, but I would appreciate it if you would agree to add it.

Anyway, I pushed another commit which I think makes the special-case a bit more localized in `run_dwarfdump_test.py`, and adds a comment explaining it.

What do you think?

eliben (approved):

PR #500, opened 2023-09-08
Support for DW_OP_WASM_location

sevaa:
Another day, another crash report on my end about a vendor DWARF extension.

This one is documented at least: https://yurydelendik.github.io/webassembly-dwarf/ section 2.2

On a second look, WASM has its own executable binary format that DWEX doesn't support. I'm not sure how did DWARF opcode 0xED make it into an ELF/PE/MachO binary. Barring the potential case of a corrupted/misparsed expression blob, in theory it could be some other vendor extension with opcode 0xED.

### DIFF
--- a/elftools/dwarf/dwarf_expr.py
+++ b/elftools/dwarf/dwarf_expr.py
@@ -10,6 +10,7 @@ from collections import namedtuple
 from io import BytesIO
 
 from ..common.utils import struct_parse, bytelist2string, read_blob
+from ..common.exceptions import DWARFError
 
 
 # DWARF expression opcodes. name -> opcode mapping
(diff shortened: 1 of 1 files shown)


eliben:
Hmm, I'm also not sure how this gets into pyelftools at all, since WASM binaries aren't ELF. How are you even testing this?

sevaa:
I have no way to test this. What I have is a crash report, telling me that DWEX threw an exception on a user's machine: KeyError in parse_expr() on dwarf_expr.py:151 with key 237 (0xED). That line goes:

`op_name = DW_OP_opcode2name.get(op, 'OP:0x%x' % op)`

The whole expression blob goes [237, 2, 1, 159]. That's pretty much all I have. There are some contrived scenarios how this could be bogus - something was interpreted as a location that shouldn't be. But the most likely scenario is that this is indeed a two op expression,  opcode 0xED with two arguments followed by `DW_OP_stack_value`. The argument structure of 0xED is consistent with the linked docs.

FYI, pyelftools can be and is used to parse DWARF in non-ELF binaries. DWEX does that. However, DWEX doesn't support WASM binaries, and the crash report was specifically from DWEX, not from pyelftools in come other context.

eliben (approved):

eliben:
OK, thanks for the additional background

PR #501, opened 2023-09-13
False positive on LocationParser.attribute_has_location()

sevaa:
The description of `DW_AT_member_location` changed between DWARFv2 and DWARFv3:

v2, section 5.5.4:

> If the member entry is defined in the structure or class body, it has a
DW_AT_data_member_location attribute whose value is a location description

v3, section 5.5.6:

 >If a data member is defined in a structure, union or class, the corresponding member entry has a DW_AT_data_member_location attribute whose value describes the location of that member relative to the base address of the structure, union, or class that most closely encloses the member declaration. **If that value is a constant, it is the offset in bytes from the beginning of the innermost enclosing structure, union or class to the beginning of the data member.** Otherwise, the value must be a location description.

The case of DW_AT_data_member_location containing an integer constant is not handled by pyelftools - the method `LocationParser.attribute_has_location()` returns True, and it shouldn't.

---
I have a crash report where an attribute with the following data:

`AttributeValue(name='DW_AT_data_member_location', form='DW_FORM_data1', value=0, raw_value=0, offset=402, indirection_length=0)`

is being interpreted as a location list pointer (`LocationParser.parse_from_attribute()` tries to parse it), which in turn exceptions, because the binary doesn't contain a loclist section (`self.location_lists` is None).

### DIFF
--- a/test/test_dwarf_locationattr.py
+++ b/test/test_dwarf_locationattr.py
@@ -11,8 +11,27 @@ from elftools.dwarf.die import AttributeValue
 
 class TestLocationAttrubute(unittest.TestCase):
     def test_has_location(self):
+        # This attribute comes from a DWARFv3 binary that doesn't have a location lists
+        # section. Before the patch, pyelftools would interpret it as an attribute with a
+        # location, more specifically with a location list offset (as opposed to an expression).
+        # Meanwhile, by the spec, DW_AT_data_member_location is not even capable
+        # of storing a location list offset (since structure layout
+        # can't vary by code location). DWARFv3 spec also provides that DW_AT_data_member_location
+        # may be a small integer with an offset from the structure's base address, and that
+        # seems to be the case here. Ergo, pyelftools should not claim this attribute a location.
+        # Since the location/loclist parse function uses the same check, ths fix will 
+        # prevent such attribute values from being misparsed, also.
+        #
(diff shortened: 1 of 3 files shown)


sevaa:
While at it, also fixed a false *negative* in `attribute_has_location()` for the DWARFv5 counterparts of `GNU_call_xxx` attributes.

test/test_dwarf_locationattr.py
```diff
@@ -0,0 +1,21 @@
...
+# This code is in the public domain
+#-------------------------------------------------------------------------------
+import unittest
+
+from elftools.dwarf.locationlists import LocationParser
+from elftools.dwarf.die import AttributeValue
+
+class TestLocationAttrubute(unittest.TestCase):
+    def test_has_location(self):
+        attr = AttributeValue(name='DW_AT_data_member_location', form='DW_FORM_data1', value=0, raw_value=0, offset=402, indirection_length=0)
```
eliben, line 14: Please add a comment here explaining these special cases and why we expect `attribute_has_location` to return false or true here

eliben (changes requested):

test/test_dwarf_locationattr.py
sevaa: Provided.

sevaa:
Done.

eliben (approved):
Thanks!

PR #509, opened 2023-09-27
Support getting RELR relocations from dynamic section

medhefgo:


sevaa:
A binary to test that against would be  nice.

elftools/elf/relocation.py
```diff
@@ -182,6 +193,16 @@ def get_relocation(self, n):
             self._cached_relocations = list(self.iter_relocations())
         return self._cached_relocations[n]
 
+
+class RelrRelocationSection(Section, RelrRelocationTable):
```
eliben, line 197: I'd rather use composition than multiple inheritance, unless there's a strong reason to do the latter.

eliben:
@sevaa other than the test comment, LGTY?

sevaa:
Never really understood how do relocations work, but all this looks reasonable.

elftools/elf/relocation.py
medhefgo: It follows the same style already used by `RelocationSection`.

eliben (approved):

Issue #523, opened 2023-11-25
Support for the .gnu_debuglink section as part of the DWARF5 standard


Kungergely:
Would it be possible to implement support for the `.gnu_debuglink` section to make the DWARF5 standard compatibility more feature complete? Since the `.gnu_debugaltlink` and `.debug_sup` section support has already been implemented I don't think that this would be an excessive change. Yes, probably the constructor of the `DWARFInfo` object would have to be extended by yet another parameter, but `gcc` seems to rely on this section rather heavily lately to link to `.debug` files that contain debug info.

sevaa:
I don't see why not - but we'd need some kind of guidance. How is it different from gnu_debugaltlink?

Also, any chance for a sample binary?

sevaa:
@Kungergely: #525 addressed this, didn't it?

PR #525, opened 2023-12-01
Support for .gnu_debuglink DWARF5 Section

io-no:
Address #523 

I've added support for the .gnu_debuglink section in DWARF5 and made it compliant with the follow_links option. This involved modifying the DWARFInfo constructor to include the new section and creating a new structure in structs.py.

Then, I ran all the tests and updated test/test_refaddr_bitness.py to align with the changes in the constructor structure. All tests were executed successfully.


sevaa:
The PR looks good. We'd like a test binary, though.

Good point re: the bitness test. It was my doing; I observed the issue on a MachO binary and made a test out of that instead of trying to reproduce on an ELF one. I'll try to supply a less fragile test.

The purpose of `gnu_debuglink` and `gnu_debugaltlink` looks, well, identical. Do you know why both exist?

Looks like you ignore the padding and the checksum altogether. It's OK to not parse the section to the very end, you know :)

io-no:
Hi, I've added a test as requested. I tried to keep the structure and format consistent with the other tests, I hope it's okay.

Regarding the differences between the two sections, I admit that it remains a bit unclear to me as well. From what I understand, the `.gnu_debugaltlink` section is related to the use of tools such as dwz and is not really part of the standard, unlike `.gnu_debuglink` which is part of the standard.

I parsed padding and checksum for consistency with what I saw for the implementation of the `.gnu_debugaltlink` section. No, I don't use them. It might be an idea to use the checksum for future implementations where, in addition to opening the external file, its integrity is also verified. In the meantime, if you prefer, I can also ignore those two fields.

davea42:
On 12/13/23 15:05, Gabriele Digregorio wrote:
> Regarding the differences between the two sections, I admit that it 
> remains a bit unclear to me as well. From what I understand, the 
> |.gnu_debugaltlink| section is related to the use of tools such as dwz 
> and is not really part of the standard, unlike |.gnu_debuglink| which is 
> part of the standard.

.gnu.debuglink is not part of the DWARF standard.
It is not a section that the DWARF standard mentions.

David Anderson

sevaa:
@eliben

What's the holdup for this one?

eliben:
@sevaa LGTM

Is this LGTY? 
Can you use the code review tool here to LGTM PRs?

sevaa:
I thought I approved this (informally) in my Dec 1 comment. I don't think I have access to the Github review button here.

PR #528, opened 2023-12-08
elftools: StringTableSection: Check table size before returning string

vapier:
Some ELF files have strings pointing to an offset outside the string table dimension, let's throw an exception in that case.

Bug-Url: https://crbug.com/788925
Reviewed-on: https://chromium-review.googlesource.com/792553

### DIFF
--- a/elftools/elf/sections.py
+++ b/elftools/elf/sections.py
@@ -140,7 +140,19 @@ class StringTableSection(Section):
         """ Get the string stored at the given offset in this string table.
         """
         table_offset = self['sh_offset']
+        table_size = self['sh_size']
+        if table_size == 0:
+            return ''
+
+        elf_assert(offset < table_size,
+                   'Expected string offset %x < table size %x' %
+                   (offset, table_size))
+
         s = parse_cstring_from_stream(self.stream, table_offset + offset)
+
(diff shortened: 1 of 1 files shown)


eliben (commented):
Adding a unit test could be nice here

Otherwise LGTM

@sevaa any objections?

sevaa:
The second assert is unnecessary - `parse_cstring_from_stream` will not read past the end of the stream (this ain't C, people). If there is indeed an unterminated string at the end of the section, `parse_cstring_from_stream` will return `None` and the expression in the `assert` will throw an exception that you are calling `len()` on a `None` object.

Remove the second check, it's pointless and wrong.

sevaa:
@vapier Are you still interested in this change?

sevaa:
Now that I look at it, should we really spend time on providing descriptive exceptions for various flavors of corrupt data? As things stand right now, pyelftools will exception on a file with a string offset leading to nowhere (`decode called on NoneType` will be the exception); with this patch, the exception will be more descriptive. On one hand, that's a user friendly thing to do; on the other hand, given the myriad ways an extremely interlinked data structure can be corrupt, should we endeavor to catalogue them all?

Issue #535, opened 2024-01-24
Crash when iterating over notes by iterating over sections when a section is 8-byte aligned and padded


haadr:
All below is tested with pyelftools Version: 0.30.

### Context

As in https://github.com/eliben/pyelftools/issues/534, I've got a `libqtwebenginequickplugin.so` where qt has injected a `.note.qt.metadata` section and note. 

```
# readelf --sections libqtwebenginequickplugin.so
There are 31 section headers, starting at offset 0x41f8:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .note.gnu.pr[...] NOTE             00000000000002a8  000002a8
       0000000000000030  0000000000000000   A       0     0     8
  [ 2] .note.qt.metadata NOTE             00000000000002d8  000002d8
       0000000000000078  0000000000000000   A       0     0     8
...
```

As we can see, the note is of size 120.

### Bug ❓

When trying to iterate over the notes I get

```
Traceback (most recent call last):
  File "pyelftools/elftools/construct/core.py", line 573, in _parse
    subobj = self.subcon._parse(stream, context)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyelftools/elftools/construct/core.py", line 316, in _parse
    return _read_stream(stream, self.length)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyelftools/elftools/construct/core.py", line 293, in _read_stream
    raise FieldError("expected %d, found %d" % (length, len(data)))
elftools.construct.core.FieldError: expected 1, found 0
```

Using pyelftools to inspect the note before the crash, I see

```
Note section ".note.qt.metadata" at offset 728 with size 120
    Name: qt-project!
    Type: 1951465473
    Namesz: 12
    Descsz: 89
    Desc (89): 01060601bf02782c6f72672e71742d70726f6a6563742e51742e51516d6c457874656e73696f6e496e746572666163652f312e3003715174576562456e67696e65506c7567696e63757269816b5174576562456e67696e65ff
```

It looks like the section (and desc?) is padded with a total of 7 bytes.

```
hexdump -C .note.qt.metadata 
[...]

PR #537, opened 2024-01-29
Fix iter_notes to handle >4 byte segment alignment

martijnthe:
An ELF note's name and data are always 4-byte aligned. However, when the containing segment has a larger alignment requirement, additional padding gets added after the last note. This was not being handled correctly by the `iter_notes` method. This PR fixes the problem by ensuring there is enough data left to at least parse a note header.

I did not find another way to determine the "real size" of a note segment.

Fixes https://github.com/eliben/pyelftools/issues/535


eliben:
@sevaa

sevaa:
Again, no objections, the code looks sensible, the business case is alien to me, but who knows what do linkers out there emit.

eliben (approved):
Thanks!

martijnthe:
Thanks @eliben for merging my PRs. Could you publish a new release to pypi? Thank you 🙏

Issue #534, opened 2024-01-24
Unable to handle notes following a NT_GNU_PROPERTY_TYPE_0 note in the same segment


haadr:
Hia!

All below is tested with pyelftools Version: 0.30.

**NOTE** For me to be able to reproduce the below behavior, I had to first address https://github.com/eliben/pyelftools/issues/535.

### Context

I've hit an issue where [qt6 has started adding note entries to elf segments and sections](https://code.qt.io/cgit/qt/qtbase.git/commit/?id=8c2969ea86c4b38cbece5c1c2295e44dd5aa4f71) to mark an elf file as a qt plugin.

The new note, `.note.qt.metadata` is referenced both as part of a segment and as its own section. The new note seems to be appended to a segment already containing a NT_GNU_PROPERTY_TYPE_0 

Example:

```
# readelf --segments libqtwebenginequickplugin.so
Elf file type is DYN (Shared object file)
Entry point 0x0
There are 11 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000001658 0x0000000000001658  R      0x1000
  LOAD           0x0000000000002000 0x0000000000002000 0x0000000000002000
                 0x00000000000006f5 0x00000000000006f5  R E    0x1000
  LOAD           0x0000000000003000 0x0000000000003000 0x0000000000003000
                 0x0000000000000398 0x0000000000000398  R      0x1000
  LOAD           0x0000000000003b70 0x0000000000004b70 0x0000000000004b70
                 0x0000000000000530 0x0000000000000558  RW     0x1000
  DYNAMIC        0x0000000000003cb0 0x0000000000004cb0 0x0000000000004cb0
                 0x0000000000000260 0x0000000000000260  RW     0x8
  NOTE           0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x00000000000000a8 0x00000000000000a8  R      0x8
  NOTE           0x0000000000000350 0x0000000000000350 0x0000000000000350
                 0x0000000000000024 0x0000000000000024  R      0x4
  GNU_PROPERTY   0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
                 0x0000000000000030 0x0000000000000030  R      0x8
[...]

PR #538, opened 2024-01-29
Fix buffer overrun in NT_GNU_PROPERTY_TYPE_0 parser

martijnthe:
The `iter_notes` method has code to parse `NT_GNU_PROPERTY_TYPE_0` type notes. The contents of the note are interpreted as an array of `elffile.structs.Elf_Prop`s.

There was a bug where it would keep on parsing from the stream until the end of the *segment or section*. This is only correct if the note would be the last in the segment/section. In general, it should stop parsing until it reaches the end of the note's data buffer.

This PR fixes this bug.

Fixes: https://github.com/eliben/pyelftools/issues/534


eliben:
@sevaa

sevaa:
No objections. Compilers are a diverse bunch...

elftools/elf/notes.py
```diff
@@ -49,7 +49,8 @@ def iter_notes(elffile, offset, size):
         elif note['n_type'] == 'NT_GNU_PROPERTY_TYPE_0':
             off = offset
             props = []
-            while off < end:
+            current_note_end = offset + note['n_descsz']
```
eliben, line 52: Can you add a small comment here explaining the semantics (the usage of `n_descsz`)? A link to some spec-like thing would be great too, if possible

eliben (changes requested):
Thanks for the fix and the thorough testing added. Just one small request

martijnthe:
@eliben I've addressed your comment and resolved the conflicts with the other PR.

eliben (approved):

PR #540, opened 2024-02-12
Skip the cache of DWARFInfo and CU.

ThinkerYzu:
Add DWARFInfo.skip_cache() and DWARFInfo.enable_cache() to give users the ability of controlling cache.

For the case of parsing the DWARF of a large binary, we may want to skip the cache to release the memory ASAP, avoiding extra CPU cycles on maintaining a cache.

One of my use cases is to extract types, functions, and call sites from the DWARF of a Linux kernel image. With caches, it takes about 573 seconds to go through all DIEs. Skipping caches reduces time to 448 seconds. It is about 27% faster. When going through every DIEs sequentially, cache doesn't help use at all.

### DIFF
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -226,6 +226,9 @@ class CompileUnit(object):
         # the top DIE and obtain a reference to its stream.
         top_die_stream = self.get_top_DIE().stream
 
+        if self.dwarfinfo._skip_cache:
+            return DIE(cu=self, stream=top_die_stream, offset=offset)
+
         # `offset` is the offset in the stream of the DIE we want to return.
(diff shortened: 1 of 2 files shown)


eliben (changes requested):
This feels like a very targeted patch for a specific problem. If I were to accept it, I'd like some more systematic thinking about the approach. The switch should be positive (enable cache, not skip cache), there should be much more comments and there should be tests.

eliben:
> > The switch should be positive (enable cache, not skip cache), there should be much more comments and there should be tests.
> 
> Do you mean that caches should be disabled by default?

No.

It can be something like "enable cache" which is true by default but can be set to false if needed

Issue #370, opened 2021-08-30
Bug: DOS when iterating over sections


pd-fkie:
Applications using pyelftools are susceptible to DOS when parsing an ELF file whose
1. `e_shentsize` is zero
2. `e_shnum` is larger than the actual number of entries in the section header table

The function `ELFFile.iter_sections` iterates over `i in range(self['e_shnum'])` and yields sections
at file offset `self['e_shoff'] + i * self['e_shentsize']`. 
If `e_shentsize` is zero this will return the first section over and over again.
Thus a malicious ELF file can make pyelftools return the first section 9223372036854775808 times
and cause a DOS.

My question here is: Can you make pyelftools throw an exception if `e_shentsize` is zero because that would get rid of the DOS?
An `e_shentsize` of zero doesn't make sense anyway.

An exemplary ELF file that can trigger this can be found here: [timeout-3561f873d79b09ce755b0981d3df30b9ac0ca0d3](https://github.com/pd-fkie/crashes/raw/main/timeout-3561f873d79b09ce755b0981d3df30b9ac0ca0d3)

eliben:
Thanks for the report. PRs to fix this are welcome.

eliben:
To be clear, `pyelftools` was never designed to be used in security-critical settings. For example, it's not particularly fast (being pure Python and all), and it can be trivially "DOS"-ed by passing a file with large DWARF data.

I'll gladly accept a PR that fixes this specific issue, but this isn't a high priority for me in any way.

sevaa:
It's not a DoS if the only computer you've sabotaged is yours.

pyelftools doesn't listen on a network - the DoS can't be triggered by a remote attacker. If the attacker already has the ability to execute arbitrary Python on your machine - you've got bigger problems. 

If a third party piece accepts binaries from untrusted callers and passes them to pyelftools uncritically - that's on the said piece.

For a more extended version of the same argument, see this:
https://devblogs.microsoft.com/oldnewthing/20220627-00/?p=106792

sevaa:
@pd-fkie Addressed in #522. Also the same for `e_phentsize`.

Issue #369, opened 2021-08-30
Bug: Unlimited recursion when parsing sections


pd-fkie:
When the class `ELFFile` parses sections that are linked via `sh_link` it doesn't check for any loops
in the linked list.
According to [Oracles Linker and Libraries Guide](https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-94076.html) a linked list of sections can be terminated by
a section that points to itself if `sh_flags` contains the flag `SHF_ORDERED`.
In the current state of pyelftools however such an endless loop would result in a `RecursionError` exception.

The following ELF file can trigger the exception when parsed with the current version of pyelftools:
[crash-d08fc9a817c40b3c241a163d9101ab425fe31db5](https://github.com/pd-fkie/crashes/raw/main/crash-d08fc9a817c40b3c241a163d9101ab425fe31db5)
As the name indicates this bug was found by fuzzing and thus the crash file does not adhere to the ELF standard however it is possible to create a valid ELF file that triggers the same exception.

eliben:
Thanks for the report. PRs to fix this are welcome.

sevaa:
To elaborate, the problem in the linked file is that section 0 has type SHT_HASH and has 0 as sh_link (e. g. points at itself). `ELFFile.get_section()` for section 0 tries to build a `ELFHashSection`, follows the link, which calls `get_section()` with 0 and so on.

sevaa:
@pd-fkie Addressed in #522 by validating which section types can link to which.

PR #552, opened 2024-04-16
Exceptions on malformed ELF header contents

sevaa:
Addresses #369, #370. Instead of infinite loops, the library will raise exceptions.

I don't know what was the original motivation for making e_shentsize and e_phentsize variable; assuming it was for some kind of vendor extensions, but if the size is *less* than the mandated fields, ELFFile will now throw.


elftools/elf/elffile.py
```diff
@@ -588,12 +608,18 @@ def _identify_file(self):
     def _section_offset(self, n):
         """ Compute the offset of section #n in the file
         """
-        return self['e_shoff'] + n * self['e_shentsize']
+        shentsize = self['e_shentsize']
+        if self['e_shoff'] > 0 and shentsize < (0x40 if self.elfclass == 64 else 0x28):
```
eliben, line 612: can we derive these values (0x40, 0x28, 0x38, 0x20) from something instead of being magic constants?

sevaa: The constants come straight from the spec, but I might be able can get them as parser sizeofs.

sevaa:
No more magic constants.

eliben (approved):

PR #553, opened 2024-04-18
24 bit integer support, forms strx3, addrx3

sevaa:
Forms strx3, addrx3 were both TODOs.

I don't have a binary with either of those forms, but I have a crash report along the lines of KeyError: DW_FORM_strx3, so it exists in the wild.

There are currently no forms that correspond to *signed* 24 bit ints.

This functionality, I believe, was the original reason for the attempted migration to construct 2.10.

---

While at it, I've noticed some gaps in the the enums. I thought we concluded that we don't need a sample binary for every single constant. The LLVM vendor extensions for LNCT come from https://llvm.org/docs/AMDGPUDwarfExtensionsForHeterogeneousDebugging.html .

### DIFF
--- a/elftools/common/construct_utils.py
+++ b/elftools/common/construct_utils.py
@@ -6,9 +6,10 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from struct import Struct
 from ..construct import (
     Subconstruct, ConstructError, ArrayError, Adapter, Field, RepeatUntil,
-    Rename, SizeofError, Construct
+    Rename, SizeofError, Construct, StaticField
     )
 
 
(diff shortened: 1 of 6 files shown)


elftools/common/construct_utils.py
```diff
@@ -110,3 +111,34 @@ def _build(self, obj, stream, context):
...
+_UBInt24_packer = Struct(">BH")
+_ULInt24_packer = Struct("<HB")
+
+class UBInt24(StaticField):
+    """unsigned, big endian 24-bit integer"""
+    def __init__(self, name):
+        StaticField.__init__(self, name, 3)
+
+    def _parse(self, stream, context):
+        global _UBInt24_packer
```
eliben, line 124: I'm surprised you need this `global` here and elsewhere

sevaa: Technically not needed, more of a hint to self and other maintainers. Might work as a microoptimization, too. Should I remove them?

sevaa:
Removed globals. There might be a teeny tiny hit on performance.

sevaa:
Speaking of possible performance enhancements. We have a bunch of statements here and there along the lines of:
```
x = struct_parse(structs.Dwarf_uleb128(''), stream)
```
where there is an unnecessary overhead of constructing a new instance of the stateless ULEB128 parser. We could wring some performance out of constructing one early on (maybe even globally :) ) and reusing it throughout. What do you think, and if so, what should we call it?

eliben:
> Speaking of possible performance enhancements. We have a bunch of statements here and there along the lines of:
> 
> ```
> x = struct_parse(structs.Dwarf_uleb128(''), stream)
> ```
> 
> where there is an unnecessary overhead of constructing a new instance of the stateless ULEB128 parser. We could wring some performance out of constructing one early on (maybe even globally :) ) and reusing it throughout. What do you think, and if so, what should we call it?

We should prove it actually matters for performance first :)

eliben (approved):

sevaa:
> We should prove it actually matters for performance first 

That's on my TODO list also :)

Issue #481, opened 2023-07-21
Open ELF's from Memory


rpm5099:
Could you add a change for `elftools.elf.elffile.ELFFile` to accept `bytes`, `bytearray`, and `memoryview` types in addition to filename/file handle?  Any elf already in memory requires either a [useless] class to make bytes look like an IO object, be converted into a BytesIO object (which duplicates data in memory), or be unnecessarily written to disk.  Thanks.

sevaa:
Perversely, if we ever get to address #479, this will provide a pathway to this one.

That said - how bad is duplication in memory, exactly?

rpm5099:
I have not done actual measurements, but at a minimum converting `memoryview` to `bytes` duplicates the contents in memory of whatever is in the `memoryview` plus I'm sure some python overheard associated with the `bytes` class.  As an alternative it is possible to wrap `memoryview`  or `bytes` in a class that makes it behave like a file handle so that you don't have to modify the other routines that accept it (i.e. so it has `read`, `tell`, `seek` etc. methods that do the equivalent).  If things are modified to accept `bytes`, then simply allowing `memoryview` in any type checking and treating it the same as `bytes` *usually* works fine.  

This makes it much more convenient if passing data in memory, such as a file downloaded or extracted within python .  This isn't going to help with #479, all it would do is prevent duplication of the original 914MB elf in memory before things got started.

sevaa:
The docs state that `BytesIO` accepts any object with `Buffer` protocol - `memoryview` included. Have you tried constructing a `BytesIO` around a `memoryview` and feeding it to the `ELFFile` constructor? It should work.

However, once you get to DWARF parsing (and you are after that, are you?), the first thing pyelftools does is reads all debug related sections into `bytes` objects. So that is duplication, if only of the DWARF related sections. You can work around that too, if you construct a `BytesIO` around a `memoryview` *for each section* and call the `DWARFInfo` constructor directly. I do that for non-ELF files in DWEX - see [here](https://github.com/sevaa/dwex/blob/master/dwex/formats.py), in the Mach-O and PE handling code. Except I construct my `BytesIO` objects around `bytes`.

For some guidance how to pull DWARF sections from ELF, see `get_dwarf_info()` in pyelftools' elftools/elf/elffile.py. Supporting compressed sections will be a problem (gzipped data are not `seek`able), but if your binaries don't have them, you can ignore that scenario.

rpm5099:
[...]

PR #557, opened 2024-04-22
Performance enhancements

sevaa:
I made the following enhancements, in the rough order of effect:

 - Rewritten the LEB128 parsers to be less idiomatic but more streamlined
 - Instantiated the handful of parsers for scalars that are used in all the one off `struct_parse` calls, and replaced all points of usage
 - Inlined some stuff on the hot path
 - Removed unneeded `stream_preserve` for streams in auxiliary sections

### DIFF
--- a/elftools/common/construct_utils.py
+++ b/elftools/common/construct_utils.py
@@ -9,7 +9,7 @@
 from struct import Struct
 from ..construct import (
     Subconstruct, ConstructError, ArrayError, Adapter, Field, RepeatUntil,
-    Rename, SizeofError, Construct, StaticField
+    Rename, SizeofError, Construct, StaticField, FieldError
     )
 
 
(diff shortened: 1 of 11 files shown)


elftools/common/construct_utils.py
```diff
@@ -47,49 +47,43 @@ def _build(self, obj, stream, context):
...
-        return value
-
-
-class _SLEB128Adapter(Adapter):
-    """ An adapter for SLEB128, given a sequence of bytes in a sub-construct.
+        shift = 0
+        while True:
+            data = stream.read(1)
+            if len(data) != 1:
+                raise FieldError("expected 1, found 0")
```
eliben, line 62: Here and below: maybe something about premature end of stream instead? Like "stream ended in the middle of a LEB128 encoded value"?

elftools/dwarf/die.py
```diff
@@ -224,60 +225,64 @@ def _parse_DIE(self):
...
+                self.attributes[name] = AttributeValue(
+                    name=name,
+                    form=form,
+                    value=value,
+                    raw_value=raw_value,
+                    offset=attr_offset,
+                    indirection_length = indirection_length)
+
+            self.size = stream.tell() - self.offset
+        except ConstructError as e:
```
eliben, line 276: What is this for? 

The extra indentation makes it difficult to review the actual change here; can it be done separately if needed?

elftools/common/construct_utils.py
sevaa: Good point. I was going for a drop-in replacement for the existing logic, including the exception wording. Will fix.

elftools/dwarf/die.py
sevaa: Before, the "read the attribute" value line was a call to `struct_parse`. That did two things:

 - seek if needed
 - set up a try/except block and translate `FieldError` into `ELFParseException`

I've replaced that line with a call to parse_stream() on the parser object, since there is no need to seek. Avoided a needless `if()` and an extra function call on the hot path. But the logic of translating FieldError to ELFParseException needed to be kept, so I've moved it to the DIE level. There was a measurable effect.

eliben: Understood, thanks

eliben (approved):

sevaa:
@eliben 

On the topic of performance, there is a disconnect between ELF proper parsing and DWARF parsing. ELF parsing, at least in the typical use case, happens against a file stream. DWARF parsing is all in memory (barring exotic scenarios where users monkeypatch or subclass `ELFFile`), and yet we still treat DWARF sections as streams instead of byte arrays in memory that they really are. Rather than reading (and constructing bytes objects all the time), the parser could have worked with `bytes` elements and slices. The workhorse method of parsing primitive datatypes - `structs..unpack()` - can take a buffer and an offset.

But that would pretty much mean abandoning `construct` for the DWARF part of it. As far as I understand, `construct` was meant to work like `protobuf`, working with wire and file streams.

eliben:
The mmap idea sounds intriguing

sevaa:
Basically already done this in #481 :)

Issue #558, opened 2024-04-27
SHT_AMD64_UNWIND missed in _DESCR_SH_TYPE


vmurashev:
Please check elftools/elf/descriptions.py
key SHT_AMD64_UNWIND is missed in _DESCR_SH_TYPE

eliben:
#559 fixes these, I believe.

PR #559, opened 2024-04-27
Missed some CPU specific section types

sevaa:
Addresses #558. Did ARM64 while at it, we were missing it.
Wording in descriptions checked against the `binutils` sources.


eliben (approved):

eliben:
Thanks!

Issue #562, opened 2024-07-03
Callframe information parsing issues


Maxicu5:
For the ELF files (generated by IAR) where the referenced CIE may go after the FDE. CIE is loaded in the scope of FDE and placed into a cache, then when the queue comes to that CIE it takes it from the cache and doesn't move a stream offset, so this function stuck:
```
    def _parse_entries(self):
        entries = []
        offset = 0
        while offset < self.size:
            entries.append(self._parse_entry_at(offset))
            offset = self.stream.tell()
        return entries

```
locally (temporary) fixed like this (it works):
```
  def _parse_entries(self):
    entries = []
    offset = 0
    while offset < self.size:
      e = self._parse_entry_at(offset)
      entries.append(e)
      offset = e.offset + e.header.length +e.structs.initial_length_field_size()
    return entries
```
Also, CIE of version 4 is parsed by the structs of version 2, but there are two additional fields, so after the header everything is shifted, I had to add the fix in two places, seems it works:
```
        if self.for_eh_frame:
            is_CIE = CIE_id == 0
        else:
            is_CIE = (
                (dwarf_format == 32 and CIE_id == 0xFFFFFFFF) or
                CIE_id == 0xFFFFFFFFFFFFFFFF)

        # Parse the header, which goes up to and excluding the sequence of
        # instructions.
        if is_CIE:
>           version = struct_parse(entry_structs.Dwarf_uint8(''), self.stream)
>           entry_structs = DWARFStructs(
>               little_endian=self.base_structs.little_endian,
>               dwarf_format=dwarf_format,
>               address_size=self.base_structs.address_size,
>                dwarf_version=version)
            header_struct = (entry_structs.EH_CIE_header
                             if self.for_eh_frame else
                             entry_structs.Dwarf_CIE_header)
            header = struct_parse(
                header_struct, self.stream, offset)
        else:
            header = self._parse_fde_header(entry_structs, offset)


        # If this is DWARF version 4 or later, we can have a more precise
        # address size, read from the CIE header.
        if not self.for_eh_frame and entry_structs.dwarf_version >= 4:
            entry_structs = DWARFStructs(
                little_endian=entry_structs.little_endian,
                dwarf_format=entry_structs.dwarf_format,
                address_size=header.address_size,
>                dwarf_version=entry_structs.dwarf_version)
```
P.S. Very good library. Looking forward removing my workarounds and use fixed version soon.

sevaa:
[...]

PR #563, opened 2024-07-13
CIEv4 and FDE ahead of its CIE

sevaa:
Addresses both issues in #562. As always, uncovered some unrelated stuff once the binary hit the test corpus.

The FDE ahead of CIE issue is not handled correctly neither by GNU readelf nor by llvm-dwarfdump nor by eu-readelf.

Along the way, I had to file two bugs against GNU readelf:
 * https://sourceware.org/bugzilla/show_bug.cgi?id=31973
 * https://sourceware.org/bugzilla/show_bug.cgi?id=31975

Also, the aranges section in the binary looks straight up malformed - there is no entry tuple alignment to the tuple size, as demanded by the standard. Readelf produces nonsensical output on that section, pyelftools conks out. Smells like a bug in IAR.

Over those issues, I had to exclude the new binary from some of the readelf tests. I'll try to patch readelf in the meantime.


eliben (approved):
Thanks!

PR #566, opened 2024-08-01
Fix crash in readelf.py if a symbol doesn't define st_shndx

berolinux:
symbol['st_shndx'] can take a value of 'SHN_UNDEF' -- causing a crash in readelf.py when dumping the symbol table since that non-numeric value can't be compared to self.elffile.num_sections().

Add a check before assuming st_shndx is numeric.


eliben (commented):
I don't love the type cast here; can we make the check more semantic with an explanation of what other values can be encountered?

@sevaa FYI

sevaa:
I'd do a plain comparison to the string literal 'SHN_UNDEF'. But no pressure :)

sevaa:
@eliben this was addressed by #574

PR #574, opened 2024-10-23
Check for SHN_UNDEF during symbol dumping

sevaa:
Supersedes #566

Now, I don't know what was the original motivation for that one, but this fix should be equivalent.

### DIFF
--- a/scripts/readelf.py
+++ b/scripts/readelf.py
@@ -487,6 +487,7 @@ class ReadElf(object):
                 symbol_name = symbol.name
                 # Print section names for STT_SECTION symbols as readelf does
                 if (symbol['st_info']['type'] == 'STT_SECTION'
+                    and symbol['st_shndx'] != 'SHN_UNDEF'
                     and symbol['st_shndx'] < self.elffile.num_sections()
                     and symbol['st_name'] == 0):
                     symbol_name = self.elffile.get_section(symbol['st_shndx']).name


eliben:
I like the change more but agree that the motivation remains mysterious, especially without any test case. Is this similar to what `readelf` itself is doing in a comparable code path?

sevaa:
There is a similar check in readelf when it retrieves the symbol name for `STT_SECTION` type symbols. In readelf's `print_symbol()` there is the following:

```
/* Get the symbol's name.  For section symbols without a
     specific name use the (already computed) section name.  */
  if (ELF_ST_TYPE (psym->st_info) == STT_SECTION
      && section_index_real (filedata, psym->st_shndx)
      && psym->st_name == 0)
    {
      ;
    }
  else
    {
      bool is_valid;

      is_valid = valid_symbol_name (strtab, strtab_size, psym->st_name);
      sstr = is_valid  ? strtab + psym->st_name : _("<corrupt>");
    }
```
Where `section_index_real()` checks, among other things, if `st_shndx>0`. So as far as understand, the logic is:
 - if `st_shndx` is nonzero, treat it as a section index, retrieve that section's name
 - if not, retrieve the name from the string table at `st_name`
 - if that is not available either, use `<corrupt>`

So this tells me that `st_shndx` being zero (`SHN_UNDEF`) is a condition readelf handles gracefully. So pyelftools shouldn't crash on that either :)

PR #583, opened 2024-12-05
use lru cache compileunit and we can configure cache size

anjiahao1:
In some larger ELF files, traversing the CUs may cause an OOM use lru and set a cache size, can avoid it

### DIFF
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -6,7 +6,7 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
-from bisect import bisect_right
+from functools import lru_cache
 from .die import DIE
 from ..common.utils import dwarf_assert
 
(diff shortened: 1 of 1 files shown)


sevaa:
@eliben

Won't be the first time users are calling for a low memory mode for handling DWARF in extra large binaries.

I think the low memory mode should be explicitly opt in - I don't like to impose size thresholds on other people, let the library consumers do it. The large-file flag will be off by default, would go as an optional parameter into `ELFFile.get_dwarf_info()` and into the `DWARFInfo()` constructor, and will be kept there for all the parsing logic to see.

Eliminating the mandatory caching of CUs and DIEs would be rather trivial (but might break some API, or make them crazily slow - I'll have to try and see). But we still have the problem of slurping whole sections into memory. We kind of discussed it in #557, and settled that memory mapping is the way to go, however, how would you prefer to deal with transforms? If there is compression and/or relocation in DWARF sections, would you rather:

 - throw an exception that the binary is not compatible with the low memory mode
 - do the transforms anyway, even if it will eat a bunch of memory
 - let the user choose by either: 
     - introducing another optional, false by default boolean flag
     - making the large-file flag a three way enum

Also, how would you like to test the large-file mode?

eliben:
@sevaa how about moving this discussion to a separate issue to collect it all in one place

sevaa:
#586

anjiahao1:
ok, thanks @sevaa @eliben

PR #584, opened 2024-12-05
Fix for parsing errors encountered when pyelftools tries to parse non …

gettomatillos:
…GNU notes that conflicting with known GNU note types.


sevaa:
Makes sense. Our notes parsing code has always been corpus driven :)

eliben (approved):

PR #587, opened 2024-12-17
[RISCV] Add GNU note property and CFI support

jerryzj:
- Based on https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/417

PR #589, opened 2024-12-23
[RISCV] add test for GNU note property and CFI support

jerryzj:
- Stack on #587 

Signed-off-by: Jerry Zhang Jian <jerry.zhangjian@sifive.com>


sevaa:
My only concern is the test strategy. A test against a hand constructed notes section is not indicative of what the binaries in the wild, where the notes section was built by a compiler, would look like.

jerryzj:
Well... I am following the implementation in https://github.com/eliben/pyelftools/blob/main/test/testfiles_for_readelf/note_gnu_property.S

Maybe we can have a C code test case after the entire CFI support being implemented on the toolchain side.

sevaa:
That was before my time :) Anyway, if you can get a C[++] compiler to emit the right notes section, I think that'd be better. This PR will be sitting for a while anyway.

sevaa:
As of binutils 2.44, no support for RISCV notes.

jerryzj:
Yes, and I know that my colleague will submit that to upstream.

Issue #1, opened 2013-06-09
Support for shared lineprograms


eliben:
Moving from bitbucket issue #22:

Attached is an ELF file for which readelf from pyelftools behaves differently than GNU readelf.

The reason is that the file consists of multiple compilation units that share a single lineprogram. When executing readelf --debug-dump=decodedline <file>, GNU readelf prints the source lines a single time, whereas readelf from pyelftools prints the same source lines once for every compilation unit.

Should we treat lineprograms as entities that are independent of compilation units?

---

My comments:

Yeah, I can reproduce the difference... Though it's just different behavior from readelf; all information seems to be parsed correctly. I'll take a look at how to make it behave similarly.

[...]

OK, I see what the problem is.

pyelftools enumerates line programs by following each CU and looking at its line program. readelf just goes over the line program section, and extracts the CU name from the line program itself.

This shouldn't be hard to fix by adding a method to DWARFInfo that enumerates line programs in order from the section and not looking up by CU.

Patches welcome!

PR #596, opened 2025-02-12
Stripped file and gnu_debuglink support take 2

### LINKED ISSUE #594: Supplemental DWARF vs DWARF elsewhere
Back when we were discussing #525, a question was posed - what is the difference between `.gnu_debuglink`, `.gnu_debugaltlink`, and `.debug_sup`. Back at the time, no good answer was given. Now, I've stumbled upon [this](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Separate-Debug-Files.html) - a description of how the GNU toolchain has the provision to store the debug info completely elsewhere, and the file with the debug info is linked to via the `.gnu_debuglink` section.

Meanwhile, supplementary DWARF is different; in that approach, the DWARF info (CUs/DIEs and so on) is present **in two files**, and the DWARF in the primary file contains explicit references to the DWARF in the supplementary file, via forms `DW_FORM_ref_supX`, `DW_FORM_strp_sup` etc. The supplementary file is linked via the `.debug_sup` section as of DWARF v5. Judging by #426, `.gnu_debugaltlink` is the pre-standard version of `.debug_sup`.

Point is, treating all three sections as semantically identical, the way #525 does, is wrong. If there is a `.gnu_debuglink` in a file, the logic should be to pull DWARF from the other file and return it as if it was primary, instead of sticking it into the supplementary reference.

sevaa:
Closes #594, amends PR #525.

#525 treats external DWARF as a special case of supplementary DWARF, which I think is conceptually wrong. This one will provide both an interface for consumers who would prefer to resolve links by hand, and a seamless DWARF retrieval via `ELFFile.load_from_file()`.

Additionally, this PR validates the checksum on the linked DWARF.

Linking by build ID, as outlined at https://sourceware.org/gdb/current/onlinedocs/gdb.html/Separate-Debug-Files.html , is not supported.

### DIFF
--- a/elftools/dwarf/dwarf_util.py
+++ b/elftools/dwarf/dwarf_util.py
@@ -61,6 +61,13 @@ def _iter_CUs_in_section(stream, structs, parser):
         offset = header.offset_after_length + header.unit_length   
 
 def _file_crc32(file):
+    """ Provided a readable binary file stream, reads the whole file
+        and computes the CRC32 checksum of its contents,
+        with the initial value of 0.
+
+        Assumes the file pointer is at the top, and rewinds to the top
+        afterwards.
+    """
     d = file.read(4096)
     checksum = 0
     while len(d):
--- a/elftools/dwarf/dwarf_util.py
+++ b/elftools/dwarf/dwarf_util.py
@@ -7,7 +7,7 @@
 # This code is in the public domain
 #-------------------------------------------------------------------------------
 
-import os
+import os, binascii
 from ..construct.macros import UBInt32, UBInt64, ULInt32, ULInt64, Array
 from ..common.exceptions import DWARFError
 from ..common.utils import preserve_stream_pos, struct_parse
@@ -60,3 +60,11 @@ def _iter_CUs_in_section(stream, structs, parser):
         yield header
         offset = header.offset_after_length + header.unit_length   
 
+def _file_crc32(file):
+    d = file.read(4096)
+    checksum = 0
+    while len(d):
+        checksum = binascii.crc32(d, checksum)
+        d = file.read(4096)
+    file.seek(0, os.SEEK_SET)
+    return checksum
--- a/elftools/dwarf/dwarf_util.py
+++ b/elftools/dwarf/dwarf_util.py
@@ -61,17 +61,13 @@ def _iter_CUs_in_section(stream, structs, parser):
         offset = header.offset_after_length + header.unit_length   
 
 def _file_crc32(file):
-    """ Provided a readable binary file stream, reads the whole file
+    """ Provided a readable binary stream, reads the stream to the end
         and computes the CRC32 checksum of its contents,
         with the initial value of 0.
-
-        Assumes the file pointer is at the top, and rewinds to the top
-        afterwards.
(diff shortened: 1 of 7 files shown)


elftools/dwarf/dwarf_util.py
```diff
@@ -60,3 +60,11 @@ def _iter_CUs_in_section(stream, structs, parser):
         yield header
         offset = header.offset_after_length + header.unit_length   
 
+def _file_crc32(file):
```
eliben, line 63: Add brief docstring saying what the function does

elftools/elf/elffile.py
```diff
@@ -94,15 +95,25 @@ def load_from_path(cls, path):
...
+        """ Return a function that takes a potentially relative path,
+            resolves it against base_path (bytes or str), and opens a file at that.
+
+            ELFFile uses functions like that for resolving DWARF links.
+        """
+        if isinstance(base_path, str):
+            base_path = base_path.encode('UTF-8') # resolver takes a bytes path
+        base_directory = os.path.dirname(base_path)
+        def loader(rel_path):
+            # FIXME: use actual path instead of str/bytes
```
eliben, line 112: I'm not sure what this FIXME means, can you clarify?

eliben (commented):
Thanks!

elftools/elf/elffile.py
sevaa: The FIXME goes back to @rdunklau and #426. I presume it was meant to say that paths should be passed around as `pathlib.Path` objects.

sevaa:
Resolved #1

elftools/elf/elffile.py
eliben: I'll say just remove the FIXME for now. It's not helpful, IMHO. If someone wants to refactor to use `pathlib.Path`, this will have to be done in more places than just here, anyway

sevaa:
Removed the FIXME

elftools/dwarf/dwarf_util.py
```diff
@@ -60,3 +60,18 @@ def _iter_CUs_in_section(stream, structs, parser):
...
+    """ Provided a readable binary file stream, reads the whole file
+        and computes the CRC32 checksum of its contents,
+        with the initial value of 0.
+
+        Assumes the file pointer is at the top, and rewinds to the top
```
eliben, line 68: This makes me realize that the whole stream setting here is a bit magical and may not be generally useful. How about using `preserve_stream_pos` here, or just at the call site if the call site wants to preserve the position?

sevaa: The one and only usage point calls this before an `ELFFile` is constructed over the stream - which also assumes a stream at the beginning. Who knows what might other usage points (if any, if ever) require. I'll move the `seek()` to the call site.

sevaa:
Addressed.

sevaa:
@io-no: sorry, I broke your API. If you update to the latest master and/or the next release, linked DWARF retrieval is different, and the `DWARFInfo` constructor no longer has the argument for `.gnu_debuglink` anymore.

Issue #514, opened 2023-10-27
Add static typing


Timmmm:
As far as I can tell this is missing static types or type stubs.

The best way to do it is to add type hints into the code, then you have to add an empty `py.typed` file in your package (next to `__init__.py`).

I would recommend using Pyright over Mypy. It's much better. The easiest way to run/enforce it is using [the Pyright pre-commit hook](https://github.com/microsoft/pyright/blob/main/docs/ci-integration.md#running-pyright-as-a-pre-commit-hook).

JonathonReinhart:
I think this is a duplicate of #371.

pmhahn:
Yes ,#371 is a duplicate. Can you please close one in the favor of the other?
#533 is also related: If type information is included with pyelftools itself instead of in python/typeshed#11262, that one could also be closed.

I'm asking because I stumbled over missing type annotation for pyelftools myself and would like to add it to pyelftools – I did similar work for other projects in the past.

Before I start working on this I have one question: What is the oldest Python version you're willing to support? Over time may things changed, most importantly if you do `from typing import …` or `from collections.abd import …` or which features you can use, e.g. `Literal` and `ParamSpec`.
According to https://devguide.python.org/versions/ Python 3.9 is the oldest version still getting security updates. Many other projects use that as a guideline to specify their oldest supported version.

Timmmm:
@pmhahn I would recommend getting confirmation your PR would be accepted before doing any work here. I have made the mistake of contributing type annotations to Python projects before and then discovering that the authors didn't really get it.

Python 3.9 seems like a reasonable minimum bar to me.

eliben:
I've closed #371 

3.9 sounds good

I'm ok with accepting types, but it would be great if this could be done gradually

sevaa:
@eliben 

This is primarily for the public facing API, right?

eliben:
> [@eliben](https://github.com/eliben)
> 
> This is primarily for the public facing API, right?

Theoretically, wherever it'd be helpful, I suppose. But the public facing API would be a priority, of course.

pmhahn:
Yes, public API is my preferred thing, but elftools uses construct and things from there become public (by accident).
And while collection information on the types to pass into a function and returned by it, I often have to go down one layer to understand the (internal) functions first, which are called and have to figure out their signature first. When I then have this, it would be a waste to not write it down. (and while doing that, I already found some minor bugs and inconsistencies.)

[...]

PR #599, opened 2025-02-24
Remove remaining legacy Python 2 code

pmhahn:
`pyelftools` is Python 3 only, but the code still carries some Python 2 legacy code.
- `print` is a function
- instead of `OrderedDict` use `dict` which is also ordered since cPython 3.6 – it still has its uses for things like LRU-caching, but that is not the case here
- remove unneeded code when inheriting from `colklections.MutableMapping`
- remove `py3compat`layer from Construct – simplifies the code and future type hinting #514


elftools/dwarf/descriptions.py
```diff
@@ -490,7 +490,7 @@ def _import_extra(attr, die, section_offset):
             # stream.
             with preserve_stream_pos(die.stream):
                 ref_die = DIE(cu, die.stream, ref_die_offset)
-            #print '&&& ref_die', ref_die
+            #print('&&& ref_die', ref_die)
```
eliben, line 493: Let's just remove this line altogether

elftools/dwarf/die.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, cu, stream, offset):
         self.stream = stream
         self.offset = offset
 
-        self.attributes = OrderedDict()
+        self.attributes = {}
```
eliben, line 89: Less sure about these. AFAIU `OrderedDict` still has some unique features vs. `dict` even if the latter preserves order now. Are you sure none of these are used?

eliben (commented):
Thank you, very nice cleanup overall!

elftools/dwarf/die.py
pmhahn: - [x] None of the special methods unique to `OrderedDict` are used.
- [x] `elftools.dwarf.die.DIE.attributes` is mostly used for lookup, only `__repr__()` profits from keeping the order
- [x] `elftools.dwarf.dwarfinfo.DWARFInfo._type_units_by_sig` is only used for internal lookup; preserving the order look unneeded
- [x] `elftools.dward.namelut.NameLUT.entries` is the most critical change as there it is part of the public API, but internally is does not change anything. External code can still use `set_entries()` to pass in an `OrderedDict`, but might fail if it really expects `get_entries()` to return an `OrderedDict` instance.

I have checked the [documented behaviour change](https://docs.python.org/3/library/collections.html#ordereddict-objects) again: I only see problems if any code reads DWARF information, modified it and then tries to serialize it back into binary data – or you're writing your own compiler and use `pyelftools` to serialize the data. Only in those cases you would use _modifying_ operations on the `OrderedDict`, which then might show slightly different results.

In all other cases the 3 attributes are mostly used for lookup, which would benefit from the speedup of `dict` compared to `OrderedDict` – but I did not measure that.

PR #600, opened 2025-02-24
Minor fixes while adding type annotaions

pmhahn:
While working on #514 I – more correctly: `mypy` – found two minor issues:
- remove duplicate definition of `ENUM_SUNW_SYMINFO_BOUNDTO`
- make several methods as `@staticmethod` as they are used as functions.


elftools/elf/relocation.py
```diff
@@ -340,167 +340,176 @@ def _do_apply_relocation(self, stream, reloc, symtab):
     _RELOCATION_RECIPE_TYPE = namedtuple('_RELOCATION_RECIPE_TYPE',
         'bytesize has_addend calc_func')
 
-    def _reloc_calc_identity(value, sym_value, offset, addend=0):
```
eliben, line 343: Interesting that this is a problem; is mypy complaining that these methods don't take a `self`?

For standalone functions, we can just make them standalone (move outside the calss) instead of adding an internal class, no?

eliben (changes requested):

elftools/elf/relocation.py
pmhahn: Yes, `mypy` is complaining about the missing `self` argument.

The strange thing was, that just adding `@staticmethod` was okay with my local Python 3.11 testing, but GitHub actions failed for Python 3.9. and complained about missing `staticmethod` then. Therefore I choose to move them into a separate sub-class as _class defining_ is strange and seems to have changed over different Python versions.

Yes, moving them completely outside of `class RelocationHandler` would be a valid alternative, which I can do if you prefer that.

eliben: Yes, let's just make them standalone functions.  Thanks!
Looking at their usage as higher-order functions, this seems more natural anyway.

pmhahn: Changed, rebased and force-pushed.

eliben (approved):

PR #602, opened 2025-02-27
Fixed found while adding type hints

pmhahn:
While working on #514 `mypy` found several issues, which I fixed myself.

1. You might want to handle this differently (in the future):
> AssertionError: Unhandled augmentation string: b'armcc+'
2. `arm-eabi-attr-names.o.elf` is based on `attr-names.s` from [binutils-gdb](https://sourceware.org/git?p=binutils-gdb.git;a=blob;f=gas/testsuite/gas/arm/attr-names.s;h=c43fb88f67b7be158cceb13cb84caddaa7cd47b8;hb=HEAD); does that need additional attribution?

Please check carefully as I might have introduced other issues.

### DIFF
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -137,7 +137,7 @@ class RelrRelocationTable(object):
 
         # If DT_RELRSZ is zero, offset is meaningless and could be None.
         if self._size == 0:
-            return []
+            return
 
         limit = self._offset + self._size
         relr = self._offset


scripts/readelf.py
```diff
@@ -815,15 +815,14 @@ def display_hex_dump(self, section_spec):
             sys.stderr.write('readelf: Warning: Section \'%s\' was not dumped because it does not exist!\n' % (
                 section_spec))
             return
-        if section['sh_type'] == 'SHT_NOBITS':
-            self._emitline("\nSection '%s' has no data to dump." % (
+        if section['sh_type'] == 'SHT_NOBITS' or not (data := section.data()):
```
eliben, line 818: Maybe I'm missing something, but what's the point of `data := ...` here? You're not using `data` inside the condition body

elftools/elf/descriptions.py
```diff
@@ -237,30 +237,30 @@ def describe_note(x, machine):
 
 
 def describe_attr_tag_arm(tag, val, extra):
+    s = _DESCR_ATTR_TAG_ARM.get(tag, '"%s"' % tag)
```
eliben, line 240: Why do we have this? Unknown tags not defined in `_DESCR_ATTR_TAG_ARM`?

Also, what's the point of the string interpolation? Isn't `tag` already a string?

eliben:
If you add a new binary test file, try to add a source file alongside it (.s or .c etc.) with a comment on top explaining where it's from and how to convert it to the binary

scripts/readelf.py
pmhahn: I moved the `data = section.data()` here from 3 lines below: Otherwise I would have to either repeat the output:
```python
if section['sh_type'] == 'SHT_NOBITS':
    self._emitline("\nSection '%s' has no data to dump." % …)
data = section.data()
if not data:
    self._emitline("\nSection '%s' has no data to dump." % …)
```
or *always* call `section.data()` even in the case where `sh_type = SHT_NOBITS`:
```python
data = section.data()
if section['sh_type'] == 'SHT_NOBITS' or not data:
    self._emitline("\nSection '%s' has no data to dump." % …)
```
By using the walrus operator I get both: no repetition and the lazy calling. YMMV

pmhahn:
> If you add a new binary test file, try to add a source file alongside it (.s or .c etc.) with a comment on top explaining where it's from and how to convert it to the binary

I have added the source file with instructions on how to compile it.

eliben (approved):
Thanks for the clarifications
LGTM

PR #607, opened 2025-03-13
Modernize python code

pmhahn:
While working on #514 I took the opportunity to modernize some old code:
- ~~Use named arguments instead of `kwargs` to allow typing them~~
- ~~Simplify `pyelftools.construct.core.Range`: `mypy` complained about `pos` not being declared in all cases; rewrite the logic.~~
- Some variables are `const`, except while declaring them.
  - ~~Build _printable with a dict-comprehension: Basically this is~~
  - Build _OPCODE_NAME_MAP with a dict-comprehension
- Some more `ruff` findings
  - Optimize check for empty containers [PLC1802](https://docs.astral.sh/ruff/rules/len-test/) et.al.
  - Simplify sequence concatenation [RUF005](https://docs.astral.sh/ruff/rules/collection-literal-concatenation/)
  - Convert to list comprehensions [PERF401](https://docs.astral.sh/ruff/rules/manual-list-comprehension/)
  - Convert yield from for-loop [UP028](https://docs.astral.sh/ruff/rules/yield-in-for-loop/)

:information_source:  There is a small change in the output of `elftools.dwarf.lineprogram.LineState.__repr__()`, where I removed the trailing `\n`. If you want it back append another`""` to the end of the tuple.

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -520,7 +520,7 @@ class CFIEntry(object):
             # line that serves as the base (first) line in the FDE's table.
             cie = self.cie
             cie_decoded_table = cie.get_decoded()
-            if len(cie_decoded_table.table) > 0:
+            if cie_decoded_table.table:
                 last_line_in_CIE = copy.copy(cie_decoded_table.table[-1])
                 cur_line = copy.copy(last_line_in_CIE)
             else:
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -725,7 +725,8 @@ _PRIMARY_ARG_MASK = 0b00111111
 # This dictionary is filled by automatically scanning the constants module
 # for DW_CFA_* instructions, and mapping their values to names. Since all
 # names were imported from constants with `import *`, we look in globals()
-_OPCODE_NAME_MAP = {}
-for name in list(globals().keys()):
-    if name.startswith('DW_CFA'):
-        _OPCODE_NAME_MAP[globals()[name]] = name
+_OPCODE_NAME_MAP = {
+    value: name
+    for name, value in globals().items()
+    if name.startswith('DW_CFA')
+}
--- a/elftools/dwarf/compileunit.py
+++ b/elftools/dwarf/compileunit.py
@@ -205,8 +205,7 @@ class CompileUnit(object):
         yield die
         if die.has_children:
             for c in die.iter_children():
-                for d in die.cu._iter_DIE_subtree(c):
-                    yield d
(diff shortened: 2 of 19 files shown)


elftools/construct/core.py
```diff
@@ -493,23 +493,17 @@ def __init__(self, mincount, maxcout, subcon):
...
-                while c < self.maxcout:
-                    pos = stream.tell()
-                    obj.append(self.subcon._parse(stream, context))
-                    c += 1
-        except ConstructError as ex:
-            if c < self.mincount:
-                raise RangeError("expected %d to %d, found %d" %
-                    (self.mincount, self.maxcout, c), ex)
-            stream.seek(pos)
+        ctx = context.__copy__ if self.subcon.conflags & self.FLAG_COPY_CONTEXT else (lambda: context)
```
eliben, line 496: If at all possible, I would prefer to minimize changes inside `elftools/construct`

Is there a simpler way to make the type checker happy? Just predeclaring pos?

elftools/construct/lib/binary.py
```diff
@@ -15,13 +15,7 @@ def int_to_bin(number, width=32):
     """
     if number < 0:
         number += 1 << width
-    i = width - 1
```
eliben, line 18: Why is this change required?

elftools/construct/lib/binary.py
```diff
@@ -57,24 +51,11 @@ def swap_bytes(bits, bytesize=8):
         >>> swap_bytes(b'00011011', 2)
         b'11100100'
     """
-    i = 0
```
eliben, line 60: same here

elftools/construct/core.py
pmhahn: Minimum would be an additional `pos = -1` before the `try`, e.g.
```diff
--- a/elftools/construct/core.py
+++ b/elftools/construct/core.py
@@ -496,1 +496,1 @@ class Range(Subconstruct):
-        c = 0
+        c = pos = -1
```

elftools/construct/lib/binary.py
pmhahn: Same here, moved to #609

elftools/construct/lib/binary.py
pmhahn: Not required, but adding type hints requires reading and understanding the code. My brain has the (bad) habit to rewrite the code to be more pregnant. Moved to #609

elftools/construct/core.py
pmhahn: Now also in #609

eliben (approved):

PR #609, opened 2025-03-15
Modernize python code for `construct`

pmhahn:
While working on #514 I took the opportunity to modernize also some old code in #607. But there are several changed related to the internal copy of `construct`. That copy shall be updated by #548, but until then the following changes might be okay:

- Fix the doctest
- Use named arguments instead of kwargs to allow typing them
- Simplify `pyelftools.construct.core.Range`: `mypy` complained about `pos` not being declared in all cases; rewrite the logic.
- Some variables are `const`, except while declaring them.
    - Build `_printable` with a dict-comprehension
    - Same for `char2bin` and `bin2char`
- Rewrite some functions to use generater expressions

### DIFF
--- a/elftools/construct/core.py
+++ b/elftools/construct/core.py
@@ -4,6 +4,18 @@ from struct import Struct as Packer
 from .lib import Container, ListContainer, LazyContainer
 
 
+__all__ = [
+    "ConstructError", "FieldError", "SizeofError", "AdaptationError", "ArrayError", "RangeError", "SwitchError", "SelectError", "TerminatorError",
+    "Construct", "Subconstruct", "Adapter",
+    "StaticField", "FormatField", "MetaField", "MetaArray", "Range", "RepeatUntil",
+    "Struct", "Sequence", "Union",
+    "Switch", "Select",
+    "Pointer", "Peek", "OnDemand", "Buffered", "Restream",
+    "Reconfig", "Anchor", "Value", "LazyBound",
+    "_Pass", "Pass", "_Terminator", "Terminator",
+]
+
+
 #===============================================================================
 # exceptions
 #===============================================================================
--- a/elftools/construct/core.py
+++ b/elftools/construct/core.py
@@ -366,14 +366,15 @@ class MetaField(Construct):
     :param callable lengthfunc: callable that takes a context and returns
                                 length as an int
 
+    >>> from ..construct import Byte
     >>> foo = Struct("foo",
     ...     Byte("length"),
     ...     MetaField("data", lambda ctx: ctx["length"])
     ... )
-    >>> foo.parse("\\x03ABC")
-    Container(data = 'ABC', length = 3)
-    >>> foo.parse("\\x04ABCD")
-    Container(data = 'ABCD', length = 4)
+    >>> foo.parse(b"\\x03ABC")
+    Container({'length': 3, 'data': b'ABC'})
+    >>> foo.parse(b"\\x04ABCD")
+    Container({'length': 4, 'data': b'ABCD'})
     """
 
     __slots__ = ["lengthfunc"]
@@ -459,29 +460,30 @@ class Range(Subconstruct):
     :param int maxcount: the maximal count
     :param Construct subcon: the subcon to repeat
 
+    >>> from ..construct import UBInt8
     >>> c = Range(3, 7, UBInt8("foo"))
-    >>> c.parse("\\x01\\x02")
+    >>> c.parse(b"\\x01\\x02")  # doctest: +IGNORE_EXCEPTION_DETAIL
     Traceback (most recent call last):
       ...
-    construct.core.RangeError: expected 3..7, found 2
-    >>> c.parse("\\x01\\x02\\x03")
+    RangeError: expected 3..7, found 2
+    >>> c.parse(b"\\x01\\x02\\x03")
     [1, 2, 3]
-    >>> c.parse("\\x01\\x02\\x03\\x04\\x05\\x06")
+    >>> c.parse(b"\\x01\\x02\\x03\\x04\\x05\\x06")
     [1, 2, 3, 4, 5, 6]
-    >>> c.parse("\\x01\\x02\\x03\\x04\\x05\\x06\\x07")
+    >>> c.parse(b"\\x01\\x02\\x03\\x04\\x05\\x06\\x07")
     [1, 2, 3, 4, 5, 6, 7]
-    >>> c.parse("\\x01\\x02\\x03\\x04\\x05\\x06\\x07\\x08\\x09")
+    >>> c.parse(b"\\x01\\x02\\x03\\x04\\x05\\x06\\x07\\x08\\x09")
     [1, 2, 3, 4, 5, 6, 7]
-    >>> c.build([1,2])
+    >>> c.build([1,2])  # doctest: +IGNORE_EXCEPTION_DETAIL
     Traceback (most recent call last):
       ...
-    construct.core.RangeError: expected 3..7, found 2
+    RangeError: expected 3..7, found 2
     >>> c.build([1,2,3,4])
-    '\\x01\\x02\\x03\\x04'
-    >>> c.build([1,2,3,4,5,6,7,8])
+    b'\\x01\\x02\\x03\\x04'
+    >>> c.build([1,2,3,4,5,6,7,8])  # doctest: +IGNORE_EXCEPTION_DETAIL
     Traceback (most recent call last):
       ...
-    construct.core.RangeError: expected 3..7, found 8
+    RangeError: expected 3..7, found 8
     """
 
     __slots__ = ["mincount", "maxcout"]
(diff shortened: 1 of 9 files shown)


elftools/construct/lib/binary.py
```diff
@@ -15,13 +15,7 @@ def int_to_bin(number, width=32):
...
     if number < 0:
         number += 1 << width
-    i = width - 1
-    bits = bytearray(width)
-    while number and i >= 0:
-        bits[i] = number & 1
-        number >>= 1
-        i -= 1
-    return bytes(bits)
+    return bytes(bool(number & (1 << (width - 1 - bit))) for bit in range(width))
```
eliben, line 18: Same comment as previous PR: please don't make code changes like this to construct, for reasons your PR header explains - I do hope to use it as a dependency eventually

elftools/construct/core.py
```diff
@@ -751,7 +749,7 @@ class Union(Construct):
     )
     """
     __slots__ = ["parser", "builder"]
-    def __init__(self, name, master, *subcons, **kw):
+    def __init__(self, name, master, *subcons) -> None:
```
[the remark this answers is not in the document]
pmhahn: No: Moved from that commit to one other later in the series.

elftools/construct/lib/binary.py
pmhahn: I have dropped those code cleanups for now; only the typing related changes should remain.

elftools/construct/debug.py
```diff
@@ -38,9 +42,9 @@ class Probe(Construct):
     ]
     counter = 0
 
-    def __init__(self, name = None, show_stream = True,
-                 show_context = True, show_stack = True,
-                 stream_lookahead = 100):
+    def __init__(self, name: str | None = None, show_stream: bool = True,
```
eliben, line 45: OOC: is it actually required to annotate `__init__` as returning None? Will things fail if you don't?

eliben (commented):
Thanks, much better this way.

As a general comment, could you please add type checker running instructions somewhere (maybe in https://github.com/eliben/pyelftools/wiki/Hacking-guide?)

e.g. how exactly to install the type checker (version/etc.) and how to run it on the entire project here

elftools/construct/core.py
```diff
@@ -137,7 +160,7 @@ def _is_flag(self, flag):
 
         return bool(self.conflags & flag)
 
-    def __getstate__(self):
+    def __getstate__(self) -> dict[str, Any]:
```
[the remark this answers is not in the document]
pmhahn: See [The `Any` type](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type). Summary:
- use `Any` only as the last resort; basically you _give up_ and tell the type-checkers to **not** check the type – because it is to complex or too painful to type in more detail.
- On `object` you can only do `str(…)`, `repo(…)`, `hash(…)` and `… == …` and `… != …`; so `object` is okay for something simple as `def myprint(args: list[object]) -> None: print(" ".join(args))`, but not much more.
- A value with the `Any` type is dynamically typed. Basically we don't care about the concrete type here, but `object` is not appropriate: In line 369 there is this:
  `attrs["packer"] = attrs["packer"].format`
  `object` has no attribute `format`, so using `Any` here _solves_ that nicely.

As an alternative to using `Any` here you would have to use a [`TypedDict`](https://mypy.readthedocs.io/en/stable/typed_dict.html) where you explicitly name all ever existing `state`-attributes with their corresponding type. This is impractical as that list can be extended by inheriting from `StaticField`, so would never be complete.

elftools/construct/debug.py
pmhahn: Since [mypy 0.640 "New Feature: Allow Omitting Return Type for __init__"](https://mypy-lang.blogspot.com/2018/10/mypy-0640-released.html) (~2018-10) it can be omitted – as long as you have *at least one argument annotated* – a completely un-annotated function/method is considered _dynamically types_ and is **not** checked. I have not checked others type-checkers like `pytype` or `pyright` how they handle it.

For consistency I prefer to always type the return type, even for `__init__(…) -> None`

elftools/construct/lib/bitstream.py
```diff
@@ -1,34 +1,58 @@
...
+
+import io
+from typing import IO, TYPE_CHECKING
+
 from .binary import encode_bin, decode_bin
 
-class BitStreamReader:
+if TYPE_CHECKING:
+    # Py3.12+ # from collections.abc import Buffer
+    from typing import Self  # 3.11+
```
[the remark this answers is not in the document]
pmhahn: That's what I do here: Runtime dependencies are all Python 3.9+, but for typing I'm using some features from newer Python versions:
- 3.10: [`Concatenate`](https://docs.python.org/3/library/typing.html#typing.Concatenate), [`ParamSpec`](https://docs.python.org/3/library/typing.html#typing.ParamSpec)
- 3.11: [`Self`](https://docs.python.org/3/library/typing.html#typing.Self), [`Unpack`](https://docs.python.org/3/library/typing.html#typing.Unpack)
- 3.12: [`Buffer`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer)

As those `imports` are all behind the `if TYPECHECKING:`-guard, they are only tried when type checking. And yes: then you must have at least Python 3.12 as of now.

pmhahn:
> As a general comment, could you please add type checker running instructions somewhere (maybe in https://github.com/eliben/pyelftools/wiki/Hacking-guide?)
> 
> e.g. how exactly to install the type checker (version/etc.) and how to run it on the entire project here

Don't know how to edit/clone that Wiki, but something like this:

```diff
diff --git a/Hacking-guide.rest b/Hacking-guide.rest
index 93957ad3777f..c2fc727e1ef1 100644
--- a/Hacking-guide.rest
+++ b/Hacking-guide.rest
@@ -64,4 +64,19 @@ Coding conventions
 
 **pyelftools** is written in Python, following the `PEP 8 <http://www.python.org/dev/peps/pep-0008/>`_ style guide.
 
+Type checking
+-------------
 
+**pyelftools** is using and providing Type Hints, following the `PEP 484 <https://peps.python.org/pep-0484/>`_.
+Please make sure new functions and methods are properly type hinted.
+There are some known limitations as **pyelftools** is dynamic for some types, which does not work well for _static typing_.
+
+Please make sure to run [pyright](https://github.com/microsoft/pyright) and/or [mypy](https://mypy.readthedocs.io/en/stable/) on your contributions.
+For this Python 3.12 (or newer) is required:
+
+.. sourcecode:: text
+
+  > uv sync --dev --group typing --python 3.12
+  > . .venv/bin/activate
+  > pyright elftools/
+  > mypy
```

eliben:
> > As a general comment, could you please add type checker running instructions somewhere (maybe in https://github.com/eliben/pyelftools/wiki/Hacking-guide?)
> > e.g. how exactly to install the type checker (version/etc.) and how to run it on the entire project here
> 
> Don't know how to edit/clone that Wiki, but something like this:
> 
> ```diff
> diff --git a/Hacking-guide.rest b/Hacking-guide.rest
> index 93957ad3777f..c2fc727e1ef1 100644
> --- a/Hacking-guide.rest
> +++ b/Hacking-guide.rest
> @@ -64,4 +64,19 @@ Coding conventions
>  
>  **pyelftools** is written in Python, following the `PEP 8 <http://www.python.org/dev/peps/pep-0008/>`_ style guide.
>  
> +Type checking
> +-------------
>  
> +**pyelftools** is using and providing Type Hints, following the `PEP 484 <https://peps.python.org/pep-0484/>`_.
> +Please make sure new functions and methods are properly type hinted.
> +There are some known limitations as **pyelftools** is dynamic for some types, which does not work well for _static typing_.
> +
> +Please make sure to run [pyright](https://github.com/microsoft/pyright) and/or [mypy](https://mypy.readthedocs.io/en/stable/) on your contributions.
> +For this Python 3.12 (or newer) is required:
> +
> +.. sourcecode:: text
> +
> +  > uv sync --dev --group typing --python 3.12
> +  > . .venv/bin/activate
> +  > pyright elftools/
> +  > mypy
> ```

Oh, OK. I'll move the wiki to a doc file you can send PRs to (don't particularly like the GitHub wikies anyhow). We should discuss this - I don't undestand the uv usage here, for example. Why not `uvx mypy` if you use `uv` already

eliben (approved):

pmhahn:
> > Oh, OK. I'll move the wiki to a doc file you can send PRs to (don't particularly like the GitHub wikies anyhow). We should discuss this - I don't undestand the uv usage here, for example. Why not `uvx mypy` if you use `uv` already
> 
> This is done now, can you please send a PR for https://github.com/eliben/pyelftools/blob/main/doc/hacking-guide.rst ?

See https://github.com/eliben/pyelftools/pull/614


> We should discuss this - I don't undestand the uv usage here, for example. Why not `uvx mypy` if you use `uv` already

I have added alternatives there and an explanation: Basically `mypy` and `pyright` need the so-called `typed-…` companion packages in *their* VENVs, but `pyelftools` currently has no external dependencies, this is okay for now. As soon as you have them, using a globally installed `mypy` / `pyright` becomes a problem.

PR #610, opened 2025-03-17
Last typing issues

pmhahn:
While splitting my _big typing patch_ #514 into smaller chunks I found some last issues, which can go in first and are not strictly related to typing, but found from doing that:
- Fix executable file permissions
- One more conversion to dict-comprehension
- Simplify fallback to empty dict in one case
- Simplify 1 range check
- Remove `self` from helper function
- Add one missing Unicode `encode()`
- Fix 1 case, where a generator-expression was stored instead of a tuple
- Return `bytes` instead of `str` in 1 case


eliben (approved):
Thanks!

PR #611, opened 2025-03-19
Add type hints

pmhahn:
This is the mayor PR to add Python type hints #514 – without #609 this will not be complete as `elftools.construct.Container` is used in many places, which is a container for `Any`thing: retrieving values from it will be typed `Any`, which basically means _untyped_: without manually type-hinting every such use case those values do propagate further and even spill into the public API.

Because of missing 6f99ce09ddf25ce80422f388b7690e6323ccb89f  running `mypy` will find the following errors:
```
elftools/dwarf/structs.py:581: error: "Container" has no attribute "first"  [attr-defined]
elftools/dwarf/structs.py:583: error: "Container" has no attribute "first"  [attr-defined]
elftools/dwarf/structs.py:585: error: "Container" has no attribute "first"  [attr-defined]
elftools/dwarf/structs.py:587: error: "Container" has no attribute "second"  [attr-defined]
elftools/dwarf/structs.py:590: error: "Container" has no attribute "first"  [attr-defined]
elftools/dwarf/callframe.py:99: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/ranges.py:41: error: "Container" has no attribute "start_index"  [attr-defined]
elftools/dwarf/ranges.py:42: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:42: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/ranges.py:42: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/ranges.py:46: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:46: error: "Container" has no attribute "address"  [attr-defined]
elftools/dwarf/ranges.py:47: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:47: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/ranges.py:47: error: "Container" has no attribute "start_offset"  [attr-defined]
elftools/dwarf/ranges.py:47: error: "Container" has no attribute "end_offset"  [attr-defined]
elftools/dwarf/ranges.py:48: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:48: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/ranges.py:48: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/ranges.py:48: error: "Container" has no attribute "end_address"  [attr-defined]
elftools/dwarf/ranges.py:49: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:49: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/ranges.py:49: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/ranges.py:49: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/ranges.py:50: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:50: error: "Container" has no attribute "index"  [attr-defined]
elftools/dwarf/ranges.py:51: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/ranges.py:51: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/ranges.py:51: error: "Container" has no attribute "start_index"  [attr-defined]
elftools/dwarf/ranges.py:51: error: "Container" has no attribute "end_index"  [attr-defined]
elftools/dwarf/ranges.py:70: error: "Container" has no attribute "version"  [attr-defined]
elftools/dwarf/ranges.py:180: error: "Container" has no attribute "offset_table_offset"  [attr-defined]
elftools/dwarf/ranges.py:180: error: "Container" has no attribute "is64"  [attr-defined]
elftools/dwarf/ranges.py:180: error: "Container" has no attribute "offset_count"  [attr-defined]
elftools/dwarf/ranges.py:181: error: "Container" has no attribute "offset_after_length"  [attr-defined]
elftools/dwarf/ranges.py:181: error: "Container" has no attribute "unit_length"  [attr-defined]
elftools/dwarf/ranges.py:188: error: "Container" has no attribute "entry_type"  [attr-defined]
elftools/dwarf/locationlists.py:53: error: "Container" has no attribute "start_index"  [attr-defined]
elftools/dwarf/locationlists.py:54: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:54: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:54: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/locationlists.py:54: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:58: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:58: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:58: error: "Container" has no attribute "address"  [attr-defined]
elftools/dwarf/locationlists.py:59: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:59: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:59: error: "Container" has no attribute "start_offset"  [attr-defined]
elftools/dwarf/locationlists.py:59: error: "Container" has no attribute "end_offset"  [attr-defined]
elftools/dwarf/locationlists.py:59: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:60: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:60: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:60: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/locationlists.py:60: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/locationlists.py:60: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:61: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:61: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:61: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/locationlists.py:61: error: "Container" has no attribute "end_address"  [attr-defined]
elftools/dwarf/locationlists.py:61: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:62: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:62: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:62: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:63: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:63: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:63: error: "Container" has no attribute "index"  [attr-defined]
elftools/dwarf/locationlists.py:64: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:64: error: "Container" has no attribute "entry_length"  [attr-defined]
elftools/dwarf/locationlists.py:64: error: "Container" has no attribute "start_index"  [attr-defined]
elftools/dwarf/locationlists.py:64: error: "Container" has no attribute "end_index"  [attr-defined]
elftools/dwarf/locationlists.py:64: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:81: error: "Container" has no attribute "version"  [attr-defined]
elftools/dwarf/locationlists.py:222: error: "Container" has no attribute "version"  [attr-defined]
elftools/dwarf/locationlists.py:286: error: "Container" has no attribute "entry_offset"  [attr-defined]
elftools/dwarf/locationlists.py:287: error: "Container" has no attribute "entry_end_offset"  [attr-defined]
elftools/dwarf/locationlists.py:288: error: "Container" has no attribute "entry_type"  [attr-defined]
elftools/dwarf/locationlists.py:290: error: "Container" has no attribute "address"  [attr-defined]
elftools/dwarf/locationlists.py:292: error: "Container" has no attribute "start_offset"  [attr-defined]
elftools/dwarf/locationlists.py:292: error: "Container" has no attribute "end_offset"  [attr-defined]
elftools/dwarf/locationlists.py:292: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:294: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/locationlists.py:294: error: "Container" has no attribute "length"  [attr-defined]
elftools/dwarf/locationlists.py:294: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:296: error: "Container" has no attribute "start_address"  [attr-defined]
elftools/dwarf/locationlists.py:296: error: "Container" has no attribute "end_address"  [attr-defined]
elftools/dwarf/locationlists.py:296: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/locationlists.py:298: error: "Container" has no attribute "loc_expr"  [attr-defined]
elftools/dwarf/dwarfinfo.py:478: error: "Container" has no attribute "address_size"  [attr-defined]
elftools/elf/structs.py:429: error: "Container" has no attribute "pr_datasz"  [attr-defined]
elftools/elf/structs.py:430: error: "Container" has no attribute "pr_datasz"  [attr-defined]
elftools/elf/structs.py:433: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/structs.py:435: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/structs.py:437: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/structs.py:439: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/structs.py:441: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/structs.py:441: error: "Container" has no attribute "pr_datasz"  [attr-defined]
elftools/elf/notes.py:71: error: "Container" has no attribute "pr_datasz"  [attr-defined]
elftools/elf/dynamic.py:67: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:68: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:69: error: "Container" has no attribute "d_val"  [attr-defined]
elftools/elf/dynamic.py:77: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:80: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:81: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:83: error: "Container" has no attribute "d_ptr"  [attr-defined]
elftools/elf/dynamic.py:84: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/dynamic.py:203: error: "Container" has no attribute "d_tag"  [attr-defined]
elftools/elf/elffile.py:152: error: "Container" has no attribute "sh_type"  [attr-defined]
elftools/elf/elffile.py:306: error: "Container" has no attribute "sh_offset"  [attr-defined]
elftools/elf/elffile.py:397: error: "Container" has no attribute "sh_offset"  [attr-defined]
elftools/elf/descriptions.py:59: error: "Container" has no attribute "d_val"  [attr-defined]
elftools/elf/descriptions.py:300: error: "Container" has no attribute "pr_type"  [attr-defined]
elftools/elf/descriptions.py:301: error: "Container" has no attribute "pr_data"  [attr-defined]
elftools/elf/descriptions.py:302: error: "Container" has no attribute "pr_datasz"  [attr-defined]
```
Similar missing db4fb21cd6040f1cca70ddc4a3fd5e42996f3f0b is responsible for
```
elftools/elf/enums.py:37: error: Dict entry 2 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:64: error: Dict entry 22 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:76: error: Dict entry 7 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:278: error: Dict entry 187 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:324: error: Dict entry 30 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:393: error: Dict entry 5 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:421: error: Dict entry 14 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:454: error: Dict entry 8 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:473: error: Dict entry 14 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:485: error: Dict entry 7 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:489: error: Dict entry 0 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:497: error: Dict entry 3 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:585: error: Dict entry 83 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:748: error: Dict entry 51 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:795: error: Dict entry 43 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:838: error: Dict entry 39 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:848: error: Dict entry 6 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:951: error: Dict entry 98 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1023: error: Dict entry 4 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1042: error: Dict entry 5 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1054: error: Dict entry 7 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1065: error: Dict entry 6 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1077: error: Dict entry 7 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
elftools/elf/enums.py:1085: error: Dict entry 4 has incompatible type "str": "type[Pass]"; expected "str": "int"  [dict-item]
```
These I do not know how to fix - their type is not static and depends dynamically on the opened ELF file:
```
elftools/dwarf/ranges.py:50: error: Cannot determine type of "dwarfinfo"  [has-type]
elftools/dwarf/ranges.py:51: error: Cannot determine type of "dwarfinfo"  [has-type]
elftools/dwarf/locationlists.py:63: error: Cannot determine type of "dwarfinfo"  [has-type]
elftools/dwarf/locationlists.py:64: error: Cannot determine type of "dwarfinfo"  [has-type]
elftools/elf/notes.py:25: error: Cannot determine type of "structs"  [has-type]
elftools/elf/notes.py:30: error: Cannot determine type of "structs"  [has-type]
elftools/elf/notes.py:48: error: Cannot determine type of "structs"  [has-type]
elftools/elf/notes.py:56: error: Cannot determine type of "structs"  [has-type]
elftools/elf/notes.py:60: error: Cannot determine type of "structs"  [has-type]
elftools/elf/notes.py:70: error: Cannot determine type of "structs"  [has-type]
elftools/elf/sections.py:42: error: Cannot determine type of "structs"  [has-type]
elftools/elf/sections.py:180: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:63: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:131: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:185: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:345: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:347: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:349: error: Cannot determine type of "structs"  [has-type]
elftools/elf/relocation.py:351: error: Cannot determine type of "structs"  [has-type]
elftools/elf/hash.py:49: error: Cannot determine type of "structs"  [has-type]
elftools/elf/hash.py:115: error: Cannot determine type of "structs"  [has-type]
elftools/elf/hash.py:120: error: Cannot determine type of "structs"  [has-type]
elftools/elf/hash.py:121: error: Cannot determine type of "structs"  [has-type]
elftools/elf/gnuversions.py:153: error: Cannot determine type of "structs"  [has-type]
elftools/elf/gnuversions.py:196: error: Cannot determine type of "structs"  [has-type]
elftools/elf/dynamic.py:110: error: Cannot determine type of "structs"  [has-type]
```
And finally the last group of issues, which are also caused by missing cc7b1eaa277fd:
```
elftools/dwarf/structs.py:413: error: Unused "type: ignore" comment  [unused-ignore]
elftools/common/utils.py:41: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/common/utils.py:41: error: A function returning TypeVar should receive at least one argument containing the same TypeVar  [type-var]
elftools/elf/structs.py:54: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:55: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:56: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:57: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:58: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:59: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:60: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:61: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/elf/structs.py:62: error: "FormatField" expects no type arguments, but 1 given  [type-arg]
elftools/dwarf/descriptions.py:93: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int")  [str-format]
elftools/dwarf/descriptions.py:135: error: Incompatible types in string interpolation (expression has type "ListContainer", placeholder has type "int | float | SupportsInt")  [str-format]
elftools/dwarf/descriptions.py:135: error: Incompatible types in string interpolation (expression has type "None", placeholder has type "int | float | SupportsInt")  [str-format]
elftools/dwarf/descriptions.py:149: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int | float | SupportsInt")  [str-format]
elftools/ehabi/ehabiinfo.py:58: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int | float | SupportsInt")  [str-format]
elftools/ehabi/ehabiinfo.py:164: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int")  [str-format]
elftools/ehabi/ehabiinfo.py:164: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int | float | SupportsInt")  [str-format]
elftools/ehabi/ehabiinfo.py:193: error: Incompatible types in string interpolation (expression has type "None", placeholder has type "int")  [str-format]
elftools/ehabi/ehabiinfo.py:204: error: Incompatible types in string interpolation (expression has type "int | None", placeholder has type "int")  [str-format]
elftools/elf/sections.py:106: error: Incompatible types in string interpolation (expression has type "str", placeholder has type "int")  [str-format]
elftools/elf/descriptions.py:349: error: Incompatible types in string interpolation (expression has type "str | int", placeholder has type "int")  [str-format]
```

Please have a 1st look.

Then we can decide on how to proceed, e.g. just merge it or try to extract a subset for only some public API files.

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import itertools
 from collections import defaultdict
 from functools import cached_property
-from typing import IO, TYPE_CHECKING, Any, TypedDict
+from typing import IO, TYPE_CHECKING, Any, Protocol, TypedDict, cast, runtime_checkable
 
 from ..common.exceptions import ELFError
 from ..common.utils import elf_assert, struct_parse, parse_cstring_from_stream
@@ -33,6 +33,16 @@ class RelocationTables(TypedDict, total=False):
     RELA: RelocationTable
     RELR: RelrRelocationTable
     JMPREL: RelocationTable
+
+
+@runtime_checkable
+class _StringTable(Protocol):
+    """Common base-class of elftools.elf.dynamic._DynamicStringTable and
+    elftools.elf.section.StringTableSection to be consumed by
+    DynamicTag|Dynamic.
+    Requires @runtime_checkable as `assert isinstance(…, _StringTable)` is
+    used."""
+    def get_string(self, offset: int, /) -> str: ...
 
 
 class _DynamicStringTable:
@@ -66,7 +76,7 @@ class DynamicTag:
     def __init__(
         self,
         entry: Container,
-        stringtable,
+        stringtable: _StringTable | None,
     ) -> None:
         if stringtable is None:
             raise ELFError('Creating DynamicTag without string table')
@@ -98,7 +108,7 @@ class Dynamic:
         self,
         stream: IO[bytes],
         elffile: ELFFile,
-        stringtable,
+        stringtable: _StringTable | Section | None,
         position: int,
         empty: bool,
     ) -> None:
@@ -129,7 +139,7 @@ class Dynamic:
         self._empty = empty
 
         # Do not access this directly yourself; use _get_stringtable() instead.
-        self._stringtable = stringtable
+        self._stringtable: _StringTable | Section | None = stringtable
 
     def get_table_offset(self, tag_name: str) -> tuple[int | None, int | None]:
         """ Return the virtual address and file offset of a dynamic table.
@@ -148,13 +158,14 @@ class Dynamic:
 
         return ptr, offset
 
-    def _get_stringtable(self):
+    def _get_stringtable(self) -> _StringTable:
         """ Return a string table for looking up dynamic tag related strings.
 
             This won't be a "full" string table object, but will at least
             support the get_string() function.
         """
         if self._stringtable:
+            assert isinstance(self._stringtable, _StringTable)
             return self._stringtable
 
         # If the ELF has stripped its section table (which is unusual, but
@@ -163,11 +174,13 @@ class Dynamic:
         _, table_offset = self.get_table_offset('DT_STRTAB')
         if table_offset is not None:
             self._stringtable = _DynamicStringTable(self._stream, table_offset)
+            assert isinstance(self._stringtable, _StringTable)
             return self._stringtable
 
         # That didn't work for some reason.  Let's use the section header
         # even though this ELF is super weird.
         self._stringtable = self.elffile.get_section_by_name('.dynstr')
+        assert isinstance(self._stringtable, _StringTable)
         return self._stringtable
 
     def _iter_tags(self, type: str | None = None) -> Iterator[Container]:
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import itertools
 from collections import defaultdict
 from functools import cached_property
-from typing import IO, TYPE_CHECKING, Any
+from typing import IO, TYPE_CHECKING, Any, TypedDict
 
 from ..common.exceptions import ELFError
 from ..common.utils import elf_assert, struct_parse, parse_cstring_from_stream
@@ -26,6 +26,13 @@ if TYPE_CHECKING:
 
     from ..construct.lib.container import Container
     from .elffile import ELFFile
+
+
+class RelocationTables(TypedDict, total=False):
+    REL: RelocationTable
+    RELA: RelocationTable
+    RELR: RelrRelocationTable
+    JMPREL: RelocationTable
 
 
 class _DynamicStringTable:
@@ -211,14 +218,14 @@ class Dynamic:
 
         return None
 
-    def get_relocation_tables(self):
+    def get_relocation_tables(self) -> RelocationTables:
         """ Load all available relocation tables from DYNAMIC tags.
 
             Returns a dictionary mapping found table types (REL, RELA,
             RELR, JMPREL) to RelocationTable objects.
         """
 
-        result = {}
+        result: RelocationTables = {}
 
         if list(self.iter_tags('DT_REL')):
             result['REL'] = RelocationTable(self.elffile,
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -317,8 +317,8 @@ class DynamicSegment(Segment, Dynamic):
         # from the corresponding hash table
         _, gnu_hash_offset = self.get_table_offset('DT_GNU_HASH')
         if gnu_hash_offset is not None:
-            hash_section = GNUHashTable(self.elffile, gnu_hash_offset, self)
-            return hash_section.get_number_of_symbols()
+            gnu_hash_section = GNUHashTable(self.elffile, gnu_hash_offset, self)
+            return gnu_hash_section.get_number_of_symbols()
 
         # If DT_GNU_HASH did not exist, maybe we can use DT_HASH
         _, hash_offset = self.get_table_offset('DT_HASH')
--- a/elftools/elf/hash.py
+++ b/elftools/elf/hash.py
@@ -9,7 +9,7 @@
 from __future__ import annotations
 
 import struct
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Protocol
 
 from ..common.utils import struct_parse
 from ..construct.lib.container import Container
@@ -18,6 +18,13 @@ from .sections import Section
 if TYPE_CHECKING:
     from .elffile import ELFFile
     from .sections import Symbol
+
+
+class _SymbolTable(Protocol):
+    """Common base-class of elftools.elf.sections.SymbolTableSection and
+    elftools.elf.dynamic.DynamicSegment to be consumed by
+    (ELF|GNU)Hash(Section|Table)."""
+    def get_symbol(self, index: int, /) -> Symbol | None: ...
 
 
 class ELFHashTable:
@@ -39,7 +46,7 @@ class ELFHashTable:
         elffile: ELFFile,
         start_offset: int,
         size: int | None,
-        symboltable,
+        symboltable: _SymbolTable,
     ) -> None:
         """
         Args:
@@ -108,7 +115,7 @@ class ELFHashSection(Section, ELFHashTable):
         header: Container,
         name: str,
         elffile: ELFFile,
-        symboltable,
+        symboltable: _SymbolTable,
     ) -> None:
         Section.__init__(self, header, name, elffile)
         ELFHashTable.__init__(self, elffile, self['sh_offset'], self['sh_size'], symboltable)
@@ -131,7 +138,7 @@ class GNUHashTable:
         self,
         elffile: ELFFile,
         start_offset: int,
-        symboltable,
+        symboltable: _SymbolTable,
     ) -> None:
         self.elffile = elffile
         self._symboltable = symboltable
@@ -225,7 +232,7 @@ class GNUHashSection(Section, GNUHashTable):
         header: Container,
         name: str,
         elffile: ELFFile,
-        symboltable,
+        symboltable: _SymbolTable,
     ) -> None:
         Section.__init__(self, header, name, elffile)
         GNUHashTable.__init__(self, elffile, self['sh_offset'], symboltable)
--- a/test/test_debuglink.py
+++ b/test/test_debuglink.py
@@ -14,6 +14,12 @@ from typing import IO
 from elftools.common.exceptions import ELFError
 from elftools.elf.elffile import ELFFile
 
+try:
+    from typeguard import suppress_type_checks
+except ImportError:
+    def suppress_type_checks(f):  # type: ignore[no-redef]
+        return f
+
 
 class TestDebuglink(unittest.TestCase):
     """ This test verifies that the .gnu_debuglink section is followed and parsed correctly.
@@ -113,6 +119,7 @@ class TestDebuglink(unittest.TestCase):
             with self.assertRaisesRegex(ELFError, 'must be relative'):
                 loader(os.path.abspath(os.path.join(tmpdir, 'outside.debug')))
 
+    @suppress_type_checks
     def test_relative_loader_rejects_bytes_paths(self):
         with self.assertRaisesRegex(TypeError, 'base_path must be str'):
             ELFFile.make_relative_loader(b'sample.elf')
--- a/test/test_hash.py
+++ b/test/test_hash.py
@@ -5,10 +5,11 @@
 # This code is in the public domain
 #-------------------------------------------------------------------------------
 import unittest
+import unittest.mock
 import os
 
 from elftools.elf.elffile import ELFFile
-from elftools.elf.hash import ELFHashTable, GNUHashTable
+from elftools.elf.hash import ELFHashTable, GNUHashTable, _SymbolTable
 
 class TestELFHash(unittest.TestCase):
     """ Tests for the ELF hash table.
@@ -61,8 +62,8 @@ class TestELFHash(unittest.TestCase):
     def test_empty_table_without_header(self):
         """ Verify we can handle an empty (0 byte) ELF hash section.
         """
-        elffile = None
-        symboltable = None
+        elffile = unittest.mock.MagicMock(ELFFile)
+        symboltable = unittest.mock.MagicMock(_SymbolTable)
         empty_hash_section = ELFHashTable(elffile, 0, 0, symboltable)
         self.assertEqual(empty_hash_section.get_number_of_symbols(), 0)
         self.assertEqual(empty_hash_section.params['nbuckets'], 0)
--- a/elftools/dwarf/descriptions.py
+++ b/elftools/dwarf/descriptions.py
@@ -27,6 +27,8 @@ if TYPE_CHECKING:
     from .die import AttributeValue
     from .structs import DWARFStructs
 
+    _INT = TypeVar("_INT", bound=int)
+
 
 def set_global_machine_arch(machine_arch: str) -> None:
     global _MACHINE_ARCH
@@ -447,7 +449,7 @@ _DESCR_CFI_REGISTER_RULE_TYPE = dict(
 )
 
 def _make_extra_mapper(
-    mapping,
+    mapping: Mapping[_INT, str],
     default: str,
     default_interpolate_value: bool = False,
 ) -> Callable[[AttributeValue, DIE, int], str]:
--- a/elftools/dwarf/descriptions.py
+++ b/elftools/dwarf/descriptions.py
@@ -9,7 +9,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal, overload
 
 from .constants import (
     DW_ACCESS, DW_ATE, DW_CC, DW_CFA, DW_ID, DW_INL, DW_LANG, DW_ORD, DW_VIRTUALITY, DW_VIS,
@@ -168,6 +168,10 @@ def describe_DWARF_expr(expr: Any, structs: DWARFStructs, cu_offset: int | None 
     return '(' + dwarf_expr_dumper.dump_expr(expr, cu_offset) + ')'
 
 
+@overload
+def describe_reg_name(regnum: int, machine_arch: str | None, default: Literal[False]) -> str | None: ...
+@overload
+def describe_reg_name(regnum: int, machine_arch: str | None = ..., default: Literal[True] = ...) -> str: ...
 def describe_reg_name(regnum: int, machine_arch: str | None = None, default: bool = True) -> str | None:
     """ Provide a textual description for a register name, given its serial
         number. The number is expected to be valid.
--- a/elftools/elf/descriptions.py
+++ b/elftools/elf/descriptions.py
@@ -19,6 +19,7 @@ from .enums import (
     ENUM_DT_FLAGS_1, ENUM_RELOC_TYPE_PPC)
 from .constants import (
     P_FLAGS, RH_FLAGS, SH_FLAGS, SUNW_SYMINFO_FLAGS, VER_FLAGS)
+from .dynamic import DynamicSection
 
 if TYPE_CHECKING:
     from collections.abc import Container as TContainer
@@ -55,6 +56,7 @@ def describe_e_type(x: str, elffile: ELFFile | None = None) -> str:
         # Detect whether this is a normal SO or a PIE executable
         dynamic = elffile.get_section_by_name('.dynamic')
         if dynamic:
+            assert isinstance(dynamic, DynamicSection)
             for t in dynamic.iter_tags('DT_FLAGS_1'):
                 if t.entry.d_val & ENUM_DT_FLAGS_1['DF_1_PIE']:
                     return 'DYN (Position-Independent Executable file)'
--- a/elftools/elf/descriptions.py
+++ b/elftools/elf/descriptions.py
@@ -8,7 +8,7 @@
 #-------------------------------------------------------------------------------
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Final
+from typing import TYPE_CHECKING, Any, Final, TypeVar
 
 from .enums import (
     ENUM_D_TAG, ENUM_E_VERSION, ENUM_P_TYPE_BASE, ENUM_SH_TYPE_BASE,
@@ -21,10 +21,14 @@ from .constants import (
     P_FLAGS, RH_FLAGS, SH_FLAGS, SUNW_SYMINFO_FLAGS, VER_FLAGS)
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
+    from collections.abc import Container as TContainer
+    from collections.abc import Iterable, Mapping
 
     from ..construct.lib.container import Container
     from .elffile import ELFFile
+
+    _K = TypeVar("_K")
+    _V = TypeVar("_V")
 
 
 def describe_ei_class(x: str) -> str:
@@ -705,7 +709,7 @@ _DESCR_NOTE_GNU_PROPERTY_RISCV_FEATURE_1_AND = (
     (2, 'ZICFISS'),
 )
 
-def _reverse_dict(d, low_priority=()):
+def _reverse_dict(d: Mapping[_K, _V], low_priority: TContainer[_K] = ()) -> dict[_V, _K]:
     """
     This is a tiny helper function to "reverse" the keys/values of a dictionary
     provided in the first argument, i.e. {k: v} becomes {v: k}.
@@ -714,7 +718,7 @@ def _reverse_dict(d, low_priority=()):
     the case of conflicting values - if a value is present in this list, it will
     not override any other entries of the same value.
     """
-    out = {}
+    out: dict[_V, _K] = {}
     for k, v in d.items():
         if v in out and k in low_priority:
             continue
--- a/elftools/elf/elffile.py
+++ b/elftools/elf/elffile.py
@@ -114,7 +114,7 @@ class ELFFile:
         return ELFFile(stream, ELFFile.make_relative_loader(os.fsdecode(path)))
 
     @staticmethod
-    def make_relative_loader(base_path) -> Callable[[str], IO[bytes]]:
+    def make_relative_loader(base_path: str) -> Callable[[str], IO[bytes]]:
         """ Return a function that takes a potentially relative path,
             resolves it against base_path (str), and opens a file at that.
 
@@ -126,7 +126,7 @@ class ELFFile:
             raise TypeError('base_path must be str')
         base_directory = os.path.realpath(os.path.dirname(base_path))
 
-        def loader(rel_path) -> IO[bytes]:
+        def loader(rel_path: str) -> IO[bytes]:
             if not isinstance(rel_path, str):
                 raise TypeError('rel_path must be str')
 
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -9,7 +9,7 @@
 from __future__ import annotations
 
 from functools import cached_property
-from typing import IO, TYPE_CHECKING, Any, Callable, NamedTuple
+from typing import IO, TYPE_CHECKING, Any, NamedTuple, Protocol
 
 from ..common.exceptions import ELFRelocationError
 from ..common.utils import elf_assert, struct_parse
@@ -21,7 +21,7 @@ from .enums import (
 from ..construct import Container
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Iterator, Mapping
 
     from .elffile import ELFFile
 
@@ -216,6 +216,10 @@ class RelrRelocationSection(Section, RelrRelocationTable):
             self['sh_offset'], self['sh_size'], self['sh_entsize'])
 
 
+class _RelocationFunction(Protocol):
+    def __call__(self, value: int, sym_value: int, offset: int, addend: int = 0) -> int: ...
+
+
 def _reloc_calc_identity(value: int, sym_value: int, offset: int, addend: int = 0) -> int:
     return value
 
@@ -394,9 +398,9 @@ class RelocationHandler:
     class _RELOCATION_RECIPE_TYPE(NamedTuple):
         bytesize: int
         has_addend: bool
-        calc_func: Callable[..., int]
+        calc_func: _RelocationFunction
 
-    _RELOCATION_RECIPES_ARM = {
+    _RELOCATION_RECIPES_ARM: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_ARM['R_ARM_ABS32']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=False,
             calc_func=_reloc_calc_sym_plus_value),
@@ -405,7 +409,7 @@ class RelocationHandler:
             calc_func=_arm_reloc_calc_sym_plus_value_pcrel),
     }
 
-    _RELOCATION_RECIPES_AARCH64 = {
+    _RELOCATION_RECIPES_AARCH64: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_AARCH64['R_AARCH64_ABS64']: _RELOCATION_RECIPE_TYPE(
             bytesize=8, has_addend=True, calc_func=_reloc_calc_sym_plus_addend),
         ENUM_RELOC_TYPE_AARCH64['R_AARCH64_ABS32']: _RELOCATION_RECIPE_TYPE(
@@ -416,14 +420,14 @@ class RelocationHandler:
     }
 
     # https://dmz-portal.mips.com/wiki/MIPS_relocation_types
-    _RELOCATION_RECIPES_MIPS_REL = {
+    _RELOCATION_RECIPES_MIPS_REL: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_MIPS['R_MIPS_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=False, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_MIPS['R_MIPS_32']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=False,
             calc_func=_reloc_calc_sym_plus_value),
     }
-    _RELOCATION_RECIPES_MIPS_RELA = {
+    _RELOCATION_RECIPES_MIPS_RELA: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_MIPS['R_MIPS_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=True, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_MIPS['R_MIPS_32']: _RELOCATION_RECIPE_TYPE(
@@ -434,7 +438,7 @@ class RelocationHandler:
             calc_func=_reloc_calc_sym_plus_value),
     }
 
-    _RELOCATION_RECIPES_PPC64 = {
+    _RELOCATION_RECIPES_PPC64: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_PPC64['R_PPC64_ADDR32']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=True, calc_func=_reloc_calc_sym_plus_addend),
         ENUM_RELOC_TYPE_PPC64['R_PPC64_REL32']: _RELOCATION_RECIPE_TYPE(
@@ -443,7 +447,7 @@ class RelocationHandler:
             bytesize=8, has_addend=True, calc_func=_reloc_calc_sym_plus_addend),
     }
 
-    _RELOCATION_RECIPES_X86 = {
+    _RELOCATION_RECIPES_X86: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_i386['R_386_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=False, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_i386['R_386_32']: _RELOCATION_RECIPE_TYPE(
@@ -454,7 +458,7 @@ class RelocationHandler:
             calc_func=_reloc_calc_sym_plus_value_pcrel),
     }
 
-    _RELOCATION_RECIPES_X64 = {
+    _RELOCATION_RECIPES_X64: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_x64['R_X86_64_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=8, has_addend=True, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_x64['R_X86_64_64']: _RELOCATION_RECIPE_TYPE(
@@ -469,7 +473,7 @@ class RelocationHandler:
     }
 
     # https://www.kernel.org/doc/html/latest/bpf/llvm_reloc.html#different-relocation-types
-    _RELOCATION_RECIPES_EBPF = {
+    _RELOCATION_RECIPES_EBPF: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_BPF['R_BPF_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=8, has_addend=False, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_BPF['R_BPF_64_64']: _RELOCATION_RECIPE_TYPE(
@@ -485,7 +489,7 @@ class RelocationHandler:
     }
 
     # https://github.com/loongson/la-abi-specs/blob/release/laelf.adoc
-    _RELOCATION_RECIPES_LOONGARCH = {
+    _RELOCATION_RECIPES_LOONGARCH: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_LOONGARCH['R_LARCH_NONE']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=False, calc_func=_reloc_calc_identity),
         ENUM_RELOC_TYPE_LOONGARCH['R_LARCH_32']: _RELOCATION_RECIPE_TYPE(
@@ -526,7 +530,7 @@ class RelocationHandler:
             calc_func=_reloc_calc_sym_plus_addend_pcrel),
     }
 
-    _RELOCATION_RECIPES_S390X = {
+    _RELOCATION_RECIPES_S390X: Mapping[int, _RELOCATION_RECIPE_TYPE] = {
         ENUM_RELOC_TYPE_S390X['R_390_32']: _RELOCATION_RECIPE_TYPE(
             bytesize=4, has_addend=True, calc_func=_reloc_calc_sym_plus_addend),
         ENUM_RELOC_TYPE_S390X['R_390_PC32']: _RELOCATION_RECIPE_TYPE(
--- a/elftools/elf/relocation.py
+++ b/elftools/elf/relocation.py
@@ -171,6 +171,7 @@ class RelrRelocationTable:
             else:
                 # We're processing a bitmap.
                 elf_assert(base is not None, 'RELR bitmap without base address')
+                assert base is not None
                 i = 0
                 while True:
                     # Iterate over all bits except the least significant one.
--- a/elftools/elf/sections.py
+++ b/elftools/elf/sections.py
@@ -17,11 +17,11 @@ from ..common.utils import struct_parse, elf_assert, parse_cstring_from_stream
 from collections import defaultdict
 from .constants import SH_FLAGS
 from .notes import iter_notes
+from elftools.construct.lib.container import Container
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
-    from ..construct.lib.container import Container
     from .elffile import ELFFile
     from .structs import ELFStructs
 
@@ -333,6 +333,8 @@ class StabSection(Section):
 class Attribute:
     """ Attribute object - representing a build attribute of ELF files.
     """
+    if TYPE_CHECKING:
+        value: Any
 
     def __init__(self, structs: ELFStructs, stream: IO[bytes]) -> None:
         self._tag = self._parse(structs, stream)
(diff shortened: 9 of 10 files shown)


eliben (commented):
Just a couple of initial questions

In general, I'd really prefer someone with Python type checking experience to take a careful look at this

elftools/common/utils.py
```diff
@@ -6,24 +6,42 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from __future__ import annotations
+
 from contextlib import contextmanager
+from typing import IO, TYPE_CHECKING, Any, TypeVar, overload
```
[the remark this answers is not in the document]
pmhahn: At runtime at least you have to import `TYPE_CHECKING`, which defaults to `False` during runtime, but type-checkers like `mypy` and `pyright` have set it to `True`. This is used to only import certain symbols when type-checking, as those imports may introduce two problems:
- `import`-cycles – which can be mitigated by only using `import $module` instead of `from $module import $symbols` 
- `typing` exists since Python 3.6+, but over time may new _constructs_ have been added to it like `Concatenate` and `ParamSpect` with 3.10, `Self` with 3.11 or `Buffer` with 3.12. Importing them (or any other symbol used for typing) at runtime will fail if your version of Python is too old. You have to  guard them with `try: … except ImportError: …` or with `if TYPE_CHECKING:` – and hope that the version of people running type-checkers is now enough to have them. For older Python versions there is `from typing_extensions import …`, which contains backports of these newer constructs.

While `TypeVar` is here only used in the `if TYPE_CHECKING:` section below, the symbol is old enough to be always available – it is there since Python 3.6, so no need to guard that import.
On the other hand having another `from typing import TypeVar, …` in that later section can be done, but it is a matter of taste: `ruff check` will not complain about an _unused import_ as actually it is used in the section below, although guarded by that `if TYPE_CHECKING:` which is irrelevant for `ruff`.

(I prefer to only do one `from typing import …` if possible; if you find a 2nd on in the `TYPE_CHECKING` block, there are subtle reasons to have it only there.)

elftools/common/utils.py
```diff
@@ -6,24 +6,42 @@
...
+    from .construct_utils import ULEB128, SLEB128, UBInt24, ULInt24
+    from ..dwarf.dwarfinfo import DebugSectionDescriptor
+
+    _T = TypeVar("_T")
+    _K = TypeVar("_K")
+    _V = TypeVar("_V")
+
 
-def merge_dicts(*dicts):
+def merge_dicts(*dicts: Mapping[_K, _V]) -> dict[_K, _V]:
```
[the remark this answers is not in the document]
pmhahn: `_K` and `_V` are just place-holders for any type – there is no restriction on their definition above like `_X = TypeVar("_X", bound=Symbol)` for _only any subclass `Symbol` or `Symbol` itself_ or `_Y = TypeVar("_Y", str, bytes)` for _everywhere `str` or `bytes`, but no mix of them_.

So the type annotation here says: you may pass in multiple dictionaries, all with the same type for keys and values, and you will get out again a dictionary with that type. So
- `merge_dict( {int: str}, {int: str} ) -> {int: str}` is okay: `_K = int`, `_V = str`.
- `merge_dict( {bool: int}, {str: float} )` is not okay as `bool != str` for `_K` and `int != float` for `_V`.
- `merge_dict( {int: str}, defaultdict[int: str], OrderedDict[int: str], UserDict[int: str] ) -> {int: str}` is also okay as long as they all implement the `Mapping`-protocol using the right types – you're not restricted to only use `dict`, but may use sub-classes of `dict` or even you complete own class as long as you implement the `Mapping`-protocol.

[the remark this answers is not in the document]
pmhahn: Yes and no:
1. You are using [Generic function](https://mypy.readthedocs.io/en/stable/generics.html#generic-functions) syntax, which is only available since Python 3.12; using that version would prevent anyone using an older version of Python using your code.
2. The extra type annotation does not give you anything extra except that you no longer have to declare those 2 `… = TypeVar("…")` yourself.
3. `Mapping` is a `Protocol` for a **read-only** `dict`, so calling `result.update(…)` triggers an error:
    > "Mapping[K, V]" has no attribute "update"  [attr-defined]                                                                                                                                                  
4. You should follow the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) (Postal's law):
    > be conservative in what you send, be liberal in what you accept.

    `dicts` is an _input_ variable and we only need it to follow the `Mapping` protocol, e.g. be `Iterable[K]` for `for d in dicts:` to work and to be a _read-only_ dict, so `result.update(d)` can do its work (e.g. `for k in d.keys(): result[k] = d[k]`).
    But for the _return result_ we know that it will be a `dict[K, V]` as we allocate it in the first line of the function.
    By declaring the return type to be `Mapping[K, V]` you force any caller to use the result in only a _read-only_ manner. Calling a function like `f(d: dict[K, V])` would no longer be allowed (by type-checking) as that function requires an _read-write_ `dict[K, V]`; you would have to call that function with `f(dict(merge_dicts(…)))`.

eliben (commented):
One question and general comment: would it be possible to split this PR to multiple? Even if just for the sake of review - could we start with a small-medium PR with a representative set of changes? It's OK if the intermediate steps don't fully type check until everything has landed

elftools/common/construct_utils.py
```diff
@@ -41,15 +48,17 @@ def _parse(self, stream, context):
...
         raise NotImplementedError('no building')
-    def _sizeof(self, context):
+    def _sizeof(self, context: Container) -> int:
         raise SizeofError("can't calculate size")
 
 class ULEB128(Construct):
     """A construct based parser for ULEB128 encoding.
     """
-    def _parse(self, stream, context):
+    if TYPE_CHECKING:
```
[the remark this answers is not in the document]
pmhahn: It's a "hack": Looking at `elftools.construct.core.Construct`:
```python
    def __init__(self, name: str | None, flags: …) -> None:
        if name is not None: …
        self.name = name
```
so `name` can be `None` and type-checkers will force you to always check for that case:
```python
if self.name is None:
   …
else:
    …
```
Currently this is not done and leads to many errors, where you must change the code then to silence these.
But: For most _Constructs_ we know from the `pyelftools` perspective, that `name` will never be `None` as it makes little to no sense to have an _anonymous_ construct of those types. By using
```python
if TYPE_CHECKING:
    name: str
```
I tell the type checkers that the class has a variable of that type, which overwrites the more generic type `name: str | None` from `Construct`. `None` is no longer an allowed type and `self.name = None` would be an error for type checkers.
At runtime there is no difference because of the `if TYPE_CHECKING:` conditional.
So it's there to give type-checkers some more information which they cannot deduce from looking at the code only, but we humans know how that code is used.

pmhahn:
> One question and general comment: would it be possible to split this PR to multiple? Even if just for the sake of review - could we start with a small-medium PR with a representative set of changes? It's OK if the intermediate steps don't fully type check until everything has landed

Yes, I can do that. Any advise on how to split best? Internel / low-level / high-level?

elftools/common/construct_utils.py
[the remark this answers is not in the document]
eliben: makes sense, please add a brief comment!

eliben:
> > One question and general comment: would it be possible to split this PR to multiple? Even if just for the sake of review - could we start with a small-medium PR with a representative set of changes? It's OK if the intermediate steps don't fully type check until everything has landed
> 
> Yes, I can do that. Any advise on how to split best? Internel / low-level / high-level?

Yes, by layers could be a great way to slice it. Starting at the lowest possible and then progressing upwards

elftools/common/construct_utils.py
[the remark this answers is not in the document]
pmhahn: Done: `name: str  # for typing disallow None`

eliben:
Where do we stand with this PR? How much of it has already been added?

pmhahn:
> Where do we stand with this PR? How much of it has already been added?

Sorry for the delay, I'm currently busy otherwise. Hope to find some time next weekend.

pmhahn:
> What's the status here?

The bad news: Sadly I'm busy otherwise .
The good news: I found some time to update the PR and it again is in a state, where it could be merged:

`pyright` got unhappy, so I again had to re-introduce 6cb25b9 and fix some hints 7a0c243 , which already have been merged. `mypy` is still unhappy with a lot of things.

I have been experimenting with [typeguard](https://typeguard.readthedocs.io/en/stable/index.html), which turns those type-hints into *runtime checks*. This allows validating those hints when running the test-suite. Sadly that drastically increases the runtime and – even more sad – shows several errors in my type annotations. 
I have locally converted the unit-test to use `pytest` instead of Pythons built-in `unittest` as that allowed me to easily get coverage reports, but that also needs more work.

elftools/common/construct_utils.py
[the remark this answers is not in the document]
Timmmm: That comment doesn't really explain it though... Anyway it is better to add `assert self.name is not None` in the appropriate places.

Timmmm (approved):
This looks great to me. It would be good to split into smaller PRs - maybe you could do one with all the basic `str`, `int`, `None` types that are pretty obvious and then a second one with everything else?

On the other hand that's super tedious to do and nobody has time for that. If this were my project I think I would just merge it as-is and improve it in future PRs.

If this passes Pyright that's already a huge achievement and improvement.

k4lizen:
Awesome work! I would love to see this merged, we got bitten by it recently here: https://github.com/pwndbg/pwndbg/pull/3470#issuecomment-3661619313

pmhahn:
- [x] I rebased my branch and fixed the conflicts.
- [x] [`mypy 1.19.1`](/https://mypy-lang.org/)  is happy again
- [x]  [`pyright 1.1.408`](https://github.com/microsoft/pyright) is also happy
- [x] [`pyrefly 0.62.0`](https://pyrefly.org/) is mostly happy except 3 constructs: https://github.com/facebook/pyrefly/issues/3241
  ```
  ERROR Argument `Construct` is not assignable to parameter `object` with type `FormatField[int] | _InitialLengthAdapter` in function `list.append` [bad-argument-type]
     --> elftools/dwarf/callframe.py:410:23
  410 |         fields.append(formats[basic_encoding]('initial_location'))
      |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ERROR Argument `Construct` is not assignable to parameter `object` with type `FormatField[int] | _InitialLengthAdapter` in function `list.append` [bad-argument-type]
     --> elftools/dwarf/callframe.py:411:23
  411 |         fields.append(formats[basic_encoding]('address_range'))
      |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ERROR Cannot assign to `TYPE_CHECKING` because it is imported as final [bad-assignment]
    --> elftools/elf/structs.py:23:20
  23 | from .enums import *
     |                    ^
  ```
- [ ] [`ty 0.0.32`](https://github.com/astral-sh/ty) still reports 43 issues. I have experienced the same in other projects that its reasoning is ~~not~~ __more__ advanced as `mypy` and `pyright` — most often they`re worth investigating. I'm ~~not~~ going to work on this, but if anyone else has time: please feel free to have a look. [PEP-767](https://peps.python.org/pep-0767/) might help in the future with some issues.
- [x] I have been running my branch with [`typeguards`](https://pypi.org/project/typeguard/) enabled, including the unit test:
    ```console
    $ cat .venv/lib/python3.12/site-packages/sitecustomize.py 
    __import__("typeguard").install_import_hook("elftools")
    ```

k4lizen (approved):

k4lizen:
@eliben 
> I'll have to think about my strategy here overall...

Any updates on this? Merging this PR is not going to break any end users because there are no functional changes. The PR is +1,686 -1,088 which is not small but honestly not that big either, I've gone through it personally. pyelftools typing is currently broken for end users, this will only have a positive impact. If there are any issues, we will fix them when we get to them (I will gladly send a PR to fix such an issue if I encounter it), and again, they will not be functional, only part of peoples type checking pipeline. Most people have probably disabled type checking for pyelftools at this point (pwndbg has, for instance) since, again, it is currently flashing red due to the typed marker and there being no types.

By delaying merging this you are only creating more work for @pmhahn who has to rebase and resolve conflicts.

Could you elaborate what your thinking on this situation is?

pmhahn:
- I've again rebased my branch and resolved the merge conflict
- After adding some minimal configuration for Meta's `pyrefly` and Astral's `ty` the number of issues was reduced drastically: most of them where for `test/` and `examples/`, which Dropbox's `mypy` and Microsoft's `pyright` are also configured to ignore.
- [ ] `pyrefly` has https://github.com/facebook/pyrefly/issues/3241; I'd like to convert the `from .enum import *` to `import .enum as e` (or `… as enum`) and add that prefix to every usage if that is okay.
- [ ] I want to look at the `ty` findings as they look valid in several cases. TBC…

eliben:
I'm appreciative of the work @pmhahn is doing here, but the current state of the Python typing ecosystem makes it really difficult for me to commit to maintain this if it lands. I simply don't have the time to make sure 4 different typing tools are all happy with pyelftools's source code at every commit.

I'm kind of waiting for one clear winner to emerge and then annotate enough to make it happy. My main candidate for this is `ty`, and this is the approach taken in https://github.com/eliben/pycparser

I understand this may be frustrating for some, but given time limitations that's the best I can do for now.

JonathonReinhart:
> I'm kind of waiting for one clear winner to emerge

This is never going to happen.

eliben:
> > I'm kind of waiting for one clear winner to emerge
> 
> This is never going to happen.

Fair enough; what I really mean here is a clear *enough* winner such that we can say "we typecheck pyelftools using FOO, and don't care about other type checkers". At this time, for me FOO=ty; I will accept a PR that type checks pyelftools with ty and doesn't try to test all the others. But I don't think this aligns with @pmhahn 's goals, hence the impasse.

@sevaa for vis

Timmmm:
The clear enough winner *today* is Pyright. But if you have something that typechecks with Ty and you don't care about any of the others, that's clearly MUCH better than having no type hints at all.

pmhahn:
:fireworks: 
- [x] [`mypy 1.20.2`](/https://mypy-lang.org/)  is happy again
- [x]  [`pyright 1.1.409`](https://github.com/microsoft/pyright) is also happy
- [x] [`pyrefly 0.62.0`](https://pyrefly.org/) is happy
- [x] [`ty 0.0.32`](https://github.com/astral-sh/ty) is happy — it requires the same `type: ignore` as `mypy`
- [x] [`typeguards`](https://pypi.org/project/typeguard/) is happy

`ty` was the hardest, but actually found several issues the others did not find. I had to change/improve several things, sadly some innards; search for `BREAKING CHANGES` in my commit messages :exclamation: 

@eliben  The PR is now very large, but I can split it into 3 smaller ones if you prefer:
1. The 1st 9 patches are to `elftools/construct/` to fix several issues my past PR changes or which are still open. I know you don't like touching it, but as `construct` builds the foundation for `elftools`, I had to research the details anyway and it would be baste to trash them now.
2. the 2nd 11 patches contains some minor cleanups and fixes, which are not related to typing at all, but I found them while reading the code and running the test-suite many times.
3. the last 42 patches add type hints and then fix issues, the type-checkers did find. That requires renaming some variables — as `mypy` back then did not like redefining variables —, caching dictionary lookups in some temporary variables, adding explicit `cast(type, object)` to help the type checkers get it right.

The most drastic changes are
- convert all `DW_*` constants into `IntEnum`s; while that is not strictly required, this allows future improvements to typing as we now can specify, which `enum DW_…` an function takes as an argument instead of some (unconstrained) `int`. (This will also allow us to replace comparing those constants by `str`-name and move their _descriptive texts_ also into the `enum`s themselves — but not today :wink: )
- add some `TypedDict`s or `@dataclass` to improve typing; `dict[Any, Any]` does not provide enough information by default and requires many more `if`s and `cast()`s to silence the type checkers.
- Use `import enum as e` and prefixing any use with `e.ENUM…` instead of doing a `from enum import *`. That change makes `pyrefly` and `ruff check` more happy :smile: 

Tell my we what you think and what needs to be done to get this (partly) merged.

eliben:
Thanks for all the work here, @pmhahn 

Let's try to make gradual progress. 

More changes in construct is ok (since we've made other changes already), but I'd like to review in detail.
Regarding "minor cleanups and fixes, which are not related to typing at all", can you send these out first so we can get them out of the way?

> convert all DW_* constants into IntEnums; while that is not strictly required, 

I think this should be ok. `IntEnum` should behave like an int.

> add some TypedDicts or @dataclass to improve typing

Can you give some examples? I just want to make sure we're not breaking existing users for this.

> Use import enum as e and prefixing a

This should be fine.

Finally, to reiterate what I said earlier - I do not plan to make all type checkers happy with future code. Only the latest ty. Happy to accept PRs that make other checkers happy in the future when the code drifts, though.

pmhahn:
> > add some TypedDicts or @DataClass to improve typing
> 
> Can you give some examples? I just want to make sure we're not breaking existing users for this.

See last 2 commits, especially "typing: Make CFI augmentation a TypedDict": It stored `{True: True}` on finding `b"S"`, which marks a _signal stack_. While investigating this I found https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst#43common-information-entries, which also added `B` and `G`. Therefor I changed it to `{Litral["S", "B", "G"]: True}`; feel free to propose a better mapping.
PS: `TypedDict` can only handle `str`-keys, so `bool` was no option; this is also the issue with `Line`, where registers use `int` as key; that's why I had to invent that _ugly_ `class Line(Protocol)` using `@overload` to return more specific types than `int | CFARule | RegisterRule` for all keys: what would force a check at all sites, where `pc`, `pfa`, … are retrieved.

> Finally, to reiterate what I said earlier - I do not plan to make all type checkers happy with future code. Only the latest ty. Happy to accept PRs that make other checkers happy in the future when the code drifts, though.

`ty`-only is okay with me. I`ll look after the others if I need them or if someone pings me in the future and my time permits.

> Let's try to make gradual progress.

:octopus: merge of #635 #636 #637 #638 #639 works

pmhahn:
So hopefully this is the last version…
There are still some commit, which might go first as they're _motivated_ by type-checker-findings, but which make sense even without type-checking, e.g.
- make `None`-checks explicit 4f554f8 
- change some variable names ti not reuse the name for another type f6caae5 & 59bbbd1 & 40220be
- some small code changes to make sure, variables are defined in all cases of a larger `if-elif-else`-construct b8e4802 & 5ced360
- Fix `printf` formats to handle corner-cases 25ba7ef 

But most are related to
- adding types
- converting other types to be more usable
- ignoring issues in corder nacses
- fixing the test-suite to run with `typecheck`

There 2 two *breaking changes* I want to call out – please have a closer look
- 68e1ffd365fb299b97a1b13292d757b6acd1cf53 changes, how _signal stacks_ are signaled
- 08024807e55d94cf9e88a6b68d5c95aef99895db changes the file position in the error case

eliben:
Let's keep peeling off related commits into separate PRs as much as possible, proceeding gradually.

Re breaking changes, https://github.com/eliben/pyelftools/commit/68e1ffd365fb299b97a1b13292d757b6acd1cf53 probably isn't worth it just for the added type safety in this particular case. `dict[Any, Any]` is a cop-out, but I wouldn't break compatibility for this

pmhahn:
@eliben While preparing the next batch I noticed, that there are some very long lines in the code; Adding types sometime make them even longer and harder to read. `doc/hacking-guide.rst` only references  [PEP-8](https://peps.python.org/pep-0008/#maximum-line-length), which recommends 79. `black` and `ruff` default to 88.

Can you give a _preferred_ `line-length`? That would make it easier to decide, when to break long lines and when not.

Here's a quick histogram of line-lengths rounded up:
```
    8   2706    ████████████████████████████████████████                                                                                                                                      
   18   1491    ██████████████████████                                                                                                                                                        
   27   2292    █████████████████████████████████▉                                                                                                                                            
   37   2648    ███████████████████████████████████████▏                                                                                                                                      
   47   2561    █████████████████████████████████████▊                                                                                                                                        
   56   1421    █████████████████████                                                                                                                                                         
   66   1188    █████████████████▌                                                                                                                                                            
   75   1302    ███████████████████▏                                                                                                                                                          
   85   982     ██████████████▌                                                                                                                                                               
   95   179     ██▋                                                                                                                                                                           
  104   95      █▍                                                                                                                                                                            
  114   45      ▋                                                                                                                                                                             
  123   34      ▌                                                                                                                                                                             
  133   20      ▎                                                                                                                                                                             
  143   26      ▍                                                                                                                                                                             
  152   10      ▏                                                                                                                                                                             
  162   2                                                                                                                                                                                     
  181   1                                                                                                                                                                                     
  191   1                                                                                                                                                                                     
```

Maybe take a look at `git grep -n '.\{140,\}' -- elftools` or play with something like `ruff format --isolated --config 'format.docstring-code-format = false' --config 'format.quote-style = "preserve"' --line-length=100 --diff` and then document your preferred style in the [pyproject.toml](https://docs.astral.sh/ruff/settings/#indent-width)?

eliben:
I'm not too religious about it, trying to keep most lines (especially comments) to 79-80 or so, but here and there long lines are ok. Really don't want to put in mechanical limitations here or any sort of large rewriting just for the sake of normalizing line length. Let's continue with the piecemeal PRs and I'll make a comment if anything specific stands out in review.

pmhahn:
Only the near final bits remain: This are now some special type hints:
- Add [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol)s as some type do not build a class hierarchy and thus requires a common base-class for typing: `_RelocationFunction`, `_StringTable`, `_SymbolTable`, `_HashTable`
- Some more [TypeVar](https://docs.python.org/3/library/typing.html#typing.TypeVar)s to handle `int` vs. `IntEnum` and `RelocationTables`
- Add type-annotations to `loader(str)`: There is one unit-test which deliberately passed `bytes`, which needs (run-time) type-checking with [typeguard](https://pypi.org/project/typeguard/) to be disabled.
- `_reverse_dict()` requires some extra work.
- Some more [@overload](https://docs.python.org/3/library/typing.html#typing.overload)s for `__getitem__()` like functions to get more precise types in several locations for `Section` and `Segment` and `describe_reg_name()`
- Some more `assert` to narrow types (`DynamicSection`, `None`)
- Some special typing for the type-hierarchy `Attribute` and `AttributeSection` to handle its dynamic behavior, e.g. having a `type` for later creating.

The last (extra) bits as in #657

pmhahn:
> Only the near final bits remain: This are now some special type hints:

@eliben Have you missed this as you already closed #514 ?

eliben:
> > Only the near final bits remain: This are now some special type hints:
> 
> @eliben Have you missed this as you already closed #514 ?

I've closed it because it's not really useful and we've been coordinating the work here. For all practical purposes, type hints have already been added; we're now working through remaining details and enhancements.

Is this PR (611) now ready for a review and merge?

pmhahn:
> I've closed it because it's not really useful and we've been coordinating the work here. For all practical purposes, type hints have already been added; we're now working through remaining details and enhancements.

Okay.

> Is this PR (611) now ready for a review and merge?

Yes, these are the final bits.

elftools/elf/dynamic.py
```diff
@@ -26,6 +26,19 @@
...
+
+
+class RelocationTables(TypedDict, total=False):
+    REL: RelocationTable
+    RELA: RelocationTable
+    RELR: RelrRelocationTable
+    JMPREL: RelocationTable
+
+
+@runtime_checkable
```
eliben, line 38: It's be good to add some comments explaining what we're doing here with the protocol and runtime_checkable

elftools/elf/hash.py
```diff
@@ -20,6 +20,15 @@
     from .sections import Symbol
 
 
+class _SymbolTable(Protocol):
```
eliben, line 23: Also here comments would be useful

elftools/elf/hash.py
```diff
@@ -20,6 +20,15 @@
...
+    def get_symbol(self, index: int, /) -> Symbol | None: ...
+
+
+class _HashTable(Protocol):  # noqa: PYI046
```
eliben: Why is this tagged? Because it's private?
Should it be?

test/test_debuglink.py
```diff
@@ -14,6 +14,12 @@
 from elftools.common.exceptions import ELFError
 from elftools.elf.elffile import ELFFile
 
+try:
+    from typeguard import suppress_type_checks
```
eliben, line 18: What is this for?

test/test_hash.py
```diff
@@ -5,10 +5,11 @@
 # This code is in the public domain
 #-------------------------------------------------------------------------------
 import unittest
+import unittest.mock
 import os
 
 from elftools.elf.elffile import ELFFile
-from elftools.elf.hash import ELFHashTable, GNUHashTable
+from elftools.elf.hash import ELFHashTable, GNUHashTable, _SymbolTable
```
eliben, line 12: I wonder: can we just leave type checking out of these tests?
It's a shame to convolute the test code for this; I'm not sure the value is worth it.

elftools/elf/hash.py
pmhahn: See https://docs.astral.sh/ruff/rules/unused-private-protocol/

See commit message: `ELFHashTable` and `GNUHashTable` do not share a common base-class, but both implement those 2 methods, which are then used in `elftool.elf.dynamic.DynamicSegment._num_symbols()`. As that method is internal, the protocol can be internal, too.

We can get rid of this `Protocol` if we rename one(or both) of those `hash_section` there to `gnu_hash` or `elf_hash`.

test/test_debuglink.py
pmhahn: [typeguard](https://typeguard.readthedocs.io/en/stable/userguide.html) turns those type-annotations into runtime checks: I've been using this to validate, that my type-annotations actually match what is used at runtime. It found many errors in my initial typing and I've been running with it since.

As the commit message documents, 4e4ef81225a3 changed the signature of the loader to use `str` only.  But the test `test_relative_loader_rejects_bytes_paths()` calls `ELFFile.make_relative_loader(bytes)`, which violates that type-hint: `typeguard` will detect this and raise an exception, which breaks the test. Thus `typeguard` must be disabled for that specific test (only):
```console
$ python3 -m unittest discover . test_debuglink.py
..E.
======================================================================
ERROR: test_relative_loader_rejects_bytes_paths (test.test_debuglink.TestDebuglink.test_relative_loader_rejects_bytes_paths)
----------------------------------------------------------------------
  File "test/test_debuglink.py", line 124, in test_relative_loader_rejects_bytes_paths
    ELFFile.make_relative_loader(b'sample.elf')
  File "elftools/elf/elffile.py", line 117, in make_relative_loader
    def make_relative_loader(base_path: str) -> Callable[[str], IO[bytes]]:
  File ".venv/lib/python3.12/site-packages/typeguard/_functions.py", line 180, in check_argument_types_internal
    check_type_internal(value, annotation, memo)
  File ".venv/lib/python3.12/site-packages/typeguard/_checkers.py", line 994, in check_type_internal
    raise TypeCheckError(f"is not an instance of {qualified_name(origin_type)}")
typeguard.TypeCheckError: argument "base_path" (bytes) is not an instance of str

----------------------------------------------------------------------
Ran 4 tests in 0.216s

FAILED (errors=1)
```

If you drop this, no-body will be able to run the test-suite with `typeguard` unless you re-add it manually.

Alternative: drop the test or broaden the exception to also accept `typeguard.TypeCheckError(Exception)`

elftools/elf/hash.py
pmhahn: See commit message: `{ELF,GNU]Hash{Table,Section}` all expect a `symboltable`, which is either an instance of  `SymbolTableSection` or `DynamicSegment`. They do not share a common base-class, but both implement this `Protocol`.

elftools/elf/dynamic.py
pmhahn: Again see commit message: `_DynamicStringTable` and `StringTableSection` do not have a common base-class, but both are passed to `DynamicTag` and `Dynamic`, which then want's to call `get_string(int)->str` on _it_.

Worse `Dynamic` accepts any `Section` and you'll get a runtime-error if that section does not implement `get_string()`. As such the code uses `assert isinstance(self._stringtable, _StringTable)` to make sure, that the instance implements at least that method.

Alternative: Turn those `assert`s into a `if not getattr(self._stringtable, "get_string", None): raise TypeError("The returned instance does not implement the required method 'get_string(offset: int)->str'")` everywhere, but I prefer to get that for free from [typing.runtime_checkable](https://docs.python.org/3/library/typing.html#typing.runtime_checkable).

elftools/elf/hash.py
pmhahn: > We can get rid of this `Protocol` if we rename one(or both) of those `hash_section` there to `gnu_hash` or `elf_hash`.

Just did that.

test/test_hash.py
pmhahn: Again `typeguard`: The tests pass `None` instead of some _real-types_, which `typeguard` does not like. As such `unitest.mock.Mock` must be used to at least fake enough to silence `typeguard`. 
Being able to run the unit-test with `typeguard` enabled was a big win, so I prefer to keep this, but YMMV.

elftools/elf/dynamic.py
eliben: SG, just add a succinct comment in the code itself so it will be easier to understand when someone reads it a year from now.

elftools/elf/hash.py
eliben: ditto

test/test_hash.py
eliben: I don't get it.... I don't find any mentions of `typeguard` in the repo right now (other than `pyproject.toml`).

Generally, I'd like to avoid this extra dependency at all, if possible. Our goals may be different here -- my main interest is ensure that pyelftools's public API has type information so clients / users can benefit from it for their own code and documentation purposes. I care much less about the library's internals being deeply type checked, and tests even less so. So please let's simplify this as much as possible.

elftools/elf/dynamic.py
pmhahn: See docstring.

elftools/elf/hash.py
pmhahn: See docstring.

test/test_hash.py
pmhahn: `typeguard` is no hard-dependency; that's why I've added that `try: import typeguard; except ImportError: fallback` thing so you can just use `elftools` as before.

It's similar to all those `if TYPE_CHECKING: import …`s which only add internal dependencies, when you run any type-checker. `typeguard` is just another one, but in contrast to `mypy`, `pyright`, `pyrefly`, `ty` being **static** – they only look at the code but do not execute it – `typeguard` is **dynamic** and checks types **while executing the code**. That's a great way to **validate** the type hints as there were many cases, where I started with – for example – `bytes` but than had it to change it to `list[int]` or add a `| None`, as reality was different from reading just the code and guessing things.

As soon as you extend `pyelftools` and add new functions/methods, you again will have the problem to add **and validate** new type hints; running the test suite with `typeguard` is a great way to get this almost for free – actually with (only) a drastic performance drop as all argument and return-values are validated, which takes time.

Now that `ELFHashTable.__init__()` is typed, your (or anybodies) editor will tell you that passing `None` to `elffile` (1st) and `symboltable` (4th argument) is invalid from a type-perspective. In this case its okay, but you don't want to add ` | None` to those 2 argument just to have the tests clean as that would required checking for those `None`s in may other locations.

While you could use `elffile = symboltable = cast('Any', None)` or add a `# type: ignore` there to silence the type-checkers, sadly that's not sufficient for `typeguard` as it explicitly checks, if the `Protocol` is implemented:
```console
$ python3 -m unittest discover . -k test.test_hash.TestELFHash.test_empty_table_without_header
E
======================================================================
ERROR: test_empty_table_without_header (test.test_hash.TestELFHash.test_empty_table_without_header)
Verify we can handle an empty (0 byte) ELF hash section.
----------------------------------------------------------------------
  File "test/test_hash.py", line 67, in test_empty_table_without_header
    empty_hash_section = ELFHashTable(elffile, 0, 0, symboltable)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "elftools/elf/hash.py", line 44, in __init__
    def __init__(
  File ".venv/lib/python3.12/site-packages/typeguard/_functions.py", line 180, in check_argument_types_internal
    check_type_internal(value, annotation, memo)
  File ".venv/lib/python3.12/site-packages/typeguard/_checkers.py", line 989, in check_type_internal
    checker(value, origin_type, args, memo)
  File ".venv/lib/python3.12/site-packages/typeguard/_checkers.py", line 862, in check_protocol
    raise TypeCheckError(
typeguard.TypeCheckError: argument "symboltable" (None) is not compatible with the _SymbolTable protocol because it has no method named 'get_symbol'

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (errors=1)
```
As `unittest.mock.MagicMock` is part of any standard Python since several years and it only needs to prototype to fake, importing `_SymbolTable` here to satisfy both static and dynamic type checkers here seems like a minor price to pay.

pmhahn:
:partying_face:

k4lizen:
yayy

thank you @pmhahn  and @eliben :D

k4lizen:
could we get a new release cut with these changes?

eliben:
We can cut a new release if this effort is done. @pmhahn anything else remaining?

pmhahn:
> We can cut a new release if this effort is done. @pmhahn anything else remaining?

@eliben No, all done.

eliben:
Please follow https://github.com/eliben/pyelftools/issues/660 for a new release, and use it to report any issues

PR #613, opened 2025-04-02
Added support for symbol relocation for RISC-V architecture

adjordjevic-TT:
Added support for symbol relocation for RISC-V architecture.

Added ENUM_RELOC_TYPE_RISCV to enums.py

Added _RELOCATION_RECIPES_RISCV, two more relocation functions (_reloc_calc_value_plus_sym_6 and _reloc_calc_value_minus_sym_addend_6) and case when e-machine is RISC-V to relocation.py.


adjordjevic-TT:
@eliben Can you take a look at this?

eliben:
@sevaa any concerns?

sevaa:
Looks in line with how relocation is handled for other architectures. Is there any way to test this?

sevaa:
An erroneously written piece of relocation logic most likely will not crash; it will produce wrong bits in the loaded section. How does one test that?

adjordjevic-TT:
Ideally, we can create elf that uses RISCV relocations with some functions/variables and if we are able to find those symbols relocations are working properly, but I don't think it is practical to test every sing one relocation type. Everything that is implemented is taken from here: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#reloc-table

sevaa:
If @eliben says it's OK not to test this exhaustively, I will be OK with that :)

adjordjevic-TT:
We are using it for finding variable's address or value and for implementing backtracing (callstack) for kernels that are running on our hardware that is based on RISCV arch.

sevaa:
The only sections that are parsed, dumped, and compared by the readelf test are the DWARF ones and notes; in executable/SO binaries, relocations against those are rare (none in the corpus, last time I checked). Were we to dump and compare the disassembly, though, or the initialized global data section, that would be the test.

eliben:
> The only sections that are parsed, dumped, and compared by the readelf test are the DWARF ones and notes; in executable/SO binaries, relocations against those are rare (none in the corpus, last time I checked). Were we to dump and compare the disassembly, though, or the initialized global data section, that would be the test.

Let's continue this discussion in https://github.com/eliben/pyelftools/discussions/601

eliben:
@adjordjevic-TT the gap here seems pretty small - can we converge on an acceptable PR?

PR #614, opened 2025-04-02
Document running type checkers

pmhahn:
As requested by https://github.com/eliben/pyelftools/pull/609#issuecomment-2772484676 add
1. basic configuration to run `mypy` and `pyright` without giving extra arguments
2. document how to install and run them


eliben (approved):
Thanks!

PR #620, opened 2025-06-12
Handle hash sections with length 0

henzef:
When parsing an ELF file with an empty hash section (length 0), an exception is raised in the ELFHashTable constructor while trying to parse the struct.
This is fixed by checking if the section is empty before creating the ELFHashSection object. Note that this still leaves room for failure if the section is e.g. 1 byte long, but such malformed files should just raise an exception.


sevaa:
This way, the return value of `_make_section` becomes `None` -  I don't like that on general grounds. Can you make the same check downstream and return an empty `ELFHashSection` (i. e. one with blank `ELFHashTable.params`) instead?

Unless @eliben decrees otherwise, naturally :)

sevaa:
Yeah, this is what I had in mind.

eliben (approved):

PR #622, opened 2025-06-27
scripts: readelf: catch BrokenPipeError

Louson:
readelf failed when the output is a pipe which has been closed by the remote.

Force the flush after display to be sure we don't have anything buffered before catching the error and follow python guidelines to catch the BrokenPipeError and redirect remaining output to devnull.

Example: readelf.py -s /sbin/init | echo 'drop'

### DIFF
--- a/scripts/readelf.py
+++ b/scripts/readelf.py
@@ -8,6 +8,7 @@
 # This code is in the public domain
 #-------------------------------------------------------------------------------
 import argparse
+import os
 import sys
 import re
 import traceback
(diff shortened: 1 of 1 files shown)


scripts/readelf.py
```diff
@@ -1959,12 +1962,19 @@ def main(stream=None):
...
             if args.debug_dump_what:
                 readelf.display_debug_dump(args.debug_dump_what)
+            output.flush()
         except ELFError as ex:
             sys.stdout.flush()
             sys.stderr.write('ELF error: %s\n' % ex)
             if args.show_traceback:
                 traceback.print_exc()
             sys.exit(1)
+        except BrokenPipeError:
```
eliben, line 1972: Presumably this follows the guideline at https://docs.python.org/3/library/signal.html#note-on-sigpipe

Why is the exit code here 0 rather than 1?

eliben (changes requested):

scripts/readelf.py
Louson: It seemed to me this is not an error but a normal behaviour. That said I have checked readelf from binutils and noticed it returns 141. What value would you like ?

PR #626, opened 2025-09-07
Limit compilation units cache size

mohamedelkony:
# Issue:
Iterated CUs (along side with each CU's DIEs cache) are kept in cache as long as DARWF object is alive, While cache can be useful for hitting DIEs shared be CUs and other use cases

A simple loop over CUs and their DIEs keeps all DIEs of ELF stored in memory until DWARF object is no longer referenced.
For a big ELF file ours is ~100 MBs this causes all DIEs of all CUs to stay in memory while we only need to parse DIEs of current CU and then we no longer need to keep those DIE in memory leading to memory leak as we will never use those DIEs on previous CUs (except for interleaved DIE case).

This leads to excessive memory usage it reaches 7 GBs, We parse 5 files in parallel this leads to 35 GBs memory usage

# Fix:

Add possibly to set max cache size on cached CU, Purge Chace FIFO on reaching max size
Kept behavior same as before cache size is unlimited

# Results:
Reduced memory usage by 90%

Memory usage for our 100MB ELF file 
  #### Before: 4GB
image

  #### After: 350 MBs
image

  #### Used test case:
image

### DIFF
--- a/elftools/dwarf/dwarfinfo.py
+++ b/elftools/dwarf/dwarfinfo.py
@@ -82,7 +82,8 @@ class DWARFInfo:
             debug_rnglists_sec,
             debug_sup_sec,
             gnu_debugaltlink_sec,
-            debug_types_sec
+            debug_types_sec,
+            max_cu_cache_size = -1
             ):
         """ config:
                 A DwarfConfig object
@@ -91,6 +92,9 @@ class DWARFInfo:
                 DebugSectionDescriptor for a section. Pass None for sections
                 that don't exist. These arguments are best given with
                 keyword syntax.
+            
+            max_cu_cache_size:
+                Enforces a limit on CU cache size, For unlimitted  cache size set to -1
         """
         self.config = config
         self.debug_info_sec = debug_info_sec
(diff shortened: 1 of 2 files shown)


mohamedelkony:
Hi @eliben Could you please have a look? Thanks!

elftools/dwarf/dwarfinfo.py
```diff
@@ -91,6 +92,9 @@ def __init__(self,
                 DebugSectionDescriptor for a section. Pass None for sections
                 that don't exist. These arguments are best given with
                 keyword syntax.
+            
+            max_cu_cache_size:
+                Enforces a limit on CU cache size, For unlimitted  cache size set to -1
```
eliben, line 97: The size limit is on the number of entries/CUs? The comment should say this in more detail.

eliben:
@sevaa WDYT?

grahamroff-dev:
I propose an alternate solution to this in #669 that avoids changing the constructor, removes the difficulty of picking an appropriate max cache size, and still provides the reduction in maximum memory use when iterating over the compile units.

PR #636, opened 2026-04-29
Typing fixes and improvements to construct

pmhahn:
This are the typing fixes related to the vendored `elftools/construct/` from https://github.com/eliben/pyelftools/pull/611#issuecomment-4337985938 — all of them have been found while working on making `mypy`/`pyright`/`pyrefly`/`ty` happy
- remove wrong and unneeded type-annotation for `__slots__`
- some minor fixes and improvements to existing type-annotation, which have been found wrong or missing
- extend `type: ignore` to equivalent `ty: ignore`
- re-implement `__getstate__()` to handle MRO and all variants of `__slots__`
- re-implement `__eq__()` to also check the type
- modernize some pre-Python-3.9 type hints not that Python>=3.10 is required
- fixed decorator `recusion_lock` to use `functools.wraps` to handle all details of wrapping.

### DIFF
--- a/elftools/construct/adapters.py
+++ b/elftools/construct/adapters.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, Callable, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 from .core import Adapter, AdaptationError, Pass
 from .lib import int_to_bin, bin_to_int, swap_bytes
 from .lib import FlagsContainer, HexString
 
 if TYPE_CHECKING:
-    from collections.abc import Hashable, Mapping, Sized
+    from collections.abc import Callable, Hashable, Mapping, Sized
 
     from .core import Construct, _Pass
     from .lib import Container, ListContainer
--- a/elftools/construct/adapters.py
+++ b/elftools/construct/adapters.py
@@ -256,7 +256,8 @@ class CStringAdapter(StringAdapter):
         self.terminators = terminators
     def _encode(self, obj: bytes | str, context: Container) -> bytes:
         return StringAdapter._encode(self, obj, context) + self.terminators[0:1]
-    def _decode(self, obj: list[bytes], context: Container) -> bytes | str:  # type: ignore[override]
+    def _decode(self, obj: list[bytes], context: Container) -> bytes | str:  # type: ignore[override] # ty: ignore[invalid-method-override]
+        # This violates the Liskov Substitution Principle: should be `obj: bytes`, but RepeatUntil() converts `bytes` to `list[byte]`
         return StringAdapter._decode(self, b''.join(obj[:-1]), context)
 
 class TunnelAdapter(Adapter):
--- a/elftools/construct/adapters.py
+++ b/elftools/construct/adapters.py
@@ -27,15 +27,15 @@ __all__ = [
 # exceptions
 #===============================================================================
 class BitIntegerError(AdaptationError):
-    __slots__: list[str] = []
+    __slots__ = ()
 class MappingError(AdaptationError):
-    __slots__: list[str] = []
+    __slots__ = ()
 class ConstError(AdaptationError):
-    __slots__: list[str] = []
+    __slots__ = ()
 class ValidationError(AdaptationError):
-    __slots__: list[str] = []
+    __slots__ = ()
 class PaddingError(AdaptationError):
-    __slots__: list[str] = []
+    __slots__ = ()
 
 #===============================================================================
 # adapters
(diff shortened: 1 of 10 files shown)


elftools/construct/adapters.py
```diff
@@ -250,13 +250,13 @@ class CStringAdapter(StringAdapter):
...
     """
-    __slots__: list[str] = ["terminators"]
+    __slots__ = ("terminators",)
     def __init__(self, subcon: Construct, terminators: bytes = b"\x00", encoding: str | None = None) -> None:
         StringAdapter.__init__(self, subcon, encoding = encoding)
         self.terminators = terminators
     def _encode(self, obj: bytes | str, context: Container) -> bytes:
         return StringAdapter._encode(self, obj, context) + self.terminators[0:1]
-    def _decode(self, obj: list[bytes], context: Container) -> bytes | str:  # type: ignore[override]
+    def _decode(self, obj: list[bytes], context: Container) -> bytes | str:  # type: ignore[override] # ty: ignore[invalid-method-override]
```
eliben, line 259: Is this something we can/should fix instead of ignoring?

If we still opt to ignore, a short comment explaining _why_ would be helpful.

elftools/construct/core.py
```diff
@@ -715,8 +712,8 @@ class Sequence(Struct):
         UBInt8("third_element"),
     )
     """
-    __slots__: list[str] = []
-    def _parse(self, stream: IO[bytes], context: Container) -> ListContainer:  # type: ignore[override]
+    __slots__ = ()
+    def _parse(self, stream: IO[bytes], context: Container) -> ListContainer:  # type: ignore[override] # ty: ignore[invalid-method-override]
```
eliben, line 716: ditto

elftools/construct/adapters.py
pmhahn: That's easy to answer: Just remove the comment and run
```console
$ ty check elftools/construct/adapters.py
error[invalid-method-override]: Invalid override of method `_decode`
   --> elftools/construct/adapters.py:259:9
    |
259 |     def _decode(self, obj: list[bytes], context: Container) -> bytes | str:
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Definition is incompatible with `StringAdapter._decode`
    |
   ::: elftools/construct/adapters.py:173:9
    |
173 |     def _decode(self, obj: bytes, context: Container) -> bytes | str:
    |         ------------------------------------------------------------ `StringAdapter._decode` defined here
    |
info: parameter `obj` has an incompatible type: `bytes` is not assignable to `list[bytes]`
info: This violates the Liskov Substitution Principle
```
See [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle): any subclass must handle things like the parent-class, which expects `bytes` here. But the subclass expects `list[bytes]`, which is not the same.

But if I do this change:
```diff
diff --git a/elftools/construct/adapters.py b/elftools/construct/adapters.py
index f3ca6ae..d6b2368 100644
--- a/elftools/construct/adapters.py
+++ b/elftools/construct/adapters.py
@@ -256,8 +256,8 @@ class CStringAdapter(StringAdapter):
         self.terminators = terminators
     def _encode(self, obj: bytes | str, context: Container) -> bytes:
         return StringAdapter._encode(self, obj, context) + self.terminators[0:1]
-    def _decode(self, obj: list[bytes], context: Container) -> bytes | str:  # type: ignore[override] # ty: ignore[invalid-method-override]
-        return StringAdapter._decode(self, b''.join(obj[:-1]), context)
+    def _decode(self, obj: bytes, context: Container) -> bytes | str:
+        return StringAdapter._decode(self, obj[:-1], context)
 
 class TunnelAdapter(Adapter):
     """
```
running the tests fails because of `elftoos/construct/code.py:563 RepeatUntil._parse(stream: IO[bytes], …) -> list[Any]:` Each input byte is **individually** parsed by one `Fields("chars", 1)` and converted into a single byte: So you get `list[bytes] = [b"a", b"b", b"c", b"\0"]` instead of a single (large) `bytes = b"abc\0"`, which `StringAdapter._parse()` then must handle.

Summary: yes, this definitely needs fixing and needs more work, but I don't see an easy fix right now. So for now I have added the `# type: ignore`.

elftools/construct/core.py
pmhahn:
```console
$ ty check
error[invalid-method-override]: Invalid override of method `_parse`
   --> elftools/construct/core.py:716:9
    |
716 |     def _parse(self, stream: IO[bytes], context: Container) -> ListContainer:
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Definition is incompatible with `Struct._parse`
    |
   ::: elftools/construct/core.py:656:9
    |
656 |     def _parse(self, stream: IO[bytes], context: Container) -> Container:
    |         ---------------------------------------------------------------- `Struct._parse` defined here
    |
info: incompatible return types: `ListContainer` is not assignable to `Container`
info: This violates the Liskov Substitution Principle
```
Sadly `Container` and `ListContainer` are unrelated, e.g. `ListContainer` is **not** a sub-class of `Container`.

eliben (commented):
got it, ok so just add a brief comment to each and we'll merge this, with an eye to maybe fixing it more comprehensively later

eliben (approved):

PR #637, opened 2026-04-29
Convert to enum.IntEnum

pmhahn:
This is the part, which converts `elftools/dwarf/constants.py` to use `enum.IntEnum` from https://github.com/eliben/pyelftools/pull/611#issuecomment-4337985938
- this will allow future type improvements as those `IntEnum`-classes can be used instead of the more generic `int` in many places
- It also moves some code related to `DW_CFA` from `callframe.py` to `constants.py` to handle the nitty-gritty details of the encoding
- It converts several large `if-elfi` sequences to _structural pattern matching_, which makes the code a little bit nicer to read and understand. (it also allows to detect the _unhandled_ cases more easily)

There is more potential future work afterwards:
- There are many cases, where `DW_CFA`-constants are compared by their `str`-representation; this could be converted to `Enum`-comparison
- Part from `elftools/dwarf/descriptions.py` could be moved to `constants.py` to better keep both in-sync.

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -14,7 +14,7 @@ from ..common.utils import (
 from ..construct import Struct, Switch
 from .enums import DW_EH_encoding_flags
 from .structs import DWARFStructs
-from .constants import *
+from .constants import DW_CFA
 
 
 class CallFrameInfo:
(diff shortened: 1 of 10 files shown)


elftools/elf/enums.py
```diff
@@ -642,10 +642,11 @@
...
 # Here is the full combined mapping from tag name to value
 
-ENUM_D_TAG = dict(ENUM_D_TAG_COMMON)
-ENUM_D_TAG.update(ENUM_D_TAG_SOLARIS)
-for k in ENUMMAP_EXTRA_D_TAG_MACHINE:
-    ENUM_D_TAG.update(ENUMMAP_EXTRA_D_TAG_MACHINE[k])
+ENUM_D_TAG = {
+    **ENUM_D_TAG_COMMON,
+    **ENUM_D_TAG_SOLARIS,
+    **{k: v for kv in ENUMMAP_EXTRA_D_TAG_MACHINE.values() for k, v in kv.items()},
```
eliben: This is much less readable than the original, IMO. Any chance of improving? Or adding a comment explaining what this is doing?

elftools/dwarf/constants.py
```diff
@@ -6,225 +6,306 @@
...
+    type = 0x02
+    partial = 0x03
+    skeleton = 0x04
+    split_compile = 0x05
+    split_type = 0x06
+    lo_user = 0x80
+    hi_user = 0xff
+
+
+# Add back constants with legacy names; invisible to typing!
```
eliben: expand the comment to explain which constants are added, just a couple of examples

does it matter that this is invisible to typing? AFAIU they're not used in the codebase anyway (just for external backwards compat)?

elftools/dwarf/constants.py
```diff
@@ -6,225 +6,306 @@
...
+globals().update({
+    f"{enum_name}_{member_name}": member.value
+    for enum_name, enum in globals().items()
+    if enum_name.startswith("DW_") and issubclass(enum, Enum)
+    for member_name, member in enum.__members__.items()
+})
+
+
+# Currently unused as this is only called for unknown keys
```
eliben: whoa, do we really need this?

pmhahn: No, its an alternative to the code above to add back the legacy names dynamically. Legacy code can still use `DW_UT_type` as `__getattr__(key)` **only** gets called, when `key` does not exist. It will then try to split key into the enum name `DW_UT` and the member-name `type` and will return that — or otherwise raise an `AttributeError()` if either there is no `enum` or `member` by that name.

The drawback of this variant is, that `from elftools.dwarf.constants import *` will no longer work.

The bonus of this variant is, that you will get a `DeprecationWarning` for each use of the legacy names and get a chance to update your code :wink: 

I can remove one or the other, but keeping this does not hurt.

So you/me might go with documenting the API change first, wait some months, switch to variant 2, wait some more month and then finally remove the legacy API all together.

elftools/elf/enums.py
pmhahn: A later patch in #611 will add a `Mapping[str, int]` to it to mark it as _read-only_. For this to work the `dict()` must be built in "one go", e.g. repeatedly updating it via `ENUM_D_TAG.update(…)` does not work. And `dict`-concatenation only was added in [3.9](https://peps.python.org/pep-0584/), so this is the portable version working in even older versions. (I really started working on this long ago …)

Reverting this commit locally worked fine with `mypy`/`pyright`/`pyrefly`/`ty`, but I still prefer to add the _read-only_ later. As such I switched to using `merge_dicts()` which several other `ENUM`s already do.

elftools/dwarf/constants.py
pmhahn: Changed to
> Add back constants with legacy names, e.g. DW_UT_type = DW_UT.type; invisible to typing!

**Static** type checkers will tell you, that there is no `DW_UT_type` as that would require **executing** these commands. You should use `DW_UT.type` instead.

Yes, just for backward compatibility so code using `DW_UT_type` will not break (yet).

elftools/dwarf/constants.py
eliben: got it, ok let's remove this and just keep the code above.

pmhahn: I have removed `__getattr__()` and force-pushed my branch.

elftools/dwarf/callframe.py
```diff
@@ -447,11 +445,8 @@ def _eh_encoding_to_field(entry_structs):
 def instruction_name(opcode):
     """ Given an opcode, return the instruction name.
     """
-    primary = opcode & _PRIMARY_MASK
-    if primary == 0:
-        return _OPCODE_NAME_MAP[opcode]
-    else:
-        return _OPCODE_NAME_MAP[primary]
+    warn("Use DW_CFA()._fqn instead", DeprecationWarning, stacklevel=2)
```
eliben: Now on another look, it's odd to have this warning here, given that `_fqn` is a non-public, undocumented property. 
If we want to recommend using properties like `_op` or `_fqn`, they should be documented? The `_` prefix is problematic, but I see why we'd want it for an enum. Could we use `@enum.property` to have nice, not-`_`-starting names?

pmhahn: Sadly [enum.property](https://docs.python.org/3.11/whatsnew/3.11.html#enum) is Python 3.11, but we still support 3.10, which has [EoL 2026-10](https://devguide.python.org/versions/). Therefore I renamed the property to `FQN` for now.

I also changed the Type from `IntFlag` to `IntEnum` as it is not a flag, but a _strange_ enum where the lower bits are used to encode the _argument_ for some members. Therefor there now is a `DW_CFA.parse_raw_opcode()` to separate the `opcode` from the optional `argument`. I hope that's nice enough.

eliben (approved):

PR #639, opened 2026-04-29
Typing unrelated fixes

pmhahn:
While working on https://github.com/eliben/pyelftools/pull/611#issuecomment-4337985938 I found several potential bugs or issues, which complicate type-checking
- using wrong format string
- missing check for `None`
- checking `Union` for _wrong_ branch
- not returning `None` explicitly
- not raising `ValueError` in all cases

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -187,6 +187,8 @@ class Dynamic:
                 self._num_tags = n + 1
                 return self._num_tags
 
+        return None
+
     def get_relocation_tables(self):
         """ Load all available relocation tables from DYNAMIC tags.
 
--- a/elftools/elf/hash.py
+++ b/elftools/elf/hash.py
@@ -63,7 +63,7 @@ class ELFHashTable:
         symndx = self.params['buckets'][hval]
         while symndx != 0:
             sym = self._symboltable.get_symbol(symndx)
-            if sym.name == name:
+            if sym and sym.name == name:
                 return sym
             symndx = self.params['chains'][symndx]
         return None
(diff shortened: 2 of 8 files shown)


elftools/elf/dynamic.py
```diff
@@ -197,21 +199,21 @@ def get_relocation_tables(self):
         result = {}
 
         if list(self.iter_tags('DT_REL')):
-            result['REL'] = RelocationTable(self.elffile,
+            result['REL'] = rel = RelocationTable(self.elffile,
```
eliben: is this fixing a bug?

elftools/elf/hash.py
```diff
@@ -111,15 +111,14 @@ class GNUHashTable:
     def __init__(self, elffile, start_offset, symboltable):
         self.elffile = elffile
         self._symboltable = symboltable
-        self.params = struct_parse(self.elffile.structs.Gnu_Hash,
-                                   self.elffile.stream,
-                                   start_offset)
+        p = struct_parse(self.elffile.structs.Gnu_Hash, self.elffile.stream, start_offset)
+        self.params = p
```
eliben: same question; why this change?

pmhahn: I have dropped this now that `elftools/construct/lib/container.py:68 Container.__getitem__()` knows about `bloom_size` and `nbuckets`.

elftools/elf/dynamic.py
pmhahn: It's storing the just created `RelocationTable()` instance in a local variable so it can be referenced with exactly type later.

`result` will later get typed as `dict[str, RelocationTable | RelrRelocationTable]`, so when you get something from it, it might be both types and the code needs to use `instanceof()` to check, which it is. The local variable remembers the exact type and you're saved from using `instanceof()`.

Thinking more about this: I can drop this and instead add the following new type #611:
```python
class RelocationTables(TypedDict, total=False):
    REL: RelocationTable
    RELA: RelocationTable
    RELR: RelrRelocationTable
    JMPREL: RelocationTable
```

pmhahn: I have dropped the commit here and have added the new `TypedDict` to my `typing` branch.

eliben (approved):

Issue #286, opened 2020-02-08
Why aren't DW_FORM_blockX attributes returned as bytes?


sevaa:
Those come across as lists of ints.

mdmillerii:
Bytes are new to python3.

sevaa:
`bytearray` on Python 2, then. No different from `bytes`, except mutable. They index the same as int arrays, unlike strings.

eliben:
@sevaa another instance of -- what to do about this now that we aren't beholden to Python 2?

sevaa:
This, I think, can be pulled off. It will break some code (e. g. explicit `isinstance(value, list)` checks to detect blob values), but won't break good news programs that don't check the type.

PR #640, opened 2026-05-01
construct: Harmonize remaining __slots__

pmhahn:
This are the typing fixes related to the vendored `elftools/construct/` from https://github.com/eliben/pyelftools/pull/611#issuecomment-4337985938 — all of them have been found while working on making `mypy`/`pyright`/`pyrefly`/`ty` happy

-  Convert the remaining `__slots__` of 4 `class`es in 2 files from `list` to `tuple` for consistency – forgotten on #636

### DIFF
--- a/elftools/construct/debug.py
+++ b/elftools/construct/debug.py
@@ -36,10 +36,10 @@ class Probe(Construct):
         UBInt8("b"),
     )
     """
-    __slots__ = [
+    __slots__ = (
         "printname", "show_stream", "show_context", "show_stack",
         "stream_lookahead"
-    ]
+    )
     counter = 0
 
     def __init__(self, name: str | None = None, show_stream: bool = True,
@@ -108,7 +108,7 @@ class Debugger(Subconstruct):
         )
     )
     """
-    __slots__ = ["retval"]
+    __slots__ = ("retval",)
     def _parse(self, stream: IO[bytes], context: Container) -> Any:
         try:
             return self.subcon._parse(stream, context)
--- a/elftools/construct/lib/container.py
+++ b/elftools/construct/lib/container.py
@@ -137,7 +137,7 @@ class ListContainer(list[Any]):
     A container for lists.
     """
 
-    __slots__ = ["__recursion_lock__"]
+    __slots__ = ("__recursion_lock__",)
 
     @recursion_lock("[...]")
     def __str__(self) -> str:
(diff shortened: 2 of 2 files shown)


eliben (commented):
Is there a mismatch between PR description and contents?

All I see here is conversion of slots to tuples instead of lists

pmhahn:
> Is there a mismatch between PR description and contents?
> 
> All I see here is conversion of slots to tuples instead of lists

Yes, sorry for the bad text: I re-used the previous commit removing the types as they were part of it initially, but they somehow got dropped on one of my many rebased/splits.

I've reworded the description of this PR.

eliben (approved):

sevaa:
FYI, run a quick autotest against my stuff, the only missing bit/breaking change is `common.utils.bytelist2string()`.

pmhahn:
> `common.utils.bytelist2string`

@sevaa Can you please provide more details: Neither is there a `common.utils` nor a `bytelist2string()`; Probably `elftools/common/utils.py` and either `bytes2str()` or `bytes2hex()`, but neither was touched by this specific PR. So I'm confused by your comment.

sevaa:
https://github.com/eliben/pyelftools/blob/v0.32/elftools/common/utils.py#L26

eliben:
This function was removed in f92bb714a90cc747d70e963608c9f927874327f4

I think it's actually OK... we've removed a bunch of these old Python 2 compatibility functions already in the past; generally user code isn't supposed to depend on them anyway, and it's a very easy fix if someone does. WDYT?

pmhahn:
> This function was removed in [f92bb71](https://github.com/eliben/pyelftools/commit/f92bb714a90cc747d70e963608c9f927874327f4)

Thank you for looking that up.

> I think it's actually OK... we've removed a bunch of these old Python 2 compatibility functions already in the past; generally user code isn't supposed to depend on them anyway, and it's a very easy fix if someone does. WDYT?

I with you: Users should not depend on the _internal API_.

sevaa:
How was that an internal API? I don't see an underscore there.

I agree (#286) that those arrays of integers should have been `bytes` ~~all along~~ since Pyhton 2 was dropped.

eliben:
> How was that an internal API? I don't see an underscore there.

Well, it's a _conceptually_ internal API :)

pyelftools is very old code, and predates my full understanding of what/how to hide. The point is that we've already removed a bunch of these old Py2 compatibility helpers; would you prefer we keep `bytelist2string` specifically just for backwards compatibility?

PR #641, opened 2026-05-05
More cleanups before adding type hints

pmhahn:
More from - fix some already existing types https://github.com/eliben/pyelftools/pull/611#issuecomment-4363868618
- fix some `printf` like output to handle all types, e.g. `None` or _not_ `int`
- rename some variables or change the code where variables are re-defined with a different type, which some type-checkers are allergic to.
- add many `assert` to make checks explicit where variables may be `None` or have multiple types
- Re-implement Attribute instantiation

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -135,6 +135,8 @@ class CallFrameInfo:
 
         # If the augmentation string is not empty, hope to find a length field
         # in order to skip the data specified augmentation.
+        lsda_pointer = None
+        aug_dict = None
         if is_CIE:
             aug_bytes, aug_dict = self._parse_cie_augmentation(
                     header, entry_structs)
@@ -147,8 +149,6 @@ class CallFrameInfo:
                 lsda_pointer = self._parse_lsda_pointer(entry_structs,
                                                         self.stream.tell() - len(aug_bytes),
                                                         lsda_encoding)
-            else:
-                lsda_pointer = None
 
         # For convenience, compute the end offset for this entry
         end_offset = (
@@ -495,6 +495,7 @@ class CFIEntry:
         """ Decode the instructions contained in the given CFI entry and return
             a DecodedCallFrameTable.
         """
+        last_line_in_CIE = None
         if isinstance(self, CIE):
             # For a CIE, initialize cur_line to an "empty" line
             cie = self
--- a/elftools/dwarf/datatype_cpp.py
+++ b/elftools/dwarf/datatype_cpp.py
@@ -60,8 +60,11 @@ def parse_cpp_datatype(var_die):
             ptr_prefix = ''
 
         if t.tag == 'subroutine':
-            params = tuple(format_function_param(p, p) for p in type_die.iter_children() if p.tag in ("DW_TAG_formal_parameter", "DW_TAG_unspecified_parameters") and 'DW_AT_artificial' not in p.attributes)
-            params = ", ".join(params)
+            params = ", ".join(
+                format_function_param(p, p)
+                for p in type_die.iter_children()
+                if p.tag in ("DW_TAG_formal_parameter", "DW_TAG_unspecified_parameters") and 'DW_AT_artificial' not in p.attributes
+            )
             if 'DW_AT_type' in type_die.attributes:
                 datatype = parse_cpp_datatype(type_die)
                 is_pointer = datatype.modifiers and datatype.modifiers[-1] == 'pointer'
--- a/elftools/dwarf/datatype_cpp.py
+++ b/elftools/dwarf/datatype_cpp.py
@@ -63,9 +63,9 @@ def parse_cpp_datatype(var_die):
             params = tuple(format_function_param(p, p) for p in type_die.iter_children() if p.tag in ("DW_TAG_formal_parameter", "DW_TAG_unspecified_parameters") and 'DW_AT_artificial' not in p.attributes)
             params = ", ".join(params)
             if 'DW_AT_type' in type_die.attributes:
-                retval_type = parse_cpp_datatype(type_die)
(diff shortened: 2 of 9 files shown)


elftools/dwarf/descriptions.py
```diff
@@ -119,16 +120,22 @@ def _full_reg_name(regnum):
 def describe_CFI_register_rule(rule):
     s = _DESCR_CFI_REGISTER_RULE_TYPE[rule.type]
     if rule.type in ('OFFSET', 'VAL_OFFSET'):
+        assert isinstance(rule.arg, int)
```
eliben: Would we able to avoid these checks one type annotations are introduced? `descriptions.py` does a lot of formatting, it'd be a shame to see these type checks in all places

elftools/dwarf/descriptions.py
```diff
@@ -119,16 +120,22 @@ def _full_reg_name(regnum):
...
         s += '%+d' % rule.arg
     elif rule.type == 'REGISTER':
-        s += describe_reg_name(rule.arg)
+        assert isinstance(rule.arg, int)
+        reg = describe_reg_name(rule.arg)
```
eliben: A bit concerned about this one too: today it will print None, after the change it will crash for minor errors.

Wouldn't it be better to have describe_reg_name return some '<error>' string in case something really unexpected happens?

elftools/dwarf/descriptions.py
```diff
@@ -119,16 +120,22 @@ def _full_reg_name(regnum):
...
+        assert reg is not None
+        s += reg
     return s
 
 
 def describe_CFI_CFA_rule(rule):
     if rule.expr:
         return 'exp'
     else:
+        assert isinstance(rule.reg, int)
```
eliben: why are these needed to call describe_reg_name?

eliben (commented):
Started leaving comments, but this applies to a lot more: does it make sense to review these without the type annotations? So many defensive checks - they are difficult to assess in context without seeing the types they're trying to guard.

pmhahn:
> Started leaving comments, but this applies to a lot more: does it make sense to review these without the type annotations? So many defensive checks - they are difficult to assess in context without seeing the types they're trying to guard.

I have found all of them **after** adding type hints, which makes them hard to validate without them.
If you prefer I can move those 2 commits back to #611

elftools/dwarf/descriptions.py
pmhahn: Here's the type annotation from #611: 
```python
class RegisterRule:
    def __init__(self, type: str, arg: int | None = None) -> None:
```

The problem here is, that `arg` is **optional** and may be `None` (which is the default).
So a `assert rule.reg is not None` is needed before `describe_reg_name()` is called as that will return `rNone` or crash, depending on your architecture:
```python
>>> from elftools.dwarf.descriptions import describe_reg_name
>>> describe_reg_name(None)
'rNone'
>>> describe_reg_name(None, 'x86')
  File "<stdin>", line 1, in <module>
  File "…/elftools/dwarf/descriptions.py", line 180, in describe_reg_name
    return _REG_NAMES_x86[regnum]
           ~~~~~~~~~~~~~~^^^^^^^^
TypeError: list indices must be integers or slices, not NoneType
```

An alternative would be to remove `arg` from `RegisterRule` itself and create a `class RegisterRuleWithArgument(RegisterRule)`, where `arg: int` is required.

elftools/dwarf/descriptions.py
pmhahn: Depends on what you want: Returning `""` here (or `rNone`) hides the underlying issue. By returning `None` here you **force** the caller to check for that case explicitly.

PS: I we can get rid of those  `assert describe_reg_name(…) is not None` by adding a [typing.overload](https://docs.python.org/3/library/typing.html#typing.overload) to it as `None` is only ever returned for `default=False`:
```python
@overload
def describe_reg_name(regnum: int, machine_arch: str | None, default: Literal[False]) -> str | None: ...
@overload
def describe_reg_name(regnum: int, machine_arch: str | None = ..., default: Literal[True] = ...) -> str: ...
def describe_reg_name(regnum: int, machine_arch: str | None = None, default: bool = True) -> str | None:
```

elftools/dwarf/descriptions.py
pmhahn: `rule.arg` can (potentially) be `None` according to its type `int | None` and `%+d` is invalid for `None`:
```python3
>>> '%+d' % None
TypeError: %d format: a real number is required, not NoneType
```
Here you would get the `TypeError` for free from Python itself, but there are cases where such `TypeErrors` are only detected deep down the stack and then are hard to debug. As such those `assert`s make them explicit and easier to find.

~~PS: Actually the `arg` was wrongly typed as `ListContainer | None`; correct is `int | None` but that `|None` still needs to be handled.~~

eliben:
> > Started leaving comments, but this applies to a lot more: does it make sense to review these without the type annotations? So many defensive checks - they are difficult to assess in context without seeing the types they're trying to guard.
> 
> I have found all of them **after** adding type hints, which makes them hard to validate without them. If you prefer I can move those 2 commits back to #611

Thanks for the detailed responses. Yes, let's move them back.

Peeling off the bit PR into smaller ones will work best by layers, starting with the lowest layer in the call stack (annotations + code changes required to make them pass), and climbing up. `descriptions` could be a good start.

pmhahn:
> Peeling off the bit PR into smaller ones will work best by layers, starting with the lowest layer in the call stack (annotations + code changes required to make them pass), and climbing up. `descriptions` could be a good start.

I've removed the 2 commits from this PR, so this PR should only contain changes, where typing is not yet needed.

I've added one more small patch for `test/run_readelf_tests.py` to print the full command: it was driving me nuts that I always had to re-run it wit `--verbose` to find the failing test file.

eliben (approved):

PR #642, opened 2026-05-07
typing: Convert to NamedTuple

pmhahn:
Convert from legacy `collections.namedtuple` to `typing.NamedTuple` to allow adding type hints. Part of #611

### DIFF
--- a/elftools/dwarf/aranges.py
+++ b/elftools/dwarf/aranges.py
@@ -6,7 +6,8 @@
 # Dorothy Chen (dorothchen@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
-from collections import namedtuple
+from typing import NamedTuple
+
 from ..common.utils import struct_parse
 from bisect import bisect_right
 import math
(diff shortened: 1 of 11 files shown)


elftools/dwarf/aranges.py
```diff
@@ -6,7 +6,10 @@
 # Dorothy Chen (dorothchen@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
-from collections import namedtuple
+from __future__ import annotations
```
eliben: is it actually required here?

eliben (commented):
Very nice improvement in readability!

Just one question about all the future imports, I wasn't sure why it's needed for this specific PR

elftools/dwarf/aranges.py
pmhahn: Not in this file as there are no _forward references_, which otherwise would have to be _stingified_ by putting them into quotes; I have the habit to add the `from __future__ import annotations` to all files I touch as then I never have to think about that afterwards.

eliben: Got it, let's remove them for now - until necessary. They should be easy to add/detect as needed.

This applies to all files in this PR

pmhahn: I have removed them except 2:
- `elftools/dwarf/lineprogram.py`: `class LineProgramEntry@41` has a forward-reference to `class LineState@48`
- `elftools/ehabi/decoder.py`: The nested `class _DECODE_RECIPE_TYPE@249` has a backward-reference to its containing `class EHABIBytecodeDecoder@14`, which is not visible until after the `class` is fully declared.

eliben (approved):

PR #643, opened 2026-05-07
Python typing: elftools.common

pmhahn:
Add type hints to `elftools.common` – part of #611 

The `from __future__ import annotations`' are needed for the following reasons:
- Some types are only defined when `TYPE_CHECKING` is enabled, which is an internal variable all type-checkers set to `True`. This is mostly used to break `import`-cycles, which only type-checkers must handle then, but not regular run-times during normal execution. Also some `TypeVar`s are only relevant for type-checking and thus are defined in an `if TYPE_CHECKING`-block. Referencing them would result in a `NameError`, which the _stringifying_ `__future__.annotations` mitigates.
- [typeguard](https://typeguard.readthedocs.io/en/stable/userguide.html#notes-on-forward-reference-handling) also needs it; otherwise it will fail to resolve those references during regular runtime when it is enabled.

### DIFF
--- a/elftools/common/construct_utils.py
+++ b/elftools/common/construct_utils.py
@@ -6,10 +6,19 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from __future__ import annotations
+
 from struct import Struct
+from typing import IO, TYPE_CHECKING, Any, NoReturn
+
 from ..construct import (
     Subconstruct, ConstructError, ArrayError, SizeofError, Construct, StaticField, FieldError
     )
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+
+    from ..construct import Container
 
 
 class RepeatUntilExcluding(Subconstruct):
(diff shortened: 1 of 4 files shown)


elftools/common/utils.py
```diff
@@ -6,24 +6,43 @@
...
         result.update(d)
     return result
 
-def bytes2str(b):
+def bytes2str(b: bytes) -> str:
     """Decode a bytes object into a string."""
     return b.decode('latin-1')
 
-def struct_parse(struct, stream, stream_pos=None):
+@overload
```
eliben, line 43: Need a comment explaining these overloads

elftools/common/construct_utils.py
```diff
@@ -65,7 +76,9 @@ def _parse(self, stream, context):
 class SLEB128(Construct):
     """A construct based parser for SLEB128 encoding.
     """
-    def _parse(self, stream, context):
+    if TYPE_CHECKING:
```
eliben: Do we even use None-named Constructs anywhere?

In any case, it'd be nice to move this to its own thing, like `NamedConstruct` that adds this annotation; would that work?

elftools/common/utils.py
pmhahn: Added

elftools/common/construct_utils.py
pmhahn: Good idea: I have added a `_NamedConstruct`.

eliben (approved):

PR #644, opened 2026-05-09
More cleanup to elftools.common

pmhahn:
While introducing `_NamedConstruct` in #643 / #611 I noticed some not-typing-related code-smell in `elftools.common` and `test/`:
- `run_dwarfdump_test` should also print the complete failing command
- Simplify `parse_cstring_from_stream()`: get rid of `found`, use `bytes.partition()`
- Refactor `save_dwarf_section()` to use `preserve_stream_pos()`
- Deprecates `bytes2hex()` now that [bytes.hex(sep)](https://docs.python.org/3/library/stdtypes.html#bytes.hex) is available since Python 3.8
- Deprecate `iterbytes()` as `iter(bytes) -> int` anyway.
- Deprecate `read_blob()`: Prefer  `IO.read() -> bytes` or `list(…) -> list[int]` if you really must

Your might directly remove those deprecated and now unused function, but maybe let's not annoy @sevaa again like in https://github.com/eliben/pyelftools/pull/640#issuecomment-4364613744 :wink: 

PS: There might be more cases where _read-only_ `bytes` could be used instead of _mutable_ `int[list]`, but I leave that for another day.


eliben (commented):
Let's just remove these deprecated helpers while we're at it. As I explained in that other PR, I don't consider these to be part of the public API (though agree they should have been hidden better) and we've removed many of them in the past as part of the Python 3 transition.

pmhahn:
> Let's just remove these deprecated helpers while we're at it. As I explained in that other PR, I don't consider these to be part of the public API (though agree they should have been hidden better) and we've removed many of them in the past as part of the Python 3 transition.

Done: removed to :wastebasket:

eliben (approved):

PR #645, opened 2026-05-09
Type hints for elftools.ehabi

pmhahn:
Next part from #611 related to `elftools.ehabi`.


elftools/ehabi/decoder.py
```diff
@@ -8,7 +8,10 @@
 # -------------------------------------------------------------------------------
 from __future__ import annotations
 
-from typing import Callable, NamedTuple
+from typing import TYPE_CHECKING, NamedTuple
+
+if TYPE_CHECKING:
```
eliben: Why is this moved under type checking?

pmhahn: 1. It's good convention to only `import` what's needed as each `import` requires RAM and slows down the startup. 
2. https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/
3. `typing.Callable` is deprecated and `collections.abc.Callable` should be used instead

pmhahn: I have reverted that for now; let's convert to `from collections.abc import …` in some future PR.

eliben (approved):

PR #646, opened 2026-05-13
typing[enum,const]: Add / fix PEP-484 type hints

pmhahn:
Here are `eflfile.{elf,dwarf}.{const,enum}` from #611

I try to walk this ~~tree~~ graph: elftools


eliben (approved):

eliben:
Thank you for the effort of breaking these into reviewable PRs :)
To save you effort I think you can make them larger by 2x or so, but otherwise it's very helpful.

PR #649, opened 2026-05-15
Typing elftools.dwarf

pmhahn:
Next batch from #611 related to `elftools.dwarf.{abbrevtable,aranges,callframe,dwarf_expr,lineprogram,namelut,structs}`. Most is really straight forward except:
- `callframe` is using a mixed `dict[str | int, int | CFARule | RegisterRule]` for `Line`. I define a `Protocol` to get the correct type for `"pc", "cfa", int` as `TypedDict` does not work for `int`, but that requires some ugly `cast()s` from/to `Line`. An alternative would be to use just `dict[Any, Any]`.
- For `Augmentation` see https://github.com/eliben/pyelftools/pull/611#issuecomment-4363868618; Because of that some `cast()`s a are needed.
- `Line` and `Augmentation` must **not** be declared under a `if TYPE_CHECKING` as those types are used to type local variables: `from __future__ import annotations` does not work there and the types must be referencable at runtime.
- In `dwarf_expr` I changed how `DW_OP_name2opcode` gets build to make it "`const`".

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -176,6 +176,7 @@ class CallFrameInfo:
                     header, entry_structs)
         else:
             cie = self._parse_cie_for_fde(offset, header, entry_structs)
+            assert isinstance(cie, CFIEntry)
             aug_bytes = self._read_augmentation_data(entry_structs)
             lsda_encoding = cast(int, cie.augmentation_dict.get('LSDA_encoding', DW_EH_encoding_flags['DW_EH_PE_omit']))
             if lsda_encoding != DW_EH_encoding_flags['DW_EH_PE_omit']:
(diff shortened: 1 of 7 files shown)


elftools/dwarf/callframe.py
```diff
@@ -6,18 +6,63 @@
...
 from .constants import DW_CFA
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from ..construct.core import Construct
+    from ..construct.lib.container import ListContainer
+
+
+class Line(Protocol):
```
eliben: Yeah, this is a lot of twisting just to type one thing. I'd say this is a good case where `dict[Any, Any]` will do for now, with a comment on followup once TypedDict catches up

pmhahn: I've converted it to `dict[Any, Any]` for now.

eliben (approved):

PR #650, opened 2026-05-15
typing: elftools.elf

pmhahn:
More typing for `elftools.elf` from #611
- more or less the boring part as most special cases are exempted
- `elfstools.elf.descriptions` has some issues: "you have a basket (`dict[int, …]`) full of vegetables and you're asking Python, if it contains a car `str`". While that's perfect valid Python code, the type-checkers call you nuts and complain. As such the `# type: ignore` is required to silence them.
- `stringtables` need some special handling; will come later
- `make_relative_loader()` and `load()` need some special handling for `test/test_debuglink.py`; will come later
- `__getitem__()` will need some `@overload`s; will come later
- Also some `Protocols` and `TypeVars` will come later

### DIFF
--- a/elftools/elf/descriptions.py
+++ b/elftools/elf/descriptions.py
@@ -6,6 +6,10 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Final
+
 from .enums import (
     ENUM_D_TAG, ENUM_E_VERSION, ENUM_P_TYPE_BASE, ENUM_SH_TYPE_BASE,
     ENUM_RELOC_TYPE_i386, ENUM_RELOC_TYPE_x64,
@@ -15,6 +19,12 @@ from .enums import (
     ENUM_DT_FLAGS_1, ENUM_RELOC_TYPE_PPC)
 from .constants import (
     P_FLAGS, RH_FLAGS, SH_FLAGS, SUNW_SYMINFO_FLAGS, VER_FLAGS)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from ..construct.lib.container import Container
+    from .elffile import ELFFile
 
 
 def describe_ei_class(x: str) -> str:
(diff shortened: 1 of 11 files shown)


elftools/elf/elffile.py
```diff
@@ -811,21 +845,21 @@ def _make_gnu_versym_section(self, section_header, name):
...
-    def _make_gnu_hash_section(self, section_header, name):
+    def _make_gnu_hash_section(self, section_header: Container, name: str) -> GNUHashSection:
         linked_symtab_index = section_header['sh_link']
         symtab_section = self._get_linked_symtab_section(linked_symtab_index)
         return GNUHashSection(
             section_header, name, self, symtab_section
         )
 
-    def _get_segment_header(self, n):
+    def _get_segment_header(self, n: int) -> Container:  # Elf_Phdr:
```
eliben, line 862: what does the comment mean?

elftools/elf/descriptions.py
```diff
@@ -128,18 +137,18 @@ def describe_symbol_local(x: int) -> str:
...
+def describe_symbol_other(x: Container) -> str:
     vis = describe_symbol_visibility(x['visibility'])
     if 1 < x['local'] < 7:
         return vis + ' ' + describe_symbol_local(x['local'])
     return vis
 
 
 def describe_symbol_shndx(x: int | str) -> str:
-    return _DESCR_ST_SHNDX.get(x, '%3s' % x)
+    return _DESCR_ST_SHNDX.get(x, f'{x:3}')  # type: ignore[arg-type] # ty: ignore[no-matching-overload]
```
eliben, line 148: Should we be able to get rid of these ignores at some point?

pmhahn: This 1st `# type: ignore` yes if you drop `mypy`; the 2nd `# ty` no.

The problem here is that `x: int | str`, while `_DESCR_ST_SHNDX: dict[int, str]`: The type-checkers tell you, that looking up a `str` in a dictionary with `int`-keys looks like an error. At runtime it isn't as `dict.get(key, default)` will return the `default` then.

elftools/elf/elffile.py
pmhahn: For improved type-safety I'd like to return a more specific type than `Container`, which just short of untyped `dict[Any, Any]`.
Here we know that it is an `Elf_Phrd`, but we have no sub-type of `Construct` to express that.

Maybe some day there is a chance to return something like a `TypedDict` or `dataclass`, which then has type-information for `p_type`, `p_offset`, `p_vaddr`, `p_filesz`, `p_memsz`, `p_flags`, `p_align`.

sevaa: I very much second that. ELF header is about the only place where a raw Container is exposed in the public API. It's been in my sights for a while. That said, we probably have consumers that access it using dict syntax; can't just replace with a strongly typed data structure.

PR #651, opened 2026-05-16
typing: elfrools.dwarf

pmhahn:
More typing for `elftools.dwarf` from https://github.com/eliben/pyelftools/pull/611

- more or less the boring part as most special cases are exempted
- some `TypeVars` and `@overloads` will follow later.
- several variables can be `None` and thus require an explicit `assert` for type-checking.
- Some of these `assert`s duplicate what `dwarf_assert()` already checked, which may be converted into a . [typing.TypeGuard](https://docs.python.org/3/library/typing.html#typing.TypeGuard), but I have not yet done that. I leave that to some future work.

### DIFF
--- a/elftools/dwarf/ranges.py
+++ b/elftools/dwarf/ranges.py
@@ -6,12 +6,22 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from __future__ import annotations
+
 import os
-from typing import NamedTuple
+from typing import IO, TYPE_CHECKING, NamedTuple, NoReturn
 
 from ..common.utils import struct_parse
 from ..common.exceptions import DWARFError
 from .dwarf_util import _iter_CUs_in_section
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterator
+
+    from ..construct.lib.container import Container
+    from .compileunit import CompileUnit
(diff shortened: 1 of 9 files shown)


elftools/dwarf/ranges.py
```diff
@@ -48,11 +58,21 @@ class RangeListsPair:
     it holds a RangeLists object for both and forwards API calls to the right one based
     on the CU version.
     """
-    def __init__(self, streamv4, streamv5, structs, dwarfinfo=None):
+    def __init__(
+        self,
```
eliben, line 62: something odd in the formatting here?

elftools/dwarf/ranges.py
```diff
@@ -105,14 +129,24 @@ class RangeLists:
...
         self.stream = stream
         self.structs = structs
         self._max_addr = 2 ** (self.structs.address_size * 8) - 1
         self.version = version
         self._dwarfinfo = dwarfinfo
 
-    def get_range_list_at_offset(self, offset, cu=None):
+    def get_range_list_at_offset(
+        self,
+            offset: int,
```
eliben: also here

elftools/dwarf/ranges.py
pmhahn: Somehow auto-indent in neovim is broken; I've fixed these 2 and one more.

eliben (approved):

PR #652, opened 2026-05-16
Convert to functools.cached_property

pmhahn:
While working on #650 and #651 I noticed a reoccurring topic: there are several private variables, which are used to lazy evaluate and then cache certain things. They are initialized with `None` in `__init__()`, which then requires many checks for `is not None`. This gets worse with type-checking as type-checkers require explicit `assert is not None`s to narrow down those union types.

Luckily there is [functools.cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property), which automates those cases: Instead of manually lazy-loading some variables, use `functools.cached_property` to get this for free: `AttributeError` on `__dict__` is handled by [`__getattr__()`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__), which then calls the registered function once and stored the result back in `__dict__`, so the value is there for all future uses.

This reduces duplicate code, type annotations and saves several (future) `assert`s.

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import copy
 import os
 from collections.abc import Iterator
+from functools import cached_property
 from typing import IO, TYPE_CHECKING, Any, Literal, NamedTuple, cast
 from warnings import warn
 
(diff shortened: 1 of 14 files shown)


eliben:
Great find on `cached_property`. Could you please isolate that and the bug fix into separate PRs?
The cached_property can be first

The code in dynamic.py is difficult to review now due to the other change

pmhahn:
> Great find on `cached_property`. Could you please isolate that and the bug fix into separate PRs? The cached_property can be first

@eliben See #654 for the bug-fix; I'll remove this from the PR once the fix is merged, so this PR will then be `cached_property` only.

sevaa (approved):

pmhahn:
Let's continue here after #654 got merged; this is now only the conversion to `cached_property`.

eliben (approved):

PR #654, opened 2026-05-18
Fix elftools/elf/dynamic.DynamicSegment.num_symbols()

pmhahn:
Part of #652 to fix the error calculating the number of symbols for "ultra stripped binaries" introduced by #219: In `elftools/elf/dynamic.py`: `DynamicSegment.num_symbols()` there an `if nearest_ptr is not None:` missing before the `return (nearest_ptr - tab_ptr) // self._symbol_size`.

---

The fallback to "Use the end of segment that contains DT_SYMTAB." also looked fishy:

```python
# Use the end of segment that contains DT_SYMTAB.
for segment in self.elffile.iter_segments():
    if (segment['p_vaddr'] <= tab_ptr and
            tab_ptr <= (segment['p_vaddr'] + segment['p_filesz'])):
        nearest_ptr = segment['p_vaddr'] + segment['p_filesz']
```

`ptr <= end` is used here since  687baa0cd6be158a32756fc7dd475e0e7f98f4d2, so `nearest_ptr` may point exactly at/after the end of a segment. That's not my definition of "contains", but maybe someone other knows this to be correct.

The [ELF spec](https://refspecs.linuxbase.org/elf/gabi4+/ch5.pheader.html) only guarantees, that
> Loadable segment entries in the program header table appear in ascending order, sorted on the `p_vaddr` member.

But there is no guarantee, that _segments may not overlap_.
I hope that heuristic here is required seldom enough that the current implementation is okay and that `pyelftools` is not used to analyze malicious binaries.

---

Maybe `PT_LOAD` should also be checked for as `iter_segments()` will walk all segments, not just which are loadable only?

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -291,6 +291,7 @@ class DynamicSegment(Segment, Dynamic):
             tab_ptr, tab_offset = self.get_table_offset('DT_SYMTAB')
             if tab_ptr is None or tab_offset is None:
                 raise ELFError('Segment does not contain DT_SYMTAB.')
+
             nearest_ptr: int | None = None
             for tag in self.iter_tags():
                 tag_ptr = tag['d_ptr']
@@ -305,11 +306,12 @@ class DynamicSegment(Segment, Dynamic):
                     nearest_ptr = tag_ptr
 
             if nearest_ptr is None:
-                # Use the end of segment that contains DT_SYMTAB.
-                for segment in self.elffile.iter_segments():
-                    if (segment['p_vaddr'] <= tab_ptr and
-                            tab_ptr <= (segment['p_vaddr'] + segment['p_filesz'])):
-                        nearest_ptr = segment['p_vaddr'] + segment['p_filesz']
+                # Use the end of last segment that contains DT_SYMTAB (or ends on it)
+                for segment in self.elffile.iter_segments(type='PT_LOAD'):
+                    start = segment['p_vaddr']
+                    end = start + segment['p_filesz']
+                    if start <= tab_ptr <= end:
+                        nearest_ptr = end
 
             if nearest_ptr is not None:
                 self._num_symbols = (nearest_ptr - tab_ptr) // self._symbol_size
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -311,8 +311,8 @@ class DynamicSegment(Segment, Dynamic):
                             tab_ptr <= (segment['p_vaddr'] + segment['p_filesz'])):
                         nearest_ptr = segment['p_vaddr'] + segment['p_filesz']
 
-            end_ptr = nearest_ptr
-            self._num_symbols = (end_ptr - tab_ptr) // self._symbol_size
+            if nearest_ptr is not None:
+                self._num_symbols = (nearest_ptr - tab_ptr) // self._symbol_size
 
         if self._num_symbols is None:
             raise ELFError('Cannot determine the end of DT_SYMTAB.')
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -357,4 +357,4 @@ class DynamicSegment(Segment, Dynamic):
             This method reads from the mandatory dynamic tag DT_SYMTAB.
         """
         for i in range(self.num_symbols()):
-            yield(self.get_symbol(i))
+            yield self.get_symbol(i)


eliben:
@rupran or @rhelmot -- any interest in reviewing?

elftools/elf/dynamic.py
```diff
@@ -305,14 +305,13 @@ def num_symbols(self) -> int:
                     nearest_ptr = tag_ptr
 
             if nearest_ptr is None:
-                # Use the end of segment that contains DT_SYMTAB.
+                # Use the end of last segment that contains DT_SYMTAB (or ends on it)
                 for segment in self.elffile.iter_segments():
-                    if (segment['p_vaddr'] <= tab_ptr and
-                            tab_ptr <= (segment['p_vaddr'] + segment['p_filesz'])):
-                        nearest_ptr = segment['p_vaddr'] + segment['p_filesz']
+                    if (start := segment['p_vaddr']) <= tab_ptr <= (end := start + segment['p_filesz']):
```
rupran: While we're at it, we could also change the calculation of the `end` value to use `p_memsz` instead of `p_filesz` as this represents the size of the segment in loaded memory. Should not make a difference for "regular" binaries as the only cases where these values differ are for .bss, but comparing virtual addresses with file sizes is not strictly correct.

elftools/elf/dynamic.py
```diff
@@ -305,14 +305,13 @@ def num_symbols(self) -> int:
                     nearest_ptr = tag_ptr
 
             if nearest_ptr is None:
-                # Use the end of segment that contains DT_SYMTAB.
+                # Use the end of last segment that contains DT_SYMTAB (or ends on it)
                 for segment in self.elffile.iter_segments():
```
rupran: As a reply to the suggestion in the PR: yes, I guess we should add `PT_LOAD` here to filter for loadable segments only.

rupran (commented):
Looks good to me in general, thanks for fixing these up. I added two comments inline which we could consider changing.

elftools/elf/dynamic.py
pmhahn: Makes sense; I have changed the code to use `p_memsz` instead.

elftools/elf/dynamic.py
pmhahn: I've changed the code to use `self.elffile.iter_segments(type='PT_LOAD')`. Thanks for for your review.

elftools/elf/dynamic.py
rupran: Hm, on second thought, `p_memsz` might actually not be what we want... as the number of symbols is used to parse information from the **file** in `get_symbol` and not from the loaded view in **memory**, this could lead to a larger number of symbols than expected if the .dynsym section is located in a segment with `p_memsz > p_filesz`... or am I missing something?

pmhahn: Now that you ask – I'm still no ELF expert, but from my current understanding `p_filesz <= p_memsz`; the tail gets zero-filled which does not make much sense for symbols. So using `p_filesz` makes more sense.
I'll change it back. Thank you for spotting this.

rupran: I checked again, and as `readelf` for example treats `p_filesz > p_memsz` as an error, we should be good with `p_filesz`.

From https://github.com/memfault/binutils-gdb/blob/master/binutils/readelf.c#L7671
```
	  if (segment->p_memsz < segment->p_filesz)
	    error (_("the segment's file size is larger than its memory size\n"));
```

rupran (approved):
LGTM now :)

elftools/elf/dynamic.py
```diff
@@ -305,14 +306,13 @@ def num_symbols(self) -> int:
...
 
             if nearest_ptr is None:
-                # Use the end of segment that contains DT_SYMTAB.
-                for segment in self.elffile.iter_segments():
-                    if (segment['p_vaddr'] <= tab_ptr and
-                            tab_ptr <= (segment['p_vaddr'] + segment['p_filesz'])):
-                        nearest_ptr = segment['p_vaddr'] + segment['p_filesz']
+                # Use the end of last segment that contains DT_SYMTAB (or ends on it)
+                for segment in self.elffile.iter_segments(type='PT_LOAD'):
+                    if (start := segment['p_vaddr']) <= tab_ptr <= (end := start + segment['p_filesz']):
```
eliben: So, in conclusion, we have the same condition here, or am I missing something?

TBH I find the original much easier to read

pmhahn: Yes, the condition is unchanged, but `type='PT_LOAD'` got added to `self.elffile.iter_segments(…)`!

Do you prefer something like this instead:
```python
        for segment in self.elffile.iter_segments(type='PT_LOAD'):
            start = segment['p_vaddr']
            end = start + segment['p_filesz']
            if start <= tab_ptr <= end:
                nearest_ptr = end
```

I find the original code hard to read because the line was really long and thus broken into 2, referenced `segment['p_vaddr']` 3 times and `segment['p_filesz']` 2 times; so a lot of repetition. Especially as Python has the chains `1 < 2 < 3` syntax; the 2nd `<=` was really hard to stop that way and added to the confusion.

eliben: I think this suggestion is a reasonable compromise.

Python's walrus operator is sometimes acceptable but this nested form is a bit much, IMO. Separating `start`. and `end` to their own lines is very readable and clear, and I don't mind the extra lines

pmhahn: Okay, I have applied that change and force-pushed by branch. Thank you.

eliben (approved):

eliben:
Thanks for the patience @pmhahn 

And thanks @rupran for the careful review!

PR #655, opened 2026-05-22
Minor typing cleanups

pmhahn:
1. I missed one wrong indentation from #649.
2. Otherwise several conversions to list-/dict-comprehensions and generator-expressions; this saves some explicit typing.
3. And using `next(…, None)` to get the 1st element instead of using `for … in …: …; break`

### DIFF
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -273,12 +273,14 @@ class DynamicSegment(Segment, Dynamic):
         # So we must look for the dynamic section contained in the dynamic
         # segment, we do so by searching for the dynamic section whose content
         # is located at the same offset as the dynamic segment
-        stringtable = None
-        for section in elffile.iter_sections():
-            if (isinstance(section, DynamicSection) and
-                    section['sh_offset'] == header['p_offset']):
-                stringtable = elffile.get_section(section['sh_link'])
-                break
+        stringtable = next(
+            (
+                elffile.get_section(section['sh_link'])
+                for section in elffile.iter_sections()
+                if isinstance(section, DynamicSection) and section['sh_offset'] == header['p_offset']
+            ),
+            None,
+        )
         Segment.__init__(self, header, stream)
         Dynamic.__init__(self, stream, elffile, stringtable, self['p_offset'],
              self['p_filesz'] == 0)
--- a/elftools/elf/dynamic.py
+++ b/elftools/elf/dynamic.py
@@ -127,16 +127,17 @@ class Dynamic:
     def get_table_offset(self, tag_name: str) -> tuple[int | None, int | None]:
         """ Return the virtual address and file offset of a dynamic table.
         """
-        ptr: int | None = None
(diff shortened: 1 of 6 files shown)


elftools/elf/dynamic.py
```diff
@@ -127,16 +127,11 @@ def __init__(
     def get_table_offset(self, tag_name: str) -> tuple[int | None, int | None]:
         """ Return the virtual address and file offset of a dynamic table.
         """
-        ptr: int | None = None
-        for tag in self._iter_tags(type=tag_name):
-            ptr = tag['d_ptr']
-            break
+        ptr: int | None = next((tag['d_ptr'] for tag in self._iter_tags(type=tag_name)), None)
```
eliben: Is this required for typing, or do you just find it clearer?

TBH I find the original code easier to understand

elftools/elf/dynamic.py
```diff
@@ -127,16 +127,11 @@ def __init__(
...
 
         # If we found a virtual address, locate the offset in the file
         # by using the program headers.
-        offset: int | None = None
-        if ptr:
-            offset = next(self.elffile.address_offsets(ptr), None)
+        offset = next(self.elffile.address_offsets(ptr), None) if ptr else None
```
eliben: same Q

elftools/elf/dynamic.py
pmhahn: Not "required" for typing, but my try to get rid of those `var: … | None = None` as those "alternatives" make the code harder to read.

Maybe the version I just force-pushed is more to your liking?

elftools/elf/dynamic.py
pmhahn: Not required, but type-checkers are able to automatically derive the type of `offset` as `int | None` if `next(…, Next)` is used. That allows to get rid of that typing redundancy and hopefully restored readability a bit. YMMV. If you prefer just drop it or have a look at the new version, which looks even nicer IMHO.

elftools/elf/dynamic.py
```diff
@@ -177,8 +178,10 @@ def _iter_tags(self, type: str | None = None) -> Iterator[Container]:
     def iter_tags(self, type: str | None = None) -> Iterator[DynamicTag]:
         """ Yield all tags (limit to |type| if specified)
         """
-        for tag in self._iter_tags(type=type):
-            yield DynamicTag(tag, self._get_stringtable())
+        return (
```
eliben: Here and below, why the change? The `yield` seems more readable and explicit to me

eliben (commented):
OK fair enough, understood about the typing inference. Left another question about something

elftools/elf/dynamic.py
pmhahn: Okay, I've dropped that patch. Reading some more about `yield` vs. generator-expression convinced me, that `yield` is more readable and even faster.

eliben (approved):

PR #657, opened 2026-05-26
Typing `readelf` and `dwarfdump`

pmhahn:
The last bits from #611 are extra as they do not type the API, but the `scripts/`. I have used this to develop the type-hints for the API, but this is not strictly required unless you consider them part of the API.

I find them very useful to understand, how the API is supposed to use. As some of their functions are large and complex, having some hint about the types being passed between those functions helps to understand their use.

Typing is not complete and `scripts/` is currently **not** checked: the path is missing in `pyproject.toml`.
1. The path cannot simply be added as both scripts define a `main()` function. Checking both files would required converting them into _modules_ so they can be `import`ed with different names.
2. There are many locations in both files, which requires more `assert isinstance(…)` or `assert … is not None` to be clean. IMHO that would clutter the code too much

If one or both is desired, I can have another look.

### DIFF
--- a/scripts/dwarfdump.py
+++ b/scripts/dwarfdump.py
@@ -13,10 +13,13 @@
 # Eli Bendersky (eliben@gmail.com)
 # This code is in the public domain
 #-------------------------------------------------------------------------------
+from __future__ import annotations
+
 import argparse
 import os
 import sys
 import traceback
+from typing import IO, TYPE_CHECKING, overload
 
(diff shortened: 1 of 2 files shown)


eliben (approved):

eliben:
Thanks, I'm OK with these minimal additions but agree that doing large changes just for the sake of typing these is overkill. These scripts are mainly used for testing.

PR #661, opened 2026-06-03
Minor typing fixes for dwarf

pmhahn:
:+1: to use `ruff`, but sadly you introduces one minor but by converting an _lambda_ into an _untyped function_.

I also had (another) look on how to remove the duplication of `{dwarf,elf}_assert()` vs. `assert`s required for typing. While doing that I stumbled over 3 `dwarf_assert(False, …)`, which should be replaced by an `raise DWARFError()` to make the error more obvious (and to reduce the length of the backtrace).

I also noticed, that `class FDE(CFIEntry)` has a slightly different type-annotation than its super-class, which I have fixed.

### DIFF
--- a/elftools/dwarf/callframe.py
+++ b/elftools/dwarf/callframe.py
@@ -14,6 +14,7 @@ from functools import cached_property
 from typing import IO, TYPE_CHECKING, Any, Literal, NamedTuple, cast
 from warnings import warn
 
+from ..common.exceptions import DWARFError
 from ..common.utils import (
     struct_parse, dwarf_assert, preserve_stream_pos)
 from ..construct import Struct, Switch
@@ -267,7 +268,7 @@ class CallFrameInfo:
                 case DW_CFA.GNU_args_size:
                     args = [struct_parse(structs.the_Dwarf_uleb128, self.stream)]
                 case _:
-                    dwarf_assert(False, f'Unknown CFI opcode: {raw_opcode:#04x}')
+                    raise DWARFError(f"Unknown CFI opcode: {raw_opcode:#04x}")
 
             instructions.append(CallFrameInstruction(opcode=opcode, args=args))
             offset = self.stream.tell()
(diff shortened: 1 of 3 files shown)


elftools/dwarf/callframe.py
```diff
@@ -267,7 +268,8 @@ def _parse_instructions(
                 case DW_CFA.GNU_args_size:
                     args = [struct_parse(structs.the_Dwarf_uleb128, self.stream)]
                 case _:
-                    dwarf_assert(False, f'Unknown CFI opcode: {raw_opcode:#04x}')
+                    msg = f"Unknown CFI opcode: {raw_opcode:#04x}"
+                    raise DWARFError(msg)
```
eliben: why not inline the message here?

elftools/dwarf/callframe.py
```diff
@@ -670,7 +672,8 @@ def _add_to_order(regnum: int) -> None:
                 case DW_CFA.nop | DW_CFA.AARCH64_negate_ra_state:
                     pass
                 case _:
-                    dwarf_assert(False, f"Unknown CFI opcode: {instr.opcode:#02x}")
+                    msg = f"Unknown CFI opcode: {instr.opcode:#02x}"
+                    raise DWARFError(msg)
```
eliben: ditto

elftools/dwarf/callframe.py
pmhahn: Because of [ruff check --select EM102](https://docs.astral.sh/ruff/rules/f-string-in-exception/).

eliben: Wow, what nonsense. Sorry, nothing personal, of course. I'm ok disabling this ruff check for the project -- it makes no sense to me.

pmhahn: I'm used to the rule in other project, so it was natural for me to use that style here, too. But as I don't care enough and you obviously don't like that style, I have changed to code to one-line `raise(f"…")` and force-pushed my branch.

eliben:
Thank you! Apparently the `ruff check` is passing with your latest changes. I'll test it more locally and will merge probably tomorrow

eliben (approved):

eliben:
EM102 checks are not enabled in this repo -- I'm just using the defaults; do you have all ruff checks enabled?

pmhahn:
> EM102 checks are not enabled in this repo -- I'm just using the defaults; do you have all ruff checks enabled?

Sometime yes, but not this time. I did the initial conversion from `dwarf_assert(False, …)` to `msg = …; raise DWARFError(msg)` from my muscle memory – as I'm used to do that for some other project where EM102 is enabled.

PR #662, opened 2026-06-08
Add support for FDO package metadata

tiran:
Parse `.note.package` sections (`NT_FDO_PACKAGING_METADATA`) as defined in "UAPI.8 Package Metadata for Executable Files". The JSON payload is decoded and exposed as a Python dict.

- Section: .note.package
- note type: 0xcafe1a7e
- Owner: FDO (FreeDesktop.org)
- Value: a single JSON object encoded as a NUL-terminated UTF-8 string

See https://uapi-group.org/specifications/specs/package_metadata_for_executable_files/


eliben (commented):
I'm actually not sure this fits in the core of pyelftools itself. Generally trying to stick to just the vanilla ELF/DWARF specs here.

There are potentially many standards/conventions built on top, and I'm concerned that if we start going in that direction there's no end to it.

Surely this functionality can be easily built on top of pyelftools, without having to modify its core?

PR #665, opened 2026-07-08
Fix iter_notes crash on GNU build-ID notes in core dumps

### LINKED ISSUE #656: Crash in iter_notes() when parsing NT_GNU_BUILD_ID notes inside ET_CORE files (type-3 note collision)
### Bug Description
When iterating over note entries inside a coredump ELF binary (`ET_CORE` type) using `iter_notes()`, the notes parser unconditionally decodes note type `3` as a process information note (`NT_PRPSINFO`), which expects a 124-byte struct payload (`Elf_Prpsinfo`).

However, type `3` inside a coredump note segment is only `NT_PRPSINFO` if the note name is `'CORE'`. If the note name is `'GNU'`, type `3` represents a standard `NT_GNU_BUILD_ID` note containing a raw binary build UUID (typically 16 or 20 bytes long).

Because `pyelftools` ignores the Note Name and switches the enums map solely on the ELF type `ET_CORE`, it incorrectly attempts to parse the 20-byte GNU Build ID note payload as the 124-byte `Elf_Prpsinfo` structure, leading to a fatal parser overrun exception:
`elftools.common.exceptions.ELFParseError: expected 4, found 0`

---

### Steps to Reproduce
Run the following minimal, self-contained Python script to reproduce the crash. It builds a valid 32-bit little-endian ARM coredump (`ET_CORE` type) containing a standard GNU Build ID note segment (type `3`, name `'GNU'`), and attempts to parse it using `iter_notes()`:

```python
import subprocess
from elftools.elf.segments import NoteSegment

# Construct a minimal, valid binary coredump containing a GNU Build ID note segment (type 3, name GNU)
# Using a 32-bit little-endian EM_ARM core header
data = bytearray()
# ELF Header (32-bit, little-endian)
data.extend(b"\x7fELF\x01\x01\x01\x00" + b"\x00" * 8) # e_ident
data.extend(b"\x04\x00")            # e_type = ET_CORE (4)
data.extend(b"\x28\x00")            # e_machine = EM_ARM (40)
data.extend(b"\x01\x00\x00\x00")    # e_version = 1
data.extend(b"\x00\x00\x00\x00")    # e_entry = 0
data.extend(b"\x34\x00\x00\x00")    # e_phoff = 52
data.extend(b"\x00\x00\x00\x00")    # e_shoff = 0
data.extend(b"\x00\x00\x00\x00")    # e_flags = 0
data.extend(b"\x34\x00")            # e_ehsize = 52
data.extend(b"\x20\x00")            # e_phentsize = 32
data.extend(b"\x01\x00")            # e_phnum = 1
data.extend(b"\x00\x00")            # e_shentsize = 0
data.extend(b"\x00\x00")            # e_shnum = 0
data.extend(b"\x00\x00")            # e_shstrndx = 0

# Program Header (PT_NOTE Segment)
data.extend(b"\x04\x00\x00\x00")    # p_type = PT_NOTE (4)
data.extend(b"\x54\x00\x00\x00")    # p_offset = 84
data.extend(b"\x54\x00\x00\x00")    # p_vaddr = 84
data.extend(b"\x54\x00\x00\x00")    # p_paddr = 84
data.extend(b"\x18\x00\x00\x00")    # p_filesz = 24
data.extend(b"\x18\x00\x00\x00")    # p_memsz = 24
data.extend(b"\x00\x00\x00\x00")    # p_flags = 0
data.extend(b"\x04\x00\x00\x00")    # p_align = 4

# Note Header (n_namesz=4, n_descsz=8, n_type=3)
data.extend(b"\x04\x00\x00\x00")    # n_namesz = 4
data.extend(b"\x08\x00\x00\x00")    # n_descsz = 8
data.extend(b"\x03\x00\x00\x00")    # n_type = 3
data.extend(b"GNU\x00")             # n_name = 'GNU'
data.extend(b"\xde\xad\xc0\xde\xde\xad\xc0\xde") # n_desc = 8 bytes Build ID (deadc0dedeadc0de)

# Write the constructed ELF data to disk
elf_filename = 'mock_core.elf'
with open(elf_filename, 'wb') as f:
    f.write(data)
print(f"Written mock core ELF to '{elf_filename}' ({len(data)} bytes)")

# Execute 'readelf --notes' on the written file
print("\nExecuting 'readelf --notes %s':" % elf_filename)
result = subprocess.run(['readelf', '--notes', elf_filename], capture_output=True, text=True, check=True)
print("--- readelf output start ---")
print(result.stdout.strip())
print("--- readelf output end ---")

# Open the file from disk and parse with pyelftools
print("\nParsing with pyelftools:")
with open(elf_filename, 'rb') as f:
    elf = ELFFile(f)
    note_seg = next(seg for seg in elf.iter_segments() if isinstance(seg, NoteSegment))
    notes = list(note_seg.iter_notes())
    
    assert len(notes) == 1, "Expected 1 note, got %d" % len(notes)
    note = notes[0]
    print("Parsed Note Type:", note['n_type'])
    print("Parsed Note Name:", note['n_name'])
    print("Parsed Note Desc:", note['n_desc'])
    
    assert note['n_type'] == 'NT_GNU_BUILD_ID', "Expected NT_GNU_BUILD_ID, got %r" % note['n_type']
    assert note['n_name'] == 'GNU', "Expected 'GNU', got %r" % note['n_name']
    expected_desc = 'deadc0dedeadc0de'
    assert note['n_desc'] == expected_desc, "Expected %r, got %r" % (expected_desc, note['n_desc'])
    print("Success!")
```

---

### Expected Behavior
`iter_notes()` should successfully parse the note segment. It should yield the GNU Build ID note entry with type `NT_GNU_BUILD_ID` and the description payload.

---

### Actual Behavior
`iter_notes()` crashes with `ELFParseError`:
```text
  File ".../site-packages/elftools/elf/notes.py", line 45, in iter_notes
    note['n_desc'] = struct_parse(elffile.structs.Elf_Prpsinfo, elffile.stream, offset)
  File ".../site-packages/elftools/common/utils.py", line 45, in struct_parse
    raise ELFParseError(str(e))
elffile.common.exceptions.ELFParseError: expected 4, found 0
```

---

### Root Cause Analysis
Inside `elftools/elf/structs.py` on lines 440-447:
```python
self.Elf_Nhdr = Struct('Elf_Nhdr',
    self.Elf_word('n_namesz'),
    self.Elf_word('n_descsz'),
    Enum(self.Elf_word('n_type'),
        **(ENUM_NOTE_N_TYPE if e_type != "ET_CORE"
           else ENUM_CORE_NOTE_N_TYPE)),
)
```
The note header type enum mapping is switched unconditionally based on the binary's `e_type`. If `e_type == "ET_CORE"`, it maps note type `3` to `'NT_PRPSINFO'` instead of `'NT_GNU_BUILD_ID'`. 

This dates back to #147 ("Better support for core dumps"), which assumed that *every* note segment inside a coredump ELF has coredump-specific types, overlooking the fact that coredumps can also contain standard generic GNU note segments.

---

### Suggested Fix Strategy
The note type enums decoder mapping should not be switched unconditionally based solely on `e_type`. It should also check the **Note Name (`n_name`)** if it is a coredump file:
*   If `n_name == 'GNU'`, use `ENUM_NOTE_N_TYPE` to decode the type (resolving type `3` to `NT_GNU_BUILD_ID`).
*   If `n_name == 'CORE'`, use `ENUM_CORE_NOTE_N_TYPE` (resolving type `3` to `NT_PRPSINFO`).

---

### Footnote
Note that it is rare, but valid, for a coredump to directly contain a `NT_GNU_BUILD_ID` note. Linux coredumps will instead typically contain the first page of all memory-mapped ELF files, and tools like GDB will discover the build ID(s) from those snapshots. However, other custom coredump producers (particularly embedded) may not have any memory-mapped ELF headers, and will emit the `NT_GNU_BUILD_ID` note directly.

apoorva-01:
Core dumps can carry a GNU build-ID note (type 3, name `GNU`). But in an `ET_CORE` file type 3 means `NT_PRPSINFO`, so `iter_notes()` reads the short build ID as a 124-byte prpsinfo struct and blows up with `ELFParseError`. readelf handles the same file fine.

Fix: if a note is named `GNU` in a core file, resolve its type against the generic GNU enum instead. Only touches core files.

Open question for you: the broader fix is to decode `n_type` as a raw int and pick the enum from `n_name` in `iter_notes`. More general, but it changes the shared struct, so I kept this contained. Your call.

Test builds a tiny ET_CORE ELF with a build-ID note. Fails before, passes after. Suite green, ruff clean.

Closes #656.

### DIFF
--- a/elftools/elf/notes.py
+++ b/elftools/elf/notes.py
@@ -12,12 +12,23 @@ from typing import TYPE_CHECKING
 
 from ..common.utils import struct_parse, roundup, bytes2str
 from ..construct import CString
+from .enums import ENUM_NOTE_N_TYPE, ENUM_CORE_NOTE_N_TYPE
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
     from ..construct.lib.container import Container
     from .elffile import ELFFile
+
+
+# In ET_CORE files the note type is decoded with the core-note enum, but a note
+# named 'GNU' uses the generic GNU note types regardless of the ELF type (e.g.
+# type 3 is NT_GNU_BUILD_ID, not the core NT_PRPSINFO). These maps recover the
+# raw type value so such notes can be re-resolved against the generic enum.
+_CORE_NOTE_TYPE_VALUE = {name: value for name, value in ENUM_CORE_NOTE_N_TYPE.items()
+                         if isinstance(value, int)}
+_GNU_NOTE_TYPE_BY_VALUE = {value: name for name, value in ENUM_NOTE_N_TYPE.items()
+                           if isinstance(value, int)}
 
 
 def iter_notes(elffile: ELFFile, offset: int, size: int) -> Iterator[Container]:
@@ -43,6 +54,14 @@ def iter_notes(elffile: ELFFile, offset: int, size: int) -> Iterator[Container]:
             offset += disk_namesz
         else:
             note['n_name'] = None
+
+        if note['n_name'] == 'GNU' and elffile['e_type'] == 'ET_CORE':
+            # A GNU-named note carries a generic GNU type even inside a core
+            # dump, so re-resolve the type that was decoded with the core enum.
+            raw_type = (note['n_type'] if isinstance(note['n_type'], int)
+                        else _CORE_NOTE_TYPE_VALUE.get(note['n_type']))
+            if raw_type in _GNU_NOTE_TYPE_BY_VALUE:
+                note['n_type'] = _GNU_NOTE_TYPE_BY_VALUE[raw_type]
 
         desc_data: bytes = elffile.stream.read(note['n_descsz'])
         note['n_descdata'] = desc_data
--- a/test/test_notes.py
+++ b/test/test_notes.py
@@ -1,8 +1,28 @@
+import io
 import os
+import struct
 import unittest
 
 from elftools.elf.elffile import ELFFile
 from elftools.elf.sections import NoteSection
+from elftools.elf.segments import NoteSegment
+
+
+def _build_core_with_gnu_note(n_type, desc):
+    """ Build a minimal 32-bit LE ARM ET_CORE ELF with a single PT_NOTE segment
+        holding one note named 'GNU' with the given type and descriptor.
+    """
+    name = b'GNU\x00'
+    note = struct.pack('<III', len(b'GNU') + 1, len(desc), n_type) + name + desc
+    ehsize, phentsize = 52, 32
+    p_offset = ehsize + phentsize
+    ehdr = (b'\x7fELF\x01\x01\x01\x00' + b'\x00' * 8 +
+            struct.pack('<HHIIIIIHHHHHH',
+                        4, 40, 1, 0, ehsize, 0, 0,
+                        ehsize, phentsize, 1, 0, 0, 0))
+    phdr = struct.pack('<IIIIIIII',
+                       4, p_offset, p_offset, p_offset, len(note), len(note), 0, 4)
+    return io.BytesIO(ehdr + phdr + note)
 
 
 class TestNotes(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


sevaa (approved):
Never had to deal with coredumps myself, but the logic looks sensible. We don't have a lot of coredumps in the corpus though, do we?

apoorva-01:
Yeah, just the three `core_linux*` fixtures. That's why I built the ET_CORE file inline in the test instead of committing another binary. Easy to point at a real coredump too if you'd rather have one in the corpus.

eliben:
While the logic looks ok, the whole issue feels extremely niche to me. It would be nicer to see a more general solution here, as well as a couple of representative core files committed in that actual comparison tests could run on.

apoorva-01:
I'll switch to the general version: resolve the type from `n_name` (GNU notes use GNU types, CORE notes use core types) instead of keying off `e_type`. That fixes the reverse case too.

For fixtures: what do you prefer? 
I can generate a small real coredump with a build-id note and wire it into the readelf comparison tests, or use a specific core if you have one in mind

eliben:
Generating a small real coredump is fine - but please include clear instructions on how to reproduce it (e.g. a simple C file, the gcc command to create the binary, and the execution steps).

Issue #482, opened 2023-07-21
Unhandled FormatField Exception


rpm5099:
This is very likely caused by corruption in an elf file, but I thought you might want to look into the possible cause.  Unfortunately I do not have a sample I can provide.  Here is the traceback:

```
---------------------------------------------------------------------------
FieldError                                Traceback (most recent call last)
File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/construct/core.py:351, in FormatField._parse(self, stream, context)
    350 try:
--> 351     return self.packer.unpack(_read_stream(stream, self.length))[0]
    352 except Exception as ex:

File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/construct/core.py:293, in _read_stream(stream, length)
    292 if len(data) != length:
--> 293     raise FieldError("expected %d, found %d" % (length, len(data)))
    294 return data

FieldError: expected 4, found 0

During handling of the above exception, another exception occurred:

FieldError                                Traceback (most recent call last)
File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/common/utils.py:41, in struct_parse(struct, stream, stream_pos)
     40         stream.seek(stream_pos)
---> 41     return struct.parse_stream(stream)
     42 except ConstructError as e:

File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/construct/core.py:190, in Construct.parse_stream(self, stream)
    183 """
    184 Parse a stream.
    185 
    186 Files, pipes, sockets, and other streaming sources of data are handled
    187 by this method.
    188 """
--> 190 return self._parse(stream, Container())

File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/construct/core.py:647, in Struct._parse(self, stream, context)
    646 else:
--> 647     subobj = sc._parse(stream, context)
    648     if sc.name is not None:

File ~/miniconda3/envs/py3114/lib/python3.11/site-packages/elftools/construct/core.py:353, in FormatField._parse(self, stream, context)
    352 except Exception as ex:
--> 353     raise FieldError(ex)

FieldError: expected 4, found 0

During handling of the above exception, another exception occurred:

ELFParseError                             Traceback (most recent call last)
Cell In[34], line 2
      1 eof = 0
----> 2 for i,seg in enumerate(elf.iter_segments()):
      3     print(i, seg.header.p_offset, seg.header.p_filesz)
[...]

PR #667, opened 2026-08-10
Raise ELFError on out-of-range string-table offset instead of OverflowError

eeshsaxena:
### Problem

`StringTableSection.get_string()` passes `sh_offset + offset` straight to `parse_cstring_from_stream()`, which seeks the stream to that absolute position. A corrupt name offset (easy to hit with a malformed or fuzzed ELF) can push the position out of range for `stream.seek()`, raising an exception that is not elftools' own `ELFError`:

```python
import io
data = bytearray(open('test/testfiles_for_readelf/simple_aarch64_gcc.o.elf','rb').read())
data[511] ^= 0xFF                       # corrupt a section-name offset
elf = ELFFile(io.BytesIO(bytes(data)))
[s.name for s in elf.iter_sections()]   # OverflowError: Python int too large to convert to C ssize_t
```

The exact exception depends on the stream type: `OverflowError`/`ValueError` on a `BytesIO`, or `ValueError`/`OSError` on a real file. Any of them can escape `ELFFile.iter_sections()` when a section (or symbol) name is read.

### Fix

Wrap the string-table lookup in `get_string()` and re-raise an out-of-range seek as `ELFError`, consistent with how elftools reports other malformed input.

### Test

Added `test_string_table_offset_out_of_range` to `test/test_corrupt_files.py`, asserting `get_string()` raises `ELFError` for both a huge and a negative offset. It fails with `OverflowError` before this change and passes after. The full unit-test suite still passes (120 tests).


elftools/elf/sections.py
```diff
@@ -156,7 +156,15 @@ def get_string(self, offset: int) -> str:
...
-        s = parse_cstring_from_stream(self.stream, table_offset + offset)
+        try:
+            s = parse_cstring_from_stream(self.stream, table_offset + offset)
+        except (OverflowError, ValueError, OSError) as e:
+            # A corrupt name offset can make the absolute stream position out
+            # of range for stream.seek(). Depending on the stream type this
+            # raises OverflowError/ValueError (BytesIO) or ValueError/OSError
+            # (a real file). Surface any of them as an ELFError.
+            raise ELFError(
+                'Invalid string offset %s in string table' % offset) from e
```
eliben: Could you use an f-string here?

eeshsaxena:
Done, switched it to an f-string.

sevaa:
Not to rain on the parade, but there are a myriad places where DWARF can be malformed in such a way that an offset leads nowhere. Are we really trying to put a semi-friendly exception on each of those cases?

eliben:
> Not to rain on the parade, but there are a myriad places where DWARF can be malformed in such a way that an offset leads nowhere. Are we really trying to put a semi-friendly exception on each of those cases?

That's a fair point!

@eeshsaxena did you encounter this in some real file you were working with?

sevaa:
Also, the "offset leads nowhere" condition doesn't necessarily mean "the offset is being generated wrongly". With DWARF being in no way self synchronizing, chances are the parser has misinterpreted a data structure elsewhere and went on to read an offset from bytes that were never meant to be one. The friendly exception might be misleading re: the root cause of the misparse. See #482.

eeshsaxena:
Honest answer: no, I did not hit this on a real binary. I was feeding deliberately malformed input, a string-table offset pointing past the end of the section, so treat it as a robustness nit rather than something from the wild.

@sevaa's point is fair and I do not want to carpet every dead-offset path with a friendly message, especially since a bad offset here is often a symptom of a misparse upstream rather than the real fault. The only thing that nudged me on this particular spot is the exception *type*: an out-of-range index leaks a bare `OverflowError` (or `IndexError`) where the surrounding code raises `ELFError` for malformed structures, so a caller wrapping `ELFError` gets surprised. If you would rather keep the parser thin and not special-case it, I am happy to close, no hard feelings. Your call.

eeshsaxena:
Honest answer to your question, @eliben: I hit this through malformed / fuzzed input while probing error handling, not a real-world file, so I don't have a genuine corrupt binary to point at.

@sevaa's point is fair and I think it's correct: since ELF/DWARF parsing isn't self-synchronizing, an out-of-range string offset here is usually a *symptom* of a misparse upstream rather than a genuinely bad offset in the file. So my message ("Invalid string offset ...") editorializes about a root cause it can't actually know.

The narrow thing I was after was just the exception *type*: `get_string()` currently lets a raw `OverflowError`/`ValueError`/`OSError` from `stream.seek()` escape, so code parsing an untrusted file can't catch it as a pyelftools error. Converting it to `ELFError` keeps the library's failure surface consistent for callers.

If that's worth having, I'm happy to reword the message to something neutral that doesn't claim the offset itself is wrong (e.g. "string table read failed at offset N"). And if you'd rather handle this closer to the root of the misparse per #482 instead, that's completely reasonable — feel free to close this one.

